PR #60, opened 2019-07-10
Add a versioning policy

wbl:
This defines our policy for versioning. Next will be releasing v0.0.1


armfazh:
Look at this https://semver.org/#how-should-i-deal-with-revisions-in-the-0yz-initial-development-phase

wbl:
That doesn't look relevant to this PR. does look relevant to the next step

PR #67, opened 2019-09-04
Fixes the output length of SignedDigit recoding

### LINKED ISSUE #61: ecc/p384: ScalarMult not constant time
I believe that since `math.SignedDigit` does not return a fixed number of coefficients,  `ScalarMult` in `ecc/p384` is not constant time.

To demonstrate, insert a `log.Print(len(L))` in the `ScalarMult` function and run the tests

```
$ go test -v -run ScalarMult
=== RUN   TestScalarMult
=== RUN   TestScalarMult/toOdd
=== RUN   TestScalarMult/k=0
=== RUN   TestScalarMult/random_k
2019/08/18 21:39:02 97
2019/08/18 21:39:02 97
2019/08/18 21:39:02 97
...
2019/08/18 21:39:02 97
2019/08/18 21:39:02 97
2019/08/18 21:39:02 96
2019/08/18 21:39:02 97
2019/08/18 21:39:02 96
2019/08/18 21:39:02 97
...
```

The `reduceScalar` function attempts to ensure the scalar is always `sizeFp` bytes, however this is lost in the conversion to `big.Int` if there are leading zero bytes.

Note that `math.SignedDigit` uses `BitLen` to determine the size of the input, which depends on the value of the integer.

https://github.com/cloudflare/circl/blob/a03c5a147111a46165b047f49053ec510d5582b4/math/wnaf.go#L22

armfazh:
Issue:
the `SignedDigit` function returns a slice whose size depends on the length of the input. 

Changes:
 - Added a third parameter for `SignedDigit`, which is the size of the maximum bit-length of n. 
 - In the `ScalarMult` call, `SignedDigit` is invoked using the bit-length of the order of the curve.

Fixes #61.


armfazh:
@henrydcase could you please take a look on CI/maintenance test.

henrydcase:
@armfazh PR breaks the build

armfazh:
> @armfazh PR breaks the build, no idea why. I'll let you figure it out

Resolved in #68

math/wnaf_test.go
```diff
@@ -35,15 +35,14 @@ func TestOmegaNAF(t *testing.T) {
...
+	Two128 := big.NewInt(1)
+	Two128.Lsh(Two128, 128)
 
 	for w := uint(2); w < 10; w++ {
 		for j := 0; j < testTimes; j++ {
-			x, _ := rand.Int(rand.Reader, &max)
+			x, _ := rand.Int(rand.Reader, Two128)
 			x.SetBit(x, 0, uint(1)) // odd-numbers
-			L := SignedDigit(x, w)
+			L := SignedDigit(x, w, 128)
```
wbl: I think we also should test that the length is what we think it is. We might also want to include edge cases in the representation such as 2^w, 2^w-1, etc.

PR #68, opened 2019-09-09
Building golangci-lint during CI using go1.13

armfazh:
Changes:
 - Fetching golangci-lint from goreleaser.com brings a version compiled
   with go1.12, which once it is run with go1.13 fails.
 - Once goreleaser.com build golangci-lint with go1.13, then it can be
   restored to the previous fetching method.
 - Removing funlen linter, which was added in the latest version.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -46,7 +46,7 @@ version: 2.1
         - run:
             name: "Linting"
             command:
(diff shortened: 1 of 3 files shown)


armfazh:
Tests are ok. 
Source of fix: https://github.com/golangci/golangci-lint/issues/658#issuecomment-528126696
One can revert these changes once golangci-lint built with go1.13 be available at goreleaser.com

PR #75, opened 2019-10-17
Implementation of p434 finite field for SIDH/SIKE

### LINKED ISSUE #69: Align SIKE with Round 2
Goal:
- implement new starting curve
- implement P434

I've already done it here, so it's a matter of porting
https://github.com/post-quantum-cryptography/go-sike

henrydcase:
* adds p434 finite field with both generic and optimized
  assembly implementation for amd64
* removes "arith_amd64_test.gotemp" template as it requires
  multiple implementations of mul and rdc for x86-64, which
  do not exist in p434. In the future we should abandon
  separated implementations for CPUs with BMI2 and MULX+ADxX
* p434 implementation doesn't contain optimized implementation
  for arm64. Because of that `gen.go` contains now a flag
  indicating whether ARM optimized implementation is available.
  This flag is used for setting up build tags correctly in
  arith_decl.gotemp and arith_generic.gotemp

Fixes #69


armfazh:
Still reviewing, 25% acceleration follows the theoretical improvement.

PR #77, opened 2019-12-03
Fixes incomplete modular reduction.

armfazh:
Reduction modulo p448 requires to propagate a carry bit.


armfazh:
> Add a test for a known-bad test vector and this will be good.

That is done in #78 . Merging both PR could be better

wbl:
Well, then merge em separately or together but either works for me.

armfazh:
Merged at #78

PR #78, opened 2019-12-04
Adds a fuzzer and tests for edge cases of reduction.

armfazh:
Includes a fuzzer target to discover edge input cases for red64.
The inputs found were added into a go test, so the current master branch will fail, indicating the presence of errors.


armfazh:
Merged #77 here.

PR #82, opened 2020-01-28
New API for signing avoids to manipulate a public key.

### LINKED ISSUE #81: Update signing API for ed25519
The API of ed25519 accepts a public key as input, this is a bad design. 

Discussions:
https://moderncrypto.org/mail-archive/curves/2020/001013.html
https://mailarchive.ietf.org/arch/msg/cfrg/pv7X8hrBtMGfmlCMHsLhJDXPHso

armfazh:
Changes:
 - The PrivateKey struct now contains the private key (seed) and
   a copy of the public key.
 - The PrivateKey struct do not allow to manipulate the public key.
 - The PrivateKey implements crypto.Signer, if someone needs it.

Fixes #81.


sign/ed25519/ed25519.go
```diff
@@ -2,42 +2,93 @@ package ed25519
...
+// len(seed) is not SeedSize. This function is provided for interoperability
+// with RFC 8032. RFC 8032's private keys correspond to seeds in this
+// package.
+func NewKeyFromSeed(seed []byte) *PrivateKey {
+	if l := len(seed); l != Size {
+		panic("ed25519: bad seed length: " + strconv.Itoa(l))
+	}
+	var P pointR1
+	var private PrivateKey
+	k := sha512.Sum512(seed)
```
Lekensteyn: Perhaps rename seed -> secret?
Following https://ed25519.cr.yp.to/ed25519-20110926.pdf and https://tools.ietf.org/html/rfc8032#section-3.2,
the input for this function is the secret key `k` (the EdDSA private key). The result of hashing `k` is internally used for signing/verification and referred to as `s`.

sign/ed25519/ed25519.go
```diff
@@ -2,42 +2,93 @@ package ed25519
...
+	copy(private.seed[:], seed[:])
+	return &private
 }
 
 // Sign returns the signature of a message using both the private and public
 // keys of the signer.
-func (e Pure) Sign(message []byte, public *PubKey, private *PrivKey) *Signature {
-	k := sha512.Sum512(private[:])
+func Sign(private *PrivateKey, message []byte) Signature {
+	k := sha512.Sum512(private.seed[:])
```
Lekensteyn: Since the original private key is not in use, would it make sense storing `s = H(k)` instead and avoid the recomputation here?

sign/ed25519/ed25519_test.go
```diff
@@ -127,296 +127,61 @@ func TestWrongPublicKey(t *testing.T) {
...
 		b.ResetTimer()
 		for i := 0; i < b.N; i++ {
-			scheme.Verify(msg, &public, sig)
+			Verify(public, msg, &signature)
 		}
 	})
 }
+
+//
+// func Example_ed25519() {
```
Lekensteyn: fix this or remove it?

Lekensteyn:
The API changes looks good to me! I just have a few questions about the naming, see the inline comments.

sign/ed25519/ed25519.go
armfazh: `H(k)` is twice as large as `k` . Also hashing a short-length string it is not so painful.

sign/ed25519/ed25519.go
armfazh: Since this is an internal variable I can rename it to `private` key. 
I was trying to follow the stdlib implementation, but `private` key is the correct name.

PR #86, opened 2020-02-12
Changes to SHAKE in preparation for Dilithium support

bwesterb:
1. (Re)add support for SHAKE-256.
2. Reduce allocations.  The byte-slice in the shake State struct
   confused Go's escape analysis and forced it onto the Heap.
3. Move out of dh/sidh as it will be used by sign/dilithium.


bwesterb:
Ok, will prepare a PR later this week.

Issue #89, opened 2020-03-03
Shake is crashing when using unsafe pointer casting


armfazh:
In CI, there is triggered an crash error inside shake package.
Details: 
 - https://app.circleci.com/jobs/github/cloudflare/circl/1647
 - https://app.circleci.com/jobs/github/cloudflare/circl/1648

Bas, could you please take a look on this.

armfazh:
https://golang.org/doc/go1.14#compiler
> This release (1.14) adds -d=checkptr as a compile-time option for adding instrumentation to check that Go code is following unsafe.Pointer safety rules dynamically.

It seems that the unsafe conversion is failing happening only on Go 1.14.

bwesterb:
This is the same issue as reported for Go in https://github.com/golang/go/issues/35173 which was fixed by https://go-review.googlesource.com/c/crypto/+/203837/

bwesterb:
Ok, so this problem also occurs upstream: https://github.com/golang/go/issues/37644

PR #91, opened 2020-03-04
Adding test and build in go 1.14

armfazh:
Known issues: coverage is failing in go 1.14. 
Merge this after #89 be solved

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -60,7 +60,7 @@ jobs:
     amd64:
         <<: *nativejob
         environment:
-            GOVER: latest
+            GOVER: 1.14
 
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -60,7 +60,7 @@ jobs:
     amd64:
         <<: *nativejob
         environment:
-            GOVER: latest
+            GOVER: 1.14
 
     # Test against older version of golang
     golang_1_12:
```
wbl: Should we add 1_13 as an older version here?

armfazh: This case is covered by `golang_1_13` job.

PR #99, opened 2020-04-04
Adding MLSBSet constant-time exponentiation algorithm

armfazh:
This PR includes the MLSBSet constant-time exponentiation algorithm.
It allows fixed-point scalar multiplication.
Reference: https://eprint.iacr.org/2013/158


armfazh:
Merged in #100

PR #100, opened 2020-04-06
Adds the goldilocks curve

armfazh:
This is the implementation of the elliptic curve operations for goldilocks, a.k.a edwards448.

Changes:
 - It's based on top #99 
 - it will serve as a base for decaf and ed448.
 - Provides three multiplication cases, kG, kP and k_0P+k_1Q.


ecc/goldilocks/curve.go
```diff
@@ -0,0 +1,80 @@
...
+}
+
+// IsOnCurve returns true if the point lies on the curve.
+func (Curve) IsOnCurve(P *Point) bool {
+	x2, y2, t, t2, z2 := &fp.Elt{}, &fp.Elt{}, &fp.Elt{}, &fp.Elt{}, &fp.Elt{}
+	rhs, lhs := &fp.Elt{}, &fp.Elt{}
+	fp.Mul(t, &P.ta, &P.tb)  // t = ta*tb
+	fp.Sqr(x2, &P.x)         // x^2
+	fp.Sqr(y2, &P.y)         // y^2
+	fp.Sqr(z2, &P.z)         // y^2
```
wbl, line 24: // z^2

ecc/goldilocks/scalar.go
```diff
@@ -0,0 +1,184 @@
...
+	for i := nCeil - _N - 1; i >= 0; i-- {
+		low := binary.LittleEndian.Uint64(x[8*i:])
+		high := z64.leftShift(low)
+		z64.reduceOneWord(high)
+	}
+	z64.modOrder()
+}
+
+// divBy4 calculates z = x/4 mod order.
+func (z *Scalar) divBy4(x *Scalar) { z.Mul(x, &invFour) }
```
wbl, line 130: If this is fast enough it's fine. Could also add (z mod 4)*p to z, then divide by 4 via shifting, and reduce.

math/mlsbset/mlsbset.go
```diff
@@ -0,0 +1,122 @@
+// Package mlsbset provides a constant-time exponentiation method with precomputation.
```
wbl, line 1: This seems to be included in the other PR: i'm reviewing it there

ecc/goldilocks/curve.go
armfazh, line 24:
```suggestion
	fp.Sqr(z2, &P.z)         // z^2
```

ecc/goldilocks/scalar.go
armfazh: Yes, it is barely used, so I went for a multiplication,

math/mlsbset/mlsbset.go
armfazh: That is because this PR is on top of the mlsbset PR.

PR #102, opened 2020-04-08
Add ed448 signature scheme.

armfazh:
Adds support for Ed448 digital signature scheme as described in [RFC8032](https://tools.ietf.org/html/rfc8032).

Relies on mlsbset (#99),  goldilocks (#100), and internal/sha3 packages.


armfazh:
@claucece please take a look on this PR

claucece:
Oh! Very nice! I'll check it tomorrow ;)

sign/ed448/ed448.go
```diff
@@ -0,0 +1,187 @@
...
+	_s.Add(&_s, &_r)
+	copy(signature[Size:], _s[:])
+	return signature
+}
+
+// Verify returns true if the signature is valid. Failure cases are invalid
+// signature, or when the public key cannot be decoded.
+func Verify(public PublicKey, message, context, signature []byte) bool {
+	P, errDecoding := goldilocks.FromBytes(public)
+	if len(public) != Size ||
```
claucece, line 139: This check should probably be executed prior to the decoding

sign/ed448/ed448.go
```diff
@@ -0,0 +1,187 @@
...
+
+func clamp(k []byte) {
+	k[0] &= 252
+	k[Size-2] |= 0x80
+	k[Size-1] = 0x00
+}
+
+func makeCopy(in *[Size]byte) []byte {
+	out := make([]byte, Size)
+	copy(out, in[:])
```
claucece, line 189: This might be good to do in constant time as it is sometimes copying the private key...

sign/ed448/ed448.go
```diff
@@ -0,0 +1,187 @@
...
+func Sign(kp *KeyPair, message, context []byte) []byte {
+	if len(context) > 255 {
+		panic("context should be at most 255 octets")
+	}
+	var r, h, hRAM [2 * Size]byte
+	H := sha3.NewShake256()
+	_, _ = H.Write(kp.private[:])
+	_, _ = H.Read(h[:])
+
+	prefix := [10]byte{'S', 'i', 'g', 'E', 'd', '4', '4', '8', byte(0), byte(len(context))}
```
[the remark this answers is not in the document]
armfazh: right, I fixed accordingly

sign/ed448/ed448.go
armfazh: does copy present time variability?

bwesterb: I think it's constant time, but in this case we can just write

```go
ret := *in
return ret[:]
```

which is definitely constant time.  Note that it is **not** the same as

```go
return (*in)[:]
```

sign/ed448/ed448.go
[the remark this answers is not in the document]
claucece: ;)

sign/ed448/ed448.go
claucece: I agree with @bwesterb ;)

ecc/goldilocks/curve_test.go
```diff
@@ -0,0 +1,113 @@
...
+import (
+	"crypto/rand"
+	"testing"
+
+	"github.com/cloudflare/circl/ecc/goldilocks"
+	"github.com/cloudflare/circl/internal/test"
+)
+
+func TestScalarMult(t *testing.T) {
+	const testTimes = 1 << 8
```
mmcloughlin, line 12: Not an issue that needs to be addressed right now, but it can be quite nice for things like this is to allow the iterations to vary based on the `-short` flag. For example:

```go
func NumTrials() {
    if testing.Short() {
        return 16
    }
    return 1 << 8 // or more
}
```

If you want to get even more fancy, something I've done before is to also define custom `-long` and `-stress` flags. So you can do:

```go
// Custom test command line flags.
var (
	long   = flag.Bool("long", false, "enable long running tests")
	stress = flag.Bool("stress", false, "enable stress tests (implies -long)")
)

// timeallowed returns how long a single test is allowed to take.
func timeallowed() time.Duration {
	switch {
	case testing.Short():
		return time.Second / 10
	case *long:
		return 30 * time.Second
	case *stress:
		return 2 * time.Minute
	default:
		return time.Second
	}
}

// Repeat the given trial function. The duration is controlled by custom
// command-line flags. The trial function returns whether it wants to continue
// testing.
//
//	-short		run for less time than usual
//	-long		allow more time
//	-stress		run for an extremely long time
func Repeat(t *testing.T, trial func(t *testing.T) bool) {
	start := time.Now()
	d := timeallowed()
	n := 1
	for time.Since(start) < d && trial(t) {
		n++
	}
	t.Logf("%d trials in %s", n, time.Since(start))
}
```

I usually put something like this in an `internal/test` package. Then you wrap your test in a `test.Repeat(...)` call and it will run until the time limit.

ecc/goldilocks/scalar.go
```diff
@@ -0,0 +1,184 @@
+package goldilocks
+
+import (
+	"encoding/binary"
+	"math/bits"
+	"unsafe"
+)
+
+// ScalarSize is the size (in bytes) of scalars.
+const ScalarSize = 56 // 448 / 8
```
mmcloughlin, line 10: nit:

```suggestion
const ScalarSize = 448 / 8
```

ecc/goldilocks/curve_test.go
armfazh: That is a good suggestion, I will open a ticket for tracking your proposal.

mmcloughlin:
By the way... it's not pretty but it is possible to "stack" PRs in github. You have three PRs `A -> B -> C`, instead of having them all with target branch `master` you can set the target branch of `B` to be `A`, and target branch of `C` to be `B`. Landing them is mildly annoying.

My company has custom tooling for this and it's really sweet. Maybe we should open source it.

Just saying because this PR stack is intimidating to review!

claucece:
Oh @mmcloughlin , that is very awesome! Does the tool have a GUI or it's from the terminal? I always wanted something like it, and I keep an eye on: https://github.com/isaacs/github/issues/867, https://github.com/isaacs/github/issues/959.

I like this post around the subject very much: https://unhashable.com/stacked-pull-requests-keeping-github-diffs-small/

mmcloughlin:
> Oh @mmcloughlin , that is very awesome! Does the tool have a GUI or it's from the terminal? I always wanted something like it, and I keep an eye on: [isaacs/github#867](https://github.com/isaacs/github/issues/867), [isaacs/github#959](https://github.com/isaacs/github/issues/959).
> 
> I like this post around the subject very much: https://unhashable.com/stacked-pull-requests-keeping-github-diffs-small/

Ah I just realized that our internal tool (we call it `bonsai`) might be hard to open source since it's tightly integrated with our submit queue, which is another custom tool for our huge monorepo.

I really hope Github comes up with an official solution for this. Seems likely they might... they've been coming out with so many new features since the Microsoft acquisition.

armfazh:
I addressed most of comments in the review, could you all give another look on this PR.
cc: @claucece @bwesterb

claucece:
Nice! Thanks for the changes @armfazh ;)

PR #105, opened 2020-04-23
Updates API for signatures.

armfazh:
Changes:
 - Unifies the API for EdDSA signatures.
 - Both packages implement `crypto.Signer` so it can be used with other signature schemes.
 - Added tests for ed25519.

Future changes (not this PR)
 - extract edwards25519 arithmetic to the `circl/ecc` package.

### DIFF
--- a/sign/ed25519/ed25519.go
+++ b/sign/ed25519/ed25519.go
@@ -16,10 +16,15 @@ import (
 )
 
 const (
-	// Size is the length in bytes of Ed25519 keys.
-	Size = 32
+	// PublicKeySize is the length in bytes of Ed25519 public keys.
+	PublicKeySize = 32
+	// PrivateKeySize is the length in bytes of Ed25519 private keys.
+	PrivateKeySize = 32
 	// SignatureSize is the length in bytes of signatures.
-	SignatureSize = 2 * Size
+	SignatureSize = 64
+)
+const (
+	paramB = 256 / 8 // Size of keys in bytes.
 )
 
 // PublicKey represents a public key of Ed25519.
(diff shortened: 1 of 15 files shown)


armfazh:
@bwesterb could you please take a look on the API for signatures.

sign/ed25519/ed25519.go
```diff
@@ -1,16 +1,26 @@
...
 	"crypto/sha512"
 	"errors"
 	"io"
 )
 
-// Size is the length in bytes of Ed25519 keys.
-const Size = 32
+const (
+	// Size is the length in bytes of Ed25519 keys.
+	Size = 32
```
bwesterb, line 20: Of both public and private keys, which is a bit of a coincidence: one is a group element and the other one is a scalar.  Perhaps define `PublicKeySize` and `PrivateKeySize`?

sign/ed25519/ed25519.go
```diff
@@ -22,117 +32,139 @@ type PrivateKey []byte
...
-func (k *KeyPair) GetPrivate() PrivateKey { return makeCopy(&k.private) }
+func (k *KeyPair) GetPrivate() PrivateKey { z := k.private; return z[:] }
+
+// GetPublic returns a copy of the public key.
+func (k *KeyPair) GetPublic() PublicKey { z := k.public; return z[:] }
 
-// GetPublic returns the public key corresponding to the private key.
-func (k *KeyPair) GetPublic() PublicKey { return makeCopy(&k.public) }
+// Seed returns the private key seed.
+func (k *KeyPair) Seed() []byte { return k.GetPrivate() }
```
bwesterb, line 41: `Seed` is the terminology used by Go's stdlib ed25519, because they view the SHA512 expansion as the private key, whereas the RFC and Dan define the "seed" as the private key.  I think `Private()` is more appropriate here.

sign/ed25519/ed25519.go
```diff
@@ -1,16 +1,26 @@
...
-const Size = 32
+const (
+	// Size is the length in bytes of Ed25519 keys.
+	Size = 32
+	// SignatureSize is the length in bytes of signatures.
+	SignatureSize = 2 * Size
+)
 
 // PublicKey represents a public key of Ed25519.
 type PublicKey []byte
```
bwesterb, line 26: Why not `*[Size]byte`?

sign/ed25519/ed25519.go
armfazh: I share this though too, however the reason for providing these methods and these terminology is to serve as a drop-in replacement wrt stdlib (which doesn't use the best terms/interfaces)
Do you see value on providing these methods?

sign/ed25519/ed25519.go
armfazh: Same as in my comment above. I agree that a constant-sized array is better. Also, how does a user of CIRCl interoperates between ed25519 and ed448 signatures?

sign/ed25519/ed25519.go
bwesterb: Ah, being a drop-in replacement for the sodlib one is a good argument.  It's value is in interoperability.

sign/ed25519/ed25519.go
armfazh: Fixed now,

PR #111, opened 2020-05-08
4-way KeccaK-f[1600] using AVX2 to speed up Dilithium

bwesterb:
Will also be useful to speed up XMSS[MT], SPHINCS+, etc.

Non-AVX2:

```
BenchmarkPermutationFunction-8   	 3594723	       326 ns/op	 613.13 MB/s
```

AVX2:

```
BenchmarkF1600x4-8   	 2926183	       399 ns/op
```

The latter computes four times as much so it's a 3.2x speedup

Applied to Dilithium (mode3) we get (new):

```
BenchmarkSkUnpack-8                      	   37950	     30844 ns/op
BenchmarkPkUnpack-8                      	   39169	     30442 ns/op
BenchmarkVerify-8                        	   76672	     15513 ns/op
BenchmarkSign-8                          	   10000	    125782 ns/op
BenchmarkGenerateKey-8                   	   17419	     68635 ns/op
BenchmarkPublicFromPrivate-8             	  134430	      9250 ns/op
```

old:

```
BenchmarkSkUnpack-8            	   18980	     61840 ns/op
BenchmarkPkUnpack-8            	   19743	     60087 ns/op
BenchmarkVerify-8              	   39998	     30421 ns/op
BenchmarkSign-8                	    5420	    228631 ns/op
BenchmarkGenerateKey-8         	   12037	     99672 ns/op
BenchmarkPublicFromPrivate-8   	  107148	     11512 ns/op
```

The non-optimized C reference implementation:

```
BenchmarkKeygen-8   	   10000	    109216 ns/op
BenchmarkSign-8     	    2206	    545616 ns/op
BenchmarkVerify-8   	   10000	    108419 ns/op
```

The AVX2 optimized C reference implementation:

```
BenchmarkKeygen-8   	   36510	     32242 ns/op
BenchmarkSign-8     	   10000	    117060 ns/op
BenchmarkVerify-8   	   31605	     34123 ns/op
```

See #113


internal/shakex4/f1600x4_amd64_src.go
```diff
@@ -0,0 +1,148 @@
...
+	// (array of) uint64 to be alligned on 8 bytes except if it's heap
+	// allocated.  Thus we do not add the noescape pragma to ensure that
+	// state is heap allocated and thus 16 byte alligned.
+
+	state_ptr := Load(Param("state"), GP64())
+	state := func(offset int) Op {
+		return Mem{Base: state_ptr, Disp: 32 * offset}
+	}
+
+	rc_ptr := Load(Param("rc"), GP64())
```
mmcloughlin, line 28: If you don't mind duplicating the round constants you could also define them as a `DATA` section.

https://github.com/mmcloughlin/avo/tree/master/examples/data

internal/shakex4/f1600x4_amd64_src.go
```diff
@@ -0,0 +1,148 @@
...
+func main() {
+	ConstraintExpr("amd64")
+
+	TEXT("f1600x4AVX2", 0, "func(state *[100]uint64, rc *[24]uint64)")
+
+	// Pragma("noescape")  // XXX
+	// We want state to be 16 byte alligned.  Go only guarantees an
+	// (array of) uint64 to be alligned on 8 bytes except if it's heap
+	// allocated.  Thus we do not add the noescape pragma to ensure that
+	// state is heap allocated and thus 16 byte alligned.
```
mmcloughlin, line 21: Hmm, this feels fragile. Do you need alignment for `VMOVDQA`, or is there another reason? Have you tried `VMOVDQU`? I suspect the unaligned version will incur a small or negligible performance hit, probably worth it for correctness.

internal/shakex4/f1600x4_amd64_src.go
```diff
@@ -0,0 +1,148 @@
...
+
+import (
+	. "github.com/mmcloughlin/avo/build"
+	. "github.com/mmcloughlin/avo/operand"
+)
+
+func main() {
+	ConstraintExpr("amd64")
+
+	TEXT("f1600x4AVX2", 0, "func(state *[100]uint64, rc *[24]uint64)")
```
mmcloughlin, line 15: Doesn't look like you are using stack space? In that case you can use the `NOSPLIT` flag.

internal/shakex4/f1600x4_amd64_src.go
armfazh: Do you know how to request the go compiler for aligned memory ?

internal/shakex4/f1600x4_amd64_src.go
bwesterb: Yes, you’rer right. Thanks.

internal/shakex4/f1600x4_amd64_src.go
bwesterb: @mmcloughlin Memory is not just loaded at the start, but it’s read and loaded between each round. I think it’ll have a measurable impact. (Vector operations which do allow unaligned pointers are generally still faster with aligned pointers.)

@armfazh I don’t think there is good support for it yet.

mmcloughlin: @bwesterb I made the change to `VMOVDQU` locally and benchmarked it each time.

```
$ cat aligned.txt unaligned.txt 
goos: linux
goarch: amd64
pkg: github.com/cloudflare/circl/internal/shakex4
BenchmarkF1600x4-8   	 2147116	       556 ns/op
PASS
ok  	github.com/cloudflare/circl/internal/shakex4	1.762s
goos: linux
goarch: amd64
pkg: github.com/cloudflare/circl/internal/shakex4
BenchmarkF1600x4-8   	 2194231	       546 ns/op
PASS
ok  	github.com/cloudflare/circl/internal/shakex4	1.754s
```

Unaligned was faster in this quick uncontrolled test, which is probably noise, but that just goes to show how small the effect is. I encourage you to benchmark this yourself, and I think you'll find the performance change (if any at all) is so small that it's not worth the trouble of forcing Go to give you aligned memory, since as you say there is not good support for it.

https://lemire.me/blog/2012/05/31/data-alignment-for-speed-myth-or-reality/

I'm happy to be proven wrong, but I just think you should use the unaligned access until you can demonstrate alignment really helps.

armfazh: The distinction in time is not observed in our current processors and laptops, say after haswell u-arch (2014 approx). 
But in it will appear in older processors. In those one probably see a difference in performance.

armfazh: For the matters of this PR and CIRCL, I am ok with using unaligned memory access, until Go provides a way to ensure this.

mmcloughlin: @armfazh AVX2 was introduced in Haswell. This code can't run on older microarchitectures anyway.

armfazh: 256-bit registers appear in the Sandy Bridge arch and it was called AVX. Hence, there was load aligned and unaligned before haswell.

mmcloughlin: Yes, but for the purposes of this PR, that doesn't matter. Other instructions in this code require AVX2.

bwesterb: I ran my own benchmarks.  There doesn't seem to be a big difference between VMOVDQA and VMOVDQU for **aligned** memory.  If I change alignment, I get a noticeable slowdown.  (I ran the benchmarks interleaved and then sorted them to account for my CPU throttling.)  

```
BenchmarkF1600x4-8               2783442           428 ns/op                     
BenchmarkF1600x4-8               2850756           421 ns/op                     
BenchmarkF1600x4-8               2858080           431 ns/op                     
BenchmarkF1600x4aligned-8        2700562           430 ns/op                     
BenchmarkF1600x4aligned-8        2888199           437 ns/op                     
BenchmarkF1600x4aligned-8        2889950           425 ns/op                     
BenchmarkF1600x4plusEight-8      2475714           487 ns/op                     
BenchmarkF1600x4plusEight-8      2558049           455 ns/op                     
BenchmarkF1600x4plusEight-8      2564606           465 ns/op                     
BenchmarkF1600x4plusFour-8       2580600           463 ns/op                     
BenchmarkF1600x4plusFour-8       2585941           475 ns/op                     
BenchmarkF1600x4plusFour-8       2585942           467 ns/op                     
BenchmarkF1600x4plusOne-8        2505846           477 ns/op                     
BenchmarkF1600x4plusOne-8        2581089           479 ns/op                     
BenchmarkF1600x4plusOne-8        2647069           463 ns/op                     
BenchmarkF1600x4plusTwo-8        2512466           474 ns/op                     
BenchmarkF1600x4plusTwo-8        2519284           470 ns/op                     
BenchmarkF1600x4plusTwo-8        2531073           478 ns/op
```

bwesterb:
I've tried not unrolling four loops, but simply looping over the rounds

```go
func f1600x4AVX2small() {
	TEXT("f1600x4AVX2small", NOSPLIT, "func(state *[100]uint64, rc *[24]uint64)")

	state_ptr := Load(Param("state"), GP64())
	state := func(offset int) Op {
		return Mem{Base: state_ptr, Disp: 32 * offset}
	}

	rc_ptr := Load(Param("rc"), GP64())

	round := GP64()
	XORQ(round, round)

	Label("loop")

	// Compute parities: p[i] = a[i] ^ a[i + 5] ^ ... ^ a[i + 20].
	p := []Op{YMM(), YMM(), YMM(), YMM(), YMM()}
	for i := 0; i < 5; i++ {
		VMOVDQA(state(i), p[i])
	}
	for j := 1; j < 5; j++ {
		for i := 0; i < 5; i++ {
			VPXOR(state(5*j+i), p[i], p[i])
		}
	}

	// Rotate and xor parities: d[i] = rotate_left(p[i+1], 1) ^ p[i-1]
	t := []Op{YMM(), YMM(), YMM(), YMM(), YMM()}
	d := []Op{YMM(), YMM(), YMM(), YMM(), YMM()}
	for i := 0; i < 5; i++ {
		VPSLLQ(U8(1), p[(i+1)%5], t[i])
	}
	for i := 0; i < 5; i++ {
		VPSRLQ(U8(63), p[(i+1)%5], d[i])
	}
	for i := 0; i < 5; i++ {
		VPOR(t[i], d[i], d[i])
	}
	for i := 0; i < 5; i++ {
		VPXOR(d[i], p[(i+4)%5], d[i])
	}

	// Inverse of the permutation π
	invPi := []int{0, 6, 12, 18, 24, 3, 9, 10, 16, 22, 1, 7,
		13, 19, 20, 4, 5, 11, 17, 23, 2, 8, 14, 15, 21}

	// Appropriate rotations
	rot := []int{0, 44, 43, 21, 14, 28, 20, 3, 45, 61, 1, 6,
		25, 8, 18, 27, 36, 10, 15, 56, 62, 55, 39, 41, 2}

	// Instead of executing ρ, π, χ one at a time writing back the
	// state after each step, we postpone the permutation π to the end
	// (using π⁻¹ χ π instad of χ) and do this in five independant chunks.
	for j := 0; j < 5; j++ {
		s := []Op{YMM(), YMM(), YMM(), YMM(), YMM()}

		// Load the right five words from the state and XOR d into them.
		for i := 0; i < 5; i++ {
			idx := invPi[5*j+i]
			VPXOR(state(idx), d[idx%5], s[i])
		}

		// Rotate each s[i] by the appropriate amount --- this is ρ
		for i := 0; i < 5; i++ {
			cr := rot[5*j+i]
			if cr != 0 {
				VPSLLQ(U8(cr), s[i], t[i])
			}
		}
		for i := 0; i < 5; i++ {
			cr := rot[5*j+i]
			if cr != 0 {
				VPSRLQ(U8(64-cr), s[i], s[i])
			}
		}
		for i := 0; i < 5; i++ {
			if rot[5*j+i] != 0 {
				VPOR(t[i], s[i], s[i])
			}
		}

		// Compute the new words s[i] ^ (s[i+2] & ~s[i+1]) --- this is χ
		for i := 0; i < 5; i++ {
			VPANDN(s[(i+2)%5], s[(i+1)%5], t[i])
		}
		for i := 0; i < 5; i++ {
			VPXOR(s[i], t[i], t[i])
		}

		// Round constant
		if j == 0 {
			rc := YMM()
			VPBROADCASTQ(Mem{Base: rc_ptr, Scale: 8, Index: round}, rc)
			VPXOR(rc, t[0], t[0])
		}

		// Store back into state
		for i := 0; i < 5; i++ {
			VMOVDQA(t[i], state(invPi[j*5+i]))
		}
	}

	// // Finally execute π^{-1}.
	// // XXX optimize
	state1 := YMM()
	tmp1 := YMM()
	last := 1
	VMOVDQA(state(1), state1)
	for j := 0; j < 23; j += 1 {
		VMOVDQA(state(invPi[last]), tmp1)
		VMOVDQA(tmp1, state(last))
		last = invPi[last]
	}
	VMOVDQA(state1, state(10))

	INCQ(round)
	CMPQ(round, Imm(24))
	JNE(LabelRef("loop"))

	RET()
}
```

This is unfortunately slower

```
BenchmarkF1600x4AVX2-8        	 2941527	       415 ns/op
BenchmarkF1600x4AVX2small-8   	 2277976	       531 ns/op
```

(even if I remove the patently unoptimised permutation π⁻¹ at the end.)

internal/shakex4/f1600x4_amd64_src.go
mmcloughlin: Oh awesome, interesting results.

I would still recommend switching to `VMOVQU`, but keep the no-`noescape` technique you are using to get aligned memory from Go. You'll get the performance gain, but it will continue to work if you don't get aligned memory _for whatever reason_.

bwesterb: The no-noescape caused quite a bit of allocator overhead, so instead I opted to allocate an `[103]uint64` and move my offset to ensure 32 byte alignment.

bwesterb:
@armfazh Ready for review.

sign/dilithium/internal/common/poly.go
```diff
@@ -105,26 +105,42 @@ func (p *Poly) MakeHint(p0, p1 *Poly) (pop uint32) {
...
+	// than calling decompose() in a loop (for instance when having AVX2.)
+
+	q.Decompose(&q0PlusQ, p)
+
 	for i := 0; i < N; i++ {
-		p[i] = useHint(q[i], hint[i])
+		if hint[i] == 0 {
+			continue
+		}
+		if q0PlusQ[i] > Q {
```
armfazh, line 120: This could be done in straight-line code.

sign/dilithium/internal/f1600x4/f1600x4_amd64.go
```diff
@@ -0,0 +1,53 @@
...
+}
+
+// Initialize the state and returns the buffer on which the four permutations
+// will act: a uint64 slice of length 100.  The first permutation will act
+// on {a[0], a[4], ..., a[96]}, the second on {a[1], a[5], ..., a[97]}, etc.
+func (s *State) Initialize() []uint64 {
+	rp := unsafe.Pointer(&s.a[0])
+
+	// remainder of address modulo 32
+	rem := (int(uintptr(rp)&31) >> 3)
```
armfazh, line 39: What it does represent the division by 8?

sign/dilithium/internal/f1600x4/f1600x4_test.go
```diff
@@ -0,0 +1,45 @@
...
+)
+
+func TestF1600x4(t *testing.T) {
+	if !Available {
+		t.Skip()
+	}
+	var state State
+	a := state.Initialize()
+
+	expected := []uint64{
```
armfazh, line 14: Add a comment indicating how this was computed.

sign/dilithium/mode3/internal/sample.go
```diff
@@ -2,9 +2,192 @@ package internal
...
+	}
+
+	// Each try requires 15 bits.  15 does not divide into 64, nor in 136,
+	// so we will have to carry some bits from a previous uint64 to the next.
+	var carry [4]uint32
+
+	// Shift is the amount of bits in the carry.
+	var shift [4]uint
+
+	for {
```
armfazh, line 57: Make this loop be more easily to read. 
Also beware that `break` in Go sometimes requires to specify a label.

```suggestion
	for !done { 
```

sign/dilithium/mode3/internal/sample.go
```diff
@@ -2,9 +2,192 @@ package internal
...
+
+// For each i, sample ps[i] uniformly from the given seed and nonces[i].
+// ps[i] may be nil and is ignored in that case.
+//
+// Can only be called when DeriveX4Available is true.
+func PolyDeriveUniformX4(ps [4]*common.Poly, seed *[32]byte, nonces [4]uint16) {
+	var perm f1600x4.State
+	state := perm.Initialize()
+
+	// Absorb the seed in the four states
```
armfazh, line 122: This code looks to be repeated in the previous function, maybe it will go into a small function.

sign/dilithium/mode3/internal/sample.go
```diff
@@ -122,8 +305,32 @@ func PolyDeriveUniformLeqEta(p *common.Poly, seed *[32]byte, nonce uint16) {
...
+	if !DeriveX4Available {
+		for i := 0; i < L; i++ {
+			PolyDeriveUniformLeGamma1(&v[i], seed, nonce+uint16(i))
+		}
+		return
+	}
+
+	// PolyDeriveUniformLeGamma1X4 is slower than, but not twice as slow as,
+	// PolyDeriveUniformLeGamma.
+	if L == 2 {
```
armfazh, line 317: You can construct the input vector iterating in a for-loop with L as bound. The non-assigned values are taken by default to 0 and nil.

sign/dilithium/mode3/internal/sample_test.go
```diff
@@ -263,3 +263,91 @@ func TestDeriveUniformB60(t *testing.T) {
...
+	for i := 0; i < len(seed); i++ {
+		seed[i] = byte(i)
+	}
+
+	PolyDeriveUniformLeGamma1X4([4]*common.Poly{&ps[0], &ps[1], &ps[2], &ps[3]},
+		&seed, nonces)
+	for i := 0; i < 4; i++ {
+		PolyDeriveUniformLeGamma1(&p, &seed, nonces[i])
+		if ps[i] != p {
+			t.Fatalf("%d\n%v\n%v", i, p, ps[i])
```
armfazh, line 308: You may want to use `circl/internal/test.ReportError`

sign/dilithium/internal/common/poly.go
bwesterb: I don't understand your suggestion.  Could you elaborate?

armfazh: I was asking if you can replace the if-then-else by constant-time statements.

bwesterb: For security, there is no requirement that this is constant-time.  The hint is basically a compression technique for the public signature.  Or do you think the constant-time version will perform better?

sign/dilithium/internal/f1600x4/f1600x4_amd64.go
bwesterb: I clarified it.

sign/dilithium/internal/f1600x4/f1600x4_test.go
bwesterb: Done.

sign/dilithium/mode3/internal/sample.go
bwesterb: Thanks, I improved it.

sign/dilithium/mode3/internal/sample.go
bwesterb: It's slightly different: the bound on `i` is 4 in one and 6 in the other

sign/dilithium/mode3/internal/sample.go
bwesterb: Ah, good trick!

sign/dilithium/mode3/internal/sample_test.go
bwesterb: Yeah, I added a reminder on #114 to fix this, not just for this PR but for the existing code as well.

sign/dilithium/internal/f1600x4/f1600x4_amd64_src.go
```diff
@@ -0,0 +1,145 @@
...
+
+			// Store back into state
+			for i := 0; i < 5; i++ {
+				VMOVDQA(t[i], state(si(i, g)))
+			}
+		}
+	}
+
+	ADDQ(Imm(8*4), rc_ptr)
+	DECQ(superRound)
```
dsprenkels, line 139: Small nit: `subq` probably  is preferred here. [Optimization Manual, section 3.5.1.1]

bwesterb: Thanks

sign/dilithium/mode3/internal/sample.go
```diff
@@ -2,9 +2,190 @@ package internal
...
+					((state[(i*3+2)*4+j] & 0x7fff) << 8))
+				t[6] = uint32((state[(i*3+2)*4+j] >> 16) & 0x7fffff)
+				t[7] = uint32((state[(i*3+2)*4+j] >> 40) & 0x7fffff)
+
+				for k := 0; k < 8; k++ {
+					if t[k] < common.Q {
+						ps[j][idx[j]] = t[k]
+						idx[j]++
+						if idx[j] == 256 {
+							break
```
armfazh, line 173: Take a look on "break with labels", it might be possible to write clearer the code.

sign/dilithium/internal/common/poly.go
armfazh: It's ok as is.

sign/dilithium/mode3/internal/sample.go
```diff
@@ -2,9 +2,190 @@ package internal
...
+	// Recall that the rate of SHAKE128 is 168 --- i.e. 21 uint64s.
+	for j := 0; j < 4; j++ {
+		state[4*4+j] = uint64(nonces[j]) | (0x1f << 16)
+		state[20*4+j] = 0x80 << 56
+	}
+
+	var idx [4]int // indices into ps
+	for j := 0; j < 4; j++ {
+		if ps[j] == nil {
+			idx[j] = 256 // mark nil polynomial as completed
```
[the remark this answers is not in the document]
bwesterb: I replaced it with `N`.  There is a reason to actually not replace it: then it signals that the code is not generic in `N` (which it indeed isn't.). I guess I'll add some comments at the definitions of the parameters that these can't simply be changed.

bwesterb:
@armfazh I addressed all issues you raised.

Issue #28, opened 2019-06-05
Consider applying Go TargetSpecific policy


mmcloughlin:
Filippo outlined some guidelines for target specific packages and intends to apply them in the Go standard library.

https://github.com/golang/go/wiki/TargetSpecific
golang/go#31470

This could help with differential testing and fuzzing #4, so it may be good to apply it in `circl` too.

PR #117, opened 2020-05-21
Make f1600x4 public

bwesterb:


bwesterb:
@armfazh

sign/dilithium/mode3/internal/sample.go
```diff
@@ -1,8 +1,8 @@
 package internal
 
 import (
+	"github.com/cloudflare/circl/hash/f1600x4"
```
armfazh, line 4: Renaming the package to some standard name following FIPS202 (e.g. KeccakF1600) will be appreciated.

f1600x4_other.go ->  f1600x4_generic.go
The generic implementation must work too, for example using circl/internal/shake/keccakf.go. 
OTOH: (If we want people don't use the generic implementation, the entire package must be marked as only for amd64.)

Handle the case of AMD64 machines without support for AVX2. -- The `Available` flag delegates this case to the user, who might not handle this situation properly.

Some guidance must be given in the documentation that shows how to effectively load four states interleaved.

hash/f1600x4/asm/go.mod
```diff
@@ -1,4 +1,4 @@
-module github.com/cloudflare/circl/sign/dilithium/internal/f1600x4/asm
+module github.com/cloudflare/circl/hash/f1600x4/asm
```
armfazh, line 1: Relocate `asm` folder to `internal/asm`.

sign/dilithium/mode3/internal/sample.go
bwesterb: > Renaming the package to some standard name following FIPS202 (e.g. KeccakF1600) will be appreciated.

I renamed it to `keccakf1600x4`.

> f1600x4_other.go -> f1600x4_generic.go
> The generic implementation must work too, for example using circl/internal/shake/keccakf.go.
> OTOH: (If we want people don't use the generic implementation, the entire package must be marked as only for amd64.)

Even on amd64, people should not use the package if AVX2 is not available.  I dislike a generic implementation as it's slower (due to transposing.) Also I would like to add an ARM optimized version.

> Handle the case of AMD64 machines without support for AVX2. -- The `Available` flag delegates this case to the user, who might not handle this situation properly.

Same argument as above: if AVX2 is not available a generic implementation would be prohibitively slow.

> Some guidance must be given in the documentation that shows how to effectively load four states interleaved.

I'll add an example.

hash/f1600x4/asm/go.mod
bwesterb: Done.

sign/dilithium/mode3/internal/sample.go
bwesterb: I've added the example.

mmcloughlin: I'd agree with @armfazh here. Ideally there's a generic implementation, doesn't have to be fast, but it works. If not, guard the entire package with `amd64` build tags, so it's literally impossible to build against this package on other platforms.

mmcloughlin:
I also wonder if it's better to name this package `keccakf1600`, and then the types/functions in the package have `X4` on them?

Suppose there were `x1`, `x2`, `x8` variants too... would you expect them to be in the same or different packages?

sign/dilithium/mode3/internal/sample.go
bwesterb: > If not, guard the entire package with `amd64` build tags, so it's literally impossible to build against this package on other platforms.

That doesn't really solve the issue as not every amd64 system has AVX2 to begin with.

bwesterb:
> I also wonder if it's better to name this package `keccakf1600`, and then the types/functions in the package have `X4` on them?
> 
> Suppose there were `x1`, `x2`, `x8` variants too... would you expect them to be in the same or different packages?

It doesn't make sense to have a separate package for plain f[1600].  It does make sense to expose X2 and X8 if available.  I'll rename accordingly.

sign/dilithium/mode3/internal/sample.go
armfazh: It is clear that the generic implementation will be slower, but it should have one because amd64 pre-AVX2 machines still need to be able to compile and use this external package.

It would be great to have build tags like machine+feature, so one can make the package  compiles only for amd64+avx2.

hash/keccakf1600x4/example_test.go
```diff
@@ -0,0 +1,80 @@
...
+		// acts on four interleaved states; that is a [100]uint64.  (A separate
+		// type is used to ensure that the encapsulated [100]uint64 is aligned
+		// properly to be used efficiently with vector instructions.)
+		var perm keccakf1600x4.State
+		state := perm.Initialize()
+
+		// state is initialized with zeroes.  As the messages fit within one
+		// block, we only need to write the messages, domain separators
+		// and padding.
+		for i := 0; i < 4; i++ {
```
armfazh, line 48: This logic can be abstrated by a function beloging to the package. 
It could be a function that receives four pointers and loads them into the internal state.

sign/dilithium/mode3/internal/sample.go
armfazh: Also for the generic implementation, you don't need to perform transpose every time provided you read data from the right place.

bwesterb: > It is clear that the generic implementation will be slower, but it should have one because amd64 pre-AVX2 machines still need to be able to compile and use this external package.

They can use it!  A pre-AVX2 machine can compile this code just fine.  If the use is guarded by the `Available` flag, everything is just fine.  (Dilithium, for instance, runs just fine on my pre-AVX2 MacBook because it uses the `Available` flag to check before using f1600x4.

bwesterb: > Also for the generic implementation, you don't need to perform transpose every time provided you read data from the right place.

But you can't view this function in isolation: for every application there is some kind of transposition going on.  The generic Dilithium code is faster than the code that uses f1600x4 if the latter has a generic implementation.  This is because the latter needs to interleave the data into the state whereas the former doesn't.

hash/keccakf1600x4/example_test.go
bwesterb: But that will be a trap!  I sped up SPHINCS+ by a factor of 2 precisely by removing such a function and doing it ad hoc.

bwesterb: And that is in C.  In Go the difference will be more stark.

https://github.com/sphincs/sphincsplus/pull/9

mmcloughlin:
I'm likely missing context from your team discussions, but I'm wondering why this package is being made public? My feeling is:

* If it's a public package, it should have a generic fallback, regardless of how slow that is. It can be documented that you probably don't want to use it without `amd64+avx2`, but it's very unfriendly if the package crashes without it.
* If it's an `internal` package, code that works on `avx2` only is completely fine.

Generic implementation can also be useful for differential testing #4.

Relevant:
#28 
https://github.com/golang/go/wiki/TargetSpecific

armfazh:
> But you can't view this function in isolation: for every application there is some kind of transposition going on.

The comments of Michael and me comes from the fact that a go package must provide the same functionality regardless the architecture. This is the expected behaviour of a public, single package. 

I understand the benefits of using the vectorized code branch. and I also understand the risk of providing a generic implementation, e.g. some user could think that the permutation is faster even on a pre-AVX2 machine.

However, in a pre-AVX2 machine and in all other architectures, this package doesn't provide any functionality, instead the user has to write code for that (as you did it for dilithium).

I think one point in the middle of this trade-off is to provide an implementation only for AMD64, with two code branches. The vectorized and the scalar one. 
It's clear that performing 4 times the scalar version versus calling 4 times the stdlib function can have different performance, but I guess, it would be small.

bwesterb:
> I'm likely missing context from your team discussions, but I'm wondering why this package is being made public?

I use it in [go-xmssmt](https://github.com/bwesterb/go-xmssmt).  Currently I have a copy of the package there, but I don't it's good to maintain a copy.
 
> * If it's a public package, it should have a generic fallback, regardless of how slow that is. It can be documented that you probably don't want to use it without `amd64+avx2`, but it's very unfriendly if the package crashes without it.

The target audience is very small: it will be used by implementors of cryptographic primitives.  I suspect there might be 10 packages max that will refer to it. I think the panic is actually helpful: it points out a performance bug and wrong use of the package which you might have otherwise missed.

> It's clear that performing 4 times the scalar version versus calling 4 times the stdlib function can have different performance, but I guess, it would be small.

I'll do a benchmark.

bwesterb:
> > It's clear that performing 4 times the scalar version versus calling 4 times the stdlib function can have different performance, but I guess, it would be small.
> 
> I'll do a benchmark.

So I ran the Dilithium public key unpacking (which contains the derivation of the matrix A, which uses a lot of SHAKE) using either the standard SHAKE code or the generic slow f1600x4 code.

The result surprised me: the generic f1600x4 has a non negligible overhead as compared to simple f1600, but it's still faster than the standard SHAKE API.

```
PkUnpack-8  96.7µs ± 1%  88.8µs ± 1%  -8.21%  (p=0.000 n=10+8)
```

bwesterb:
I've renamed the package (again) and I've added a generic fallback.  I would like to protest again: people will use the package without checking the `AvailableX4` which is a performance bug.  Not `panic()`ing is an implicit API: we might break packages in the future if we `panic()` if the fast vectorized versions are not supported.

simd/keccakf1600/f1600x4.go
```diff
@@ -0,0 +1,30 @@
+// f1600x4 implements a fast fourway interleaved KeccaK-f[1600] permutation on
```
armfazh, line 1:
```suggestion
// keccakf1600 implements a fast fourway interleaved Keccak-f[1600] permutation on
```

armfazh: KeccaK -> Keccak

simd/keccakf1600/f1600x4.go
```diff
@@ -0,0 +1,30 @@
...
+// not provide a generic implementation as it would be signficantly slower
+// than using non-interleaved f[1600].  Check keccakf1600.AvailableX4 to see
+// if the current system is supported.
+package keccakf1600
+
+import (
+	"github.com/cloudflare/circl/internal/shake"
+)
+
+func genericF1600x4(a []uint64) {
```
armfazh, line 19: This function should be in the _generic go file.

simd/keccakf1600/f1600x4_amd64.go
```diff
@@ -30,7 +30,7 @@ type State struct {
 // Initialize the state and returns the buffer on which the four permutations
 // will act: a uint64 slice of length 100.  The first permutation will act
 // on {a[0], a[4], ..., a[96]}, the second on {a[1], a[5], ..., a[97]}, etc.
-func (s *State) Initialize() []uint64 {
```
armfazh, line 33: You can follow this pattern:
```
type X [100]uint64
func Initialize() X  { return X{} }
func (X) Permute() {}
```
and use it as:
```
X := keccakf1600.Initialize()
X[0] = 1
X.Permute()
```

simd/keccakf1600/f1600x4_test.go
```diff
@@ -1,14 +1,14 @@
-package f1600x4
+package keccakf1600
 
 import (
 	"testing"
 )
 
 func TestF1600x4(t *testing.T) {
-	if !Available {
+	if !AvailableX4 {
```
armfazh, line 8: should test/bench the generic implementation too.

simd/keccakf1600/internal/asm/go.mod
```diff
@@ -1,4 +1,4 @@
-module github.com/cloudflare/circl/sign/dilithium/internal/f1600x4/asm
+module github.com/cloudflare/circl/hash/keccakf1600x4/internal/asm
```
armfazh, line 1: Is this path correct?

simd/keccakf1600/f1600x4_test.go
bwesterb: I added a test for the generic implementation.

simd/keccakf1600/f1600x4_amd64.go
bwesterb: This doesn't work as `Initialize()` doesn't see the final address of the buffer it returns and so it cannot allign it properly.

simd/keccakf1600/f1600x4.go
bwesterb: No: need to access it if AVX2 is unavailable, and `_generic` is only included on `!amd64`.

simd/keccakf1600/f1600x4_amd64.go
armfazh: I was talking about the API pattern, not about the actual implementation.

bwesterb: Yes, but any possible implementation of your API can't align the buffer correctly.

go.mod
```diff
@@ -2,4 +2,7 @@ module github.com/cloudflare/circl
 
 go 1.12
 
-require golang.org/x/sys v0.0.0-20190602015325-4c4f7f33c9ed
+require (
+	golang.org/x/crypto v0.0.0-20200622213623-75b288015ac9
+	golang.org/x/sys v0.0.0-20190602015325-4c4f7f33c9ed
```
bwesterb, line 7: @armfazh You're ware your change pulled in `/x/crypto`?  I don't think it's a big issue.

simd/keccakf1600/example_test.go
```diff
@@ -0,0 +1,79 @@
+package keccakf1600_test
+
+import (
+	"encoding/binary"
+	"fmt"
+
+	"github.com/cloudflare/circl/simd/keccakf1600"
+	"golang.org/x/crypto/sha3"
```
armfazh, line 8: Apply this change, so we don't need `x/crypto` in go.mod

```suggestion
	sha3 "github.com/cloudflare/circl/internal/shake"
```

PR #119, opened 2020-05-25
Add s390x machine

claucece:
This PR is still WP.

@armfazh , where did you get the sha for the golang image for the other machines?

### DIFF
--- a/.github/workflows/ci-actions.yml
+++ b/.github/workflows/ci-actions.yml
@@ -48,6 +48,22 @@ jobs:
         run: |
           docker run --rm --privileged multiarch/qemu-user-static --reset -p yes
           docker run --rm -v `pwd`:`pwd` -w `pwd` arm64v8/golang@$SHA256 go test -v ./...
+  s390x_job:
+    runs-on: ubuntu-18.04
+    name: s390x/Go-1.14
+    env:
+        GOVER: 1.14
+        SHA256: sha256:213e0cc560e8e93767fdd9d4af890cfc8c11b07b16b34bf1fdf5e548459cb2ad
+    steps:
+      - uses: actions/checkout@v2
+      - name: Pulling Images
+        run: |
+          docker pull -q multiarch/qemu-user-static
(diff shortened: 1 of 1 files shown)


armfazh:
> @armfazh , where did you get the sha for the golang image for the other machines?

I used `arm64v8/golang:1.14.0` however, I looked for that image today and has a different sha value.
Maybe we must move to tags rather than shasum.

.github/workflows/ci-actions.yml
```diff
@@ -48,6 +48,22 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@v2
+      - name: Pulling Images
+        run: |
+          docker pull -q multiarch/qemu-user-static
+          docker pull -q s390x/golang@$SHA256
+      - name: Testing
+        run: |
+          docker run --rm --privileged multiarch/qemu-user-static --reset -p yes
+          docker run --rm -v `pwd`:`pwd` -w `pwd` s390x/golang@$SHA256 go test -v ./...
```
armfazh, line 66: I was investigating what happens with s390x, it seems fetching code causes invalid signature, hence one trick is to vendor dependencies. Include the vendor folder and the following:

```suggestion
          docker run --rm -v `pwd`:`pwd` -w `pwd` s390x/golang@$SHA256 go test -mod=vendor -v ./...
```

claucece:
> I used arm64v8/golang:1.14.0 however, I looked for that image today and has a different sha value.

Yeah, I also saw that. I'll see how it looks with tags ;)

.github/workflows/ci-actions.yml
claucece: Mmm.. not a big fan of vendoring. I'll investigate tomorrow if there is an alternative.. if not, let's use vendor.

armfazh:
@claucece please take a look on #121. I did lots of changes to make it work.
The bug was raised by s390x because it is big endian. In some parts of the code, it always assumes little endian ordering.

claucece:
Closing this PR as the other was merged.

Issue #109, opened 2020-04-30
Make ed25519 and ed448 usable for Golang interfaces


claucece:
Hi!

Thanks so much for all the work on this!

So, the implementation for ed25519 and ed448 expose a certain API that is quite similar to the way golang handles it in its libraries (ecdsa, rsa, and ed25519). However, when I tried to use it for the crypto signer interface, it panics due to:

```
panic: interface conversion: ed25519.PrivateKey is not crypto.Signer: missing method Public [recovered]
	panic: interface conversion: ed25519.PrivateKey is not crypto.Signer: missing method Public
```

I was wondering if we could create a method `func (priv *PrivateKey) Public() crypto.PublicKey` (in similar fashion as https://golang.org/pkg/crypto/ecdsa/#PrivateKey.Public and https://golang.org/pkg/crypto/ed25519/#PrivateKey.Public) so the APIs can be used for other interfaces as the crypto signer...

claucece:
@armfazh I can implement this.. but I want to know if you think it is a good idea ;)

bwesterb:
The `*KeyPair` type seems to implement `crypto.Signer` already, but if that wasn't obvious then we might want to change the API anyways.

claucece:
Yeah, the `KeyPair` type does indeed (and it will work with it). But, as the other APIs (in Golang libraries) do it over the PrivateKey, it seems a little odd. But it can still work ;)

armfazh:
@claucece do you want to proceed with the changes needed?

claucece:
@armfazh I can make a PR changing the API to work with the PrivateKey type... let me know if it is ok, as it can still work with the KeyPair type..

armfazh:
please go forward with that PR. However, I see some trade-offs regarding compatibility. Let me know your feedback.

On the one hand, I would like that circl be a drop-in replacement of stdlib. This implies that the functions `NewKeyFromSeed`, `GenerateKey`, `Sign`,  `Verify` and `PrivateKey` implementing `crypto.signer` must be preserved (but internally using faster circl functions).

On the other hand, having a keypair type allows us to add different flavors for signing e.g. `SignPure`, `SignPh`, and `SignCtx` (as in RFC8032). Currently, only SignPure is supported, and SignPh can be added soon (https://github.com/golang/go/issues/ 31804)

claucece:
So, mmm.. I did a little bit of research around this. On most libraries I checked, there is two interfaces exposed: 

1. signPure
[...]

PR #123, opened 2020-06-04
Add eddsa for prehashed and context use #109

claucece:
This is WIP.

Tasks:

* [x] Add ed25519 prehashed with no context
* [x] Test ed25519 prehashed with no context with rfc8032 and wycheproof vectors (if any)
* [x] Add ed25519 prehashed with context and pure ed25519 with context
* [x] Test ed25519 prehashed with context and pure ed25519 with context with rfc8032 and wycheproof vectors (if any)
* [ ] Add ed448 prehashed with no context
* [ ] Test ed448 prehashed with no context with rfc8032 and wycheproof vectors (if any)
* [ ] Add ed448 prehashed with context and pure ed448 with context
* [ ] Test ed448 prehashed with context and pure ed448 with context with rfc8032 and wycheproof vectors (if any)
* [ ] Consider changing the external API to use the PrivateKey type


sign/ed25519/ed25519.go
```diff
@@ -93,8 +104,10 @@ func NewKeyFromSeed(seed PrivateKey) *KeyPair {
 	return pair
 }
 
+const dom2 = "SigEd25519 no Ed25519 collisions\x01\x00"
+
 // SignPure creates a signature of a message.
-func (kp *KeyPair) SignPure(message []byte) ([]byte, error) {
+func (kp *KeyPair) SignPure(message []byte, preHash bool) ([]byte, error) {
```
armfazh, line 110: I was thinking on having two functions `SignPure` and `SignPh`.

armfazh:
This is a great plan! Let's breakdown in several PRs to increasingly adding functionality and easing reviews. Mark @wbl or me when you think is ready for review.

claucece:
@armfazh For sure! I finish this one only for ed25519 then, and make another PR for ed448. ;)

claucece:
Hi, @armfazh !

So, the first part for ed25519 is done in this PR.

There a couple of things we might further discuss though:

- I have exposed all the Sign functions as the `SignPure` function was originally exposed here, so I thought maybe the same principle applied. But this can change.
- I can unify `Sign` and `SignCtx` into one, by using an struct like this:

```
type Options struct {
	// Hash can be crypto.Hash(0) for Ed25519/Ed25519ctx, or crypto.SHA512
	// for Ed25519ph.
	Hash crypto.Hash

	// Context is an optional domain separation context for Ed25519ph and
	// Ed25519ctx. It must be less than or equal to 255
	//
	Context string
}
```

but I don't know if it will make it confusing for users.

- Unfortunately, the verify functions cannot be unified into one as stdlib does not accept options in their function. But I can create a `Verify` and `VerifyWithOpts` functions that handle the compatibility (the first), and allows options (the second).

Let me know what you think.

claucece:
Adding here @wbl, as well

sign/ed25519/ed25519.go
```diff
@@ -197,9 +188,13 @@ func sign(kp *KeyPair, message []byte, preHash bool, ctx []byte) ([]byte, error)
 // The opts.HashFunc() must return SHA512 to indicate the message has been
 // hashed with SHA512. This can be achieved by passing crypto.SHA512 as the value
 // for opts.
+// Context should be passed to this function, which length should be more than
+// 255.
```
chris-wood, line 192:
```suggestion
// Context should be passed to this function, which length should be no more than
// 255.
```

I think this applies elsewhere where the length of `ctx` is encoded in a single byte, right?

sign/ed25519/ed25519.go
```diff
@@ -347,7 +345,52 @@ func VerifyPh(public PublicKey, message, signature []byte, opts crypto.SignerOpt
...
+		len(signature) != SignatureSize ||
+		!isLessThanOrder(signature[paramB:]) {
+		return false
+	}
+
+	if opts != crypto.Hash(0) {
+		return false
+	}
+
+	if len(ctx) <= 0 || len(ctx) > ContextMaxSize {
```
chris-wood, line 367: Why are zero-length contexts prohibited?

sign/ed25519/ed25519.go
claucece: Oh, yes! That is right. I'll change ;)

claucece: Just fixed it. I also fixed some context checking, as `SignPhCtx` should be able to accept an empty context. The RFC only states: "For Ed25519ph, phflag=1 and PH is SHA512 instead.  That is, the input is hashed using SHA-512 before signing with Ed25519"; but not restriction on the context.. so I assume that an empty string can be passed ;) But let me know if you think otherwise, as it is not the clearest thing on the RFC ;)

Thanks!

armfazh: There are only three variants:

| Name in RFC | Prehash | Valid Context `len(ctx)` | Name in Code |
|-----|-----|-----|-----|
| Ed25519 | No | No - None | SignPure or Sign |
| Ed25519ctx | No | Yes - (1-255) | SignWithCtx |
| Ed25519ph | Yes | Yes - (0-255) | SignPh | 

That leads to the following declarations:
` Sign(message []byte, opts crypto.SignerOpts)  []byte`
` SignWithCtx(message []byte, ctx []byte)  []byte`
` SignPh(message []byte, ctx []byte)  []byte`
and
` Verify(publicKey PublicKey, message, sig []byte) bool`
` VerifyWithCtx(publicKey PublicKey, message, ctx, sig []byte) bool`
` VerifyPh(publicKey PublicKey, message, ctx, sig []byte) bool`

Now, the special case to support the `crypto.Signer` interface can only call to `SignPure` because calling to `SignPh` still requires of passing a context, which is not included in the `crypto.SignerOpts`.

sign/ed25519/ed25519_test.go
```diff
@@ -188,3 +243,30 @@ func Example_ed25519() {
...
+	// Generating Alice's key pair
+	keys, err := ed25519.GenerateKey(rand.Reader)
+	if err != nil {
+		panic("error on generating keys")
+	}
+
+	// Alice signs a message.
+	message := []byte("A message to be signed")
+	opts := crypto.SHA512
+	h := sha512.New()
```
armfazh, line 259: This is not the right example, as the prehashed mode avoids the user hashes the message.

sign/ed25519/ed25519.go
```diff
@@ -131,24 +176,133 @@ func (kp *KeyPair) SignPure(message []byte) ([]byte, error) {
...
+	return sign(kp, message, true, []byte(ctx))
+}
+
+// SignPureCtx creates a signature of a message given a keypair.
+// This function handles unhashed messages with context.
+// The opts.HashFunc() must return zero to indicate the message hasn't been
+// hashed. This can be achieved by passing crypto.Hash(0) as the value for opts.
+// Context should be passed to this function, which length should be no more than
+// 255. It should not be empty.
+func (kp *KeyPair) SignPureCtx(message []byte, opts crypto.SignerOpts, ctx string) ([]byte, error) {
```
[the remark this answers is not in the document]
claucece: yes, it is. But in the golang proposal (https://github.com/golang/go/issues/31804#issuecomment-537243120) they decided to go for string as well. I kept it as an string, as well to be compatible to their future changes... But, if we don't want that, we can change it here for []byte ;)

sign/ed25519/ed25519.go
claucece: Thanks for this @armfazh ! So, the reason why the `crypto.Signer` interface one calls now both to `SignPure` and `SignPh` is due to the golang proposal where this was what the accepted api will look like. I can also used the method described here: https://github.com/golang/go/issues/31804#issuecomment-537230419 , and then the `crypto.Signer`  can accept context.

sign/ed25519/ed25519_test.go
claucece: Oh, really? It was my understanding that it expected the input to be  hashed using SHA-512 before signing with Ed25519: "For Ed25519ph, phflag=1 and PH is SHA512 instead.  That is, the input is hashed using SHA-512 before signing with Ed25519". But maybe it is not the user that manually hashes it.. is that what you mean? The example can work as well with out the hash but I thought it made it clearer.. what do you think?

sign/ed25519/ed25519.go
```diff
@@ -22,6 +23,9 @@ const (
 	PrivateKeySize = 32
 	// SignatureSize is the length in bytes of signatures.
 	SignatureSize = 64
+
+	// ContextMaxSize is the maximum allowed context length.
```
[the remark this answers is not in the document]
claucece: Solved ;)

sign/ed25519/ed25519.go
```diff
@@ -131,24 +176,133 @@ func (kp *KeyPair) SignPure(message []byte) ([]byte, error) {
...
+// The opts.HashFunc() must return zero to indicate the message hasn't been
+// hashed. This can be achieved by passing crypto.Hash(0) as the value for opts.
+// Context should be passed to this function, which length should be no more than
+// 255. It should not be empty.
+func (kp *KeyPair) SignPureCtx(message []byte, opts crypto.SignerOpts, ctx string) ([]byte, error) {
+	if opts != crypto.Hash(0) {
+		return nil, errors.New("ed25519: incorrect message for non prehashed signing")
+	}
+
+	if !(len(ctx) > 0) || len(ctx) > ContextMaxSize {
```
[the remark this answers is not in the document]
claucece: Solved ;)

sign/ed25519/ed25519_test.go
armfazh: > But maybe it is not the user that manually hashes it.. is that what you mean?

yes

sign/ed25519/ed25519.go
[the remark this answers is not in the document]
armfazh: ok

sign/ed25519/ed25519_test.go
claucece: My bad: the routine actually fails if the message is not hashed (as it also happens on the RFC test cases). I'll add a note that says 'the message is hashed'.

armfazh: I believe there is a misconception, I added a comment ([here](https://github.com/golang/go/issues/31804#issuecomment-653405864)).

claucece: I see what you mean now. Mmm.. hashing internally might be the right way to go for two reasons:

1. The function can't know if the message was hashed or not, and there is no way of knowing
2. The user should not need to manually hash, as it will create confusion for them

I also read 'the input is hashed using SHA-512 before signing with Ed25519.' on those terms now.

Alright, I'll change the functions ;)

claucece: Should be fixed now.

claucece:
Ok, @armfazh should be good now. I added a multipurpose signing and verifying function. Let me know what you think.

sign/ed25519/ed25519.go
```diff
@@ -22,11 +24,33 @@ const (
...
 )
 const (
 	paramB = 256 / 8 // Size of keys in bytes.
 )
 
+// Options are the options that the ed25519 functions can take.
+type Options struct {
+	// Hash can be crypto.Hash(0) for Ed25519/Ed25519ctx, or crypto.SHA512
+	// for Ed25519ph.
+	Hash crypto.Hash
```
armfazh, line 39:
```suggestion
	crypto.Hash
```

sign/ed25519/ed25519.go
```diff
@@ -22,11 +24,33 @@ const (
...
+}
+
+// HashFunc returns an identifier for the hash function to be used
+// to hash the message.
+// For the Ed25519 and  Ed25519ctx this must be crypto.Hash(0); for Ed25519ph
+// this must be crypto.SHA512.
+func (opt *Options) HashFunc() crypto.Hash {
+	return opt.Hash
+}
+
```
armfazh, line 53:
```suggestion

```

sign/ed25519/ed25519.go
```diff
@@ -52,15 +76,32 @@ func (kp *KeyPair) Seed() []byte { return kp.GetPrivate() }
...
-// Ed25519 performs two passes over messages to be signed and therefore cannot
-// handle pre-hashed messages.
+// This function can handle unhashed messages or messages that will be
+// prehashed with SHA512. It also handles context.
 // The opts.HashFunc() must return zero to indicate the message hasn't been
 // hashed. This can be achieved by passing crypto.Hash(0) as the value for opts.
+// The opts.HashFunc() must return SHA512 to indicate that the message will be
+// hashed with SHA512. This can be achieved by passing crypto.SHA512 as the value
+// for opts.
+// Messages to be prehashed with other algorithms are not handled.
```
armfazh, line 86: Replace the doc string by this shorter one.

```suggestion
// This function supports all the three signature variants defined in RFC-8032, namely Ed25519 (or pure EdDSA), Ed25519Ph, and Ed25519Ctx.
// The context is passed inside a Options struct defined in this package.
```

armfazh: I think specifying Hash(0) or SHA-512 inside ops is useless, since the hash function is fixed.
However, let's keep it (in this function) as is. And let's follow how upstream stdlib solves this API.

sign/ed25519/ed25519.go
```diff
@@ -93,8 +134,9 @@ func NewKeyFromSeed(seed PrivateKey) *KeyPair {
 	return pair
 }
 
-// SignPure creates a signature of a message.
-func (kp *KeyPair) SignPure(message []byte) ([]byte, error) {
+const dom2 = "SigEd25519 no Ed25519 collisions"
```
armfazh, line 137: Move this variable inside ~defineDom~ writeDom.

sign/ed25519/ed25519.go
```diff
@@ -104,6 +146,9 @@ func (kp *KeyPair) SignPure(message []byte) ([]byte, error) {
 
 	// 2.  Compute SHA-512(dom2(F, C) || prefix || PH(M))
 	H.Reset()
+
+	H = defineDom(H, ctx, preHash)
```
armfazh, line 150:
```suggestion
	writeDom(H, ctx, preHash)
```

sign/ed25519/ed25519.go
```diff
@@ -131,24 +180,76 @@ func (kp *KeyPair) SignPure(message []byte) ([]byte, error) {
...
-func Verify(public PublicKey, message, signature []byte) bool {
-	if len(public) != PublicKeySize ||
-		len(signature) != SignatureSize ||
-		!isLessThanOrder(signature[paramB:]) {
-		return false
+// SignPure creates a signature of a message given a keypair.
+// This function handles unhashed messages.
+// The opts.HashFunc() must return zero to indicate the message hasn't been
+// hashed. This can be achieved by passing crypto.Hash(0) as the value for opts.
+func (kp *KeyPair) SignPure(message []byte, opts crypto.SignerOpts) ([]byte, error) {
```
armfazh, line 191: The API for SignPure, SignPh, SignWithCtx (and its verfication methods) do not need options. Because every one asks only for the parameters it requires.

sign/ed25519/ed25519.go
```diff
@@ -131,24 +180,76 @@ func (kp *KeyPair) SignPure(message []byte) ([]byte, error) {
...
+// Messages to be prehashed with other algorithms are not handled.
+// Context could be passed to this function, which length should be no more than
+// 255. It can be empty.
+func (kp *KeyPair) SignPh(message []byte, opts crypto.SignerOpts, ctx string) ([]byte, error) {
+	if opts.HashFunc() != crypto.SHA512 {
+		return nil, errors.New("ed25519: incorrect message for prehashed signing")
+	}
+
+	if len(ctx) > ContextMaxSize {
+		return nil, errors.New("ed25519: bad context length: " + strconv.Itoa(len(ctx)))
```
armfazh, line 214:
```suggestion
		return nil, fmt.Errorf("ed25519: bad context length: %v > %v", len(ctx), ContentMaxSize)
```

sign/ed25519/ed25519.go
```diff
@@ -131,24 +180,76 @@ func (kp *KeyPair) SignPure(message []byte) ([]byte, error) {
...
+	}
+
+	h := sha512.New()
```
armfazh, line 217: Move this pre-hashing inside `sign`, so you can reuse the object H.

sign/ed25519/ed25519.go
```diff
@@ -163,6 +264,98 @@ func Verify(public PublicKey, message, signature []byte) bool {
 	return bytes.Equal(R, encR)
 }
 
+// Verify returns true if the signature is valid. Failure cases are invalid
+// signature, or when the public key cannot be decoded.
+// This function does not handle prehashed messages.
+func Verify(public PublicKey, message, signature []byte) bool {
```
armfazh, line 270: Let's match the name of these functions as follows:  
`Sign`  -  `Verify`
`SignPure`  -  `VerifyPure`
`SignPh`  -  `VerifyPh`
`SignWithCtx`  -  `VerifyWithCtx`

sign/ed25519/ed25519.go
```diff
@@ -163,6 +264,98 @@ func Verify(public PublicKey, message, signature []byte) bool {
...
+	if len(public) != PublicKeySize ||
+		len(signature) != SignatureSize ||
+		!isLessThanOrder(signature[paramB:]) {
+		return false
+	}
```
armfazh, line 275: These checks can go inside `verify`

sign/ed25519/ed25519.go
```diff
@@ -176,3 +369,20 @@ func isLessThanOrder(x []byte) bool {
 	}
 	return x[i] < order[i]
 }
+
+func defineDom(h hash.Hash, ctx []byte, preHash bool) hash.Hash {
```
armfazh, line 373:
```suggestion
func defineDom(h io.Writer, ctx []byte, preHash bool) {
```

sign/ed25519/ed25519_test.go
```diff
@@ -188,3 +344,53 @@ func Example_ed25519() {
 	fmt.Println(ok)
 	// Output: true
 }
+
+func Example_ed25519Ph() {
+	// import "github.com/cloudflare/circl/sign/ed25519"
```
armfazh, line 349:
```suggestion
	// import "github.com/cloudflare/circl/sign/ed25519"
	// import "crypto/rand"
```

claucece:
Done @armfazh . I took the current commits of master branch as well, just in case. I noticed some problems with the linting directive of the project.. I'll fix it in another PR.

sign/ed25519/ed25519_test.go
```diff
@@ -119,7 +235,7 @@ func TestErrors(t *testing.T) {
 		ops := crypto.SHA224
 		key, _ := ed25519.GenerateKey(nil)
 		_, got := key.Sign(nil, msg[:], ops)
-		want := errors.New("ed25519: cannot sign hashed message")
+		want := errors.New("ed25519: expected unhashed message or message to be hashed with SHA-512")
```
armfazh, line 238:
```suggestion
		want := errors.New("ed25519: bad hash algorithm")
```

sign/ed25519/ed25519.go
```diff
@@ -52,15 +68,31 @@ func (kp *KeyPair) Seed() []byte { return kp.GetPrivate() }
...
+	case crypto.SHA512:
+		return kp.SignPh(message, ctx)
+	case crypto.Hash(0):
+		if len(ctx) > 0 {
+			return kp.SignWithCtx(message, ctx)
+		}
+
+		return kp.SignPure(message)
+	default:
+		return nil, errors.New("ed25519: expected unhashed message or message to be hashed with SHA-512")
```
armfazh, line 94:
```suggestion
		return nil, errors.New("ed25519: bad hash algorithm")
```

sign/ed25519/ed25519.go
```diff
@@ -163,6 +258,64 @@ func Verify(public PublicKey, message, signature []byte) bool {
...
+// 255. It can be empty.
+func VerifyPh(public PublicKey, message, signature []byte, ctx string) bool {
+	return verify(public, message, signature, true, []byte(ctx))
+}
+
+// VerifyWithCtx returns true if the signature is valid. Failure cases are invalid
+// signature, or when the public key cannot be decoded, or when context is
+// not provided.
+// This function supports the signature variant defined in RFC-8032: Ed25519ctx,
+// meaning it does not handle prehashed messages. Context string must be
```
armfazh, line 309:
```suggestion
// meaning it does not handle prehashed messages. Non-empty context string must be
```

sign/ed25519/ed25519.go
```diff
@@ -163,6 +258,64 @@ func Verify(public PublicKey, message, signature []byte) bool {
...
+// This function does not handle prehashed messages.
+func VerifyPure(public PublicKey, message, signature []byte) bool {
+	ctx := ""
+	return verify(public, message, signature, false, []byte(ctx))
+}
+
+// VerifyPh returns true if the signature is valid. Failure cases are invalid
+// signature, or when the public key cannot be decoded.
+// This function supports the signature variant defined in RFC-8032: Ed25519ph,
+// meaning it handles messages to be prehashed with SHA512.
```
armfazh, line 298:
```suggestion
// meaning it internally hashes the message using SHA-512.
```

sign/ed25519/ed25519.go
```diff
@@ -131,27 +179,74 @@ func (kp *KeyPair) SignPure(message []byte) ([]byte, error) {
...
+// This function supports the signature variant defined in RFC-8032: Ed25519,
+// meaning it handles messages that will not be prehashed, with no context.
+func (kp *KeyPair) SignPure(message []byte) ([]byte, error) {
+	ctx := ""
+	return sign(kp, message, false, []byte(ctx))
+}
+
+// SignPh creates a signature of a message given a keypair.
+// This function supports the signature variant defined in RFC-8032: Ed25519ph,
+// meaning it handles messages to be prehashed with SHA512.
```
armfazh, line 196:
```suggestion
// meaning it internally hashes the message using SHA-512.
```

sign/ed25519/ed25519.go
```diff
@@ -163,6 +258,64 @@ func Verify(public PublicKey, message, signature []byte) bool {
...
+
+		return VerifyPure(public, message, signature)
+	default:
+		return false
+	}
+}
+
+// VerifyPure returns true if the signature is valid. Failure cases are invalid
+// signature, or when the public key cannot be decoded.
+// This function does not handle prehashed messages.
```
armfazh, line 289:
```suggestion
```

sign/ed25519/ed25519.go
```diff
@@ -131,27 +179,74 @@ func (kp *KeyPair) SignPure(message []byte) ([]byte, error) {
...
+
 	return signature[:], err
 }
 
-// Verify returns true if the signature is valid. Failure cases are invalid
-// signature, or when the public key cannot be decoded.
-func Verify(public PublicKey, message, signature []byte) bool {
+// SignPure creates a signature of a message given a keypair.
+// This function supports the signature variant defined in RFC-8032: Ed25519,
+// meaning it handles messages that will not be prehashed, with no context.
```
armfazh, line 188:
```suggestion
// also known as the pure version of EdDSA.
```

sign/ed25519/ed25519.go
```diff
@@ -52,15 +68,31 @@ func (kp *KeyPair) Seed() []byte { return kp.GetPrivate() }
...
 // Sign creates a signature of a message given a key pair.
-// Ed25519 performs two passes over messages to be signed and therefore cannot
-// handle pre-hashed messages.
+// This function supports all the three signature variants defined in RFC-8032,
+// namely Ed25519 (or pure EdDSA), Ed25519Ph, and Ed25519Ctx.
 // The opts.HashFunc() must return zero to indicate the message hasn't been
 // hashed. This can be achieved by passing crypto.Hash(0) as the value for opts.
+// The opts.HashFunc() must return SHA512 to indicate that the message will be
+// hashed with SHA512. This can be achieved by passing crypto.SHA512 as the value
+// for opts.
```
armfazh, line 77:
```suggestion
// The opts.HashFunc() must return zero to specify either Ed25519 or Ed25519Ctx 
// variant. This can be achieved by passing crypto.Hash(0) as the value for opts.
// The opts.HashFunc() must return SHA512 to specify the Ed25519Ph variant.
// This can be achieved by passing crypto.SHA512 as the value for opts.
// Use an Options struct to pass a context string for signing.
```

sign/ed25519/ed25519.go
```diff
@@ -22,11 +24,25 @@ const (
...
 )
 
+// Options are the options that the ed25519 functions can take.
+type Options struct {
+	// Hash can be crypto.Hash(0) for Ed25519/Ed25519ctx, or crypto.SHA512
+	// for Ed25519ph.
+	crypto.Hash
+
+	// Context is an optional domain separation string for Ed25519ph and a
+	// must for Ed25519ctx. It must be less than or equal to 255.
```
armfazh, line 42:
```suggestion
	// must for Ed25519ctx. Its length must be less or equal than 255 bytes.
```

claucece:
@armfazh Thanks! I just fixed it. I'll open another PR for ed448, following this API ;)

PR #125, opened 2020-06-18
Dilithium: MulBy2toD, UnpackLeGamma1, DeriveUniformB60

bwesterb:
(AVX2) optimised versions of

1. MulBy2toD
2. UnpackLeGamma1
3. **DeriveUniformB60**. To be honest, I implemented this fourway function by accident: DeriveUniformB60 is only used for a single polynomial — not for a vector.  However, in the future we might try to share/parallelise computation between attempts and then this function is useful again.  I think on its own it's not worth batching four attempts at the same time yet.  But, let's add it anyway.


bwesterb:
> Could you please give a brief of waht these functions do.

They're clearly documented in the code.

> It's worth if the cost of doing this is parallel is close (or better) to their sequential counterpart.

It's not that simple in this case, because the number of times DeriveUniformB60 is called is not fixed, but varies.  (It's called once for each attempted signature and there are an average of 7 attempts.)

PR #126, opened 2020-06-25
eddilithium3: add hybrid of Dilithium3 and Ed25519

bwesterb:
I'll add EdDilithium4 (which would be Dilithium4 + Ed448) when we're happy with this one.

### DIFF
--- a/sign/dilithium/dilithium.go
+++ b/sign/dilithium/dilithium.go
@@ -6,13 +6,24 @@
 // https://pq-crystals.org/dilithium/data/dilithium-specification-round2.pdf
 //
 // Each of the eight different modes of Dilithium is implemented by a
-// subpackage.  For instance, Dilithium III can be found in
+// subpackage.  For instance, Dilithium III (the recommended mode)
+// can be found in
 //
 //  github.com/cloudflare/circl/sign/dilithium/mode3
 //
 // If your choice for mode is fixed compile-time, use the subpackages.
 // This package provides a convenient wrapper around all of the subpackages
 // so one can be chosen at runtime.
+//
+// The authors of Dilithium recommend to combine it with a "pre-quantum"
+// signature scheme.  The packages
+//
+//  github.com/cloudflare/circl/sign/eddilithium3
+//  github.com/cloudflare/circl/sign/eddilithium4
+//
+// implement such hybrids of Dilithium3 with Ed25519 respectively and
+// Dilithium4 with Ed448.  These packages are a drop in replacements for the
+// mode subpackages of this package.
 package dilithium
 
 import (
--- a/sign/dilithium/templates/mode.templ.go
+++ b/sign/dilithium/templates/mode.templ.go
@@ -7,7 +7,6 @@
 package dilithium
 
 import (
-	"errors"
 	"fmt"
 	"io"
 
@@ -21,8 +20,8 @@ type {{ .Impl }} struct{}
 // {{ .Mode }} is Dilithium in mode "{{ .Name }}".
 var {{ .Mode }} Mode = &{{ .Impl }}{}
 
-func (m *{{ .Impl }}) GenerateKey(rand io.Reader) (PublicKey,
-	PrivateKey, error) {
+func (m *{{ .Impl }}) GenerateKey(rand io.Reader) (
+	PublicKey, PrivateKey, error) {
 	return {{ .Pkg }}.GenerateKey(rand)
 }
 
@@ -56,7 +55,7 @@ func (m *{{ .Impl }}) Verify(pk PublicKey, msg []byte, signature []byte) bool {
 func (m *{{ .Impl }}) PublicKeyFromBytes(data []byte) PublicKey {
 	var ret {{ .Pkg }}.PublicKey
 	if len(data) != {{ .Pkg }}.PublicKeySize {
-		panic(errors.New("packed public key must be of {{ .Pkg }}.PublicKeySize bytes"))
+		panic("packed public key must be of {{ .Pkg }}.PublicKeySize bytes")
 	}
 	var buf [{{ .Pkg }}.PublicKeySize]byte
 	copy(buf[:], data)
@@ -67,7 +66,7 @@ func (m *{{ .Impl }}) PublicKeyFromBytes(data []byte) PublicKey {
 func (m *{{ .Impl }}) PrivateKeyFromBytes(data []byte) PrivateKey {
 	var ret {{ .Pkg }}.PrivateKey
 	if len(data) != {{ .Pkg }}.PrivateKeySize {
-		panic(errors.New("packed public key must be of {{ .Pkg }}.PrivateKeySize bytes"))
+		panic("packed public key must be of {{ .Pkg }}.PrivateKeySize bytes")
 	}
 	var buf [{{ .Pkg }}.PrivateKeySize]byte
 	copy(buf[:], data)
--- a/sign/eddilithium3/eddilithium.go
+++ b/sign/eddilithium3/eddilithium.go
@@ -1,0 +1,218 @@
+// eddilithium3 implements the hybrid signature scheme Ed25519-Dilithium3.
+package eddilithium3
+
+import (
+	"crypto"
+	cryptoRand "crypto/rand"
+	"errors"
+	"io"
+
+	"github.com/cloudflare/circl/internal/shake"
+	"github.com/cloudflare/circl/sign/dilithium/mode3"
+	"github.com/cloudflare/circl/sign/ed25519"
+)
+
+const (
+	// Size of seed for NewKeyFromSeed
+	SeedSize = mode3.SeedSize // = ed25519.SeedSize = 32
+
+	// Size of a packed PublicKey
+	PublicKeySize = mode3.PublicKeySize + ed25519.PublicKeySize
+
+	// Size of a packed PrivateKey
+	PrivateKeySize = mode3.PrivateKeySize + ed25519.PrivateKeySize
+
+	// Size of a signature
+	SignatureSize = mode3.SignatureSize + ed25519.SignatureSize
+)
+
+// PublicKey is the type of an EdDilithium3 public key.
+type PublicKey struct {
+	e ed25519.PublicKey
+	d mode3.PublicKey
+}
+
+// PrivateKey is the type of an EdDilithium3 private key.
+type PrivateKey struct {
+	e ed25519.KeyPair
+	d mode3.PrivateKey
+}
+
+// GenerateKey generates a public/private key pair using entropy from rand.
+// If rand is nil, crypto/rand.Reader will be used.
+func GenerateKey(rand io.Reader) (*PublicKey, *PrivateKey, error) {
+	var seed [SeedSize]byte
+	if rand == nil {
+		rand = cryptoRand.Reader
+	}
+	_, err := io.ReadFull(rand, seed[:])
+	if err != nil {
+		return nil, nil, err
+	}
+
+	pk, sk := NewKeyFromSeed(&seed)
+	return pk, sk, nil
+}
+
+// NewKeyFromSeed derives a public/private key pair using the given seed.
+func NewKeyFromSeed(seed *[SeedSize]byte) (*PublicKey, *PrivateKey) {
+	var seed1 [32]byte
+	var seed2 [32]byte
+
+	// Internally, Ed25519 and Dilithium hash the seeds they are passed again
+	// with different hash functions, so it would be safe to use exactly the
+	// same seed for Ed25519 and Dilithium here.  However, in general, when
+	// combining any two signature schemes it might not be the case that this
+	// is safe.  Setting a bad example here isn't worth the tiny gain in
+	// performance.
+
+	h := shake.NewShake256()
+	_, _ = h.Write(seed[:])
+	_, _ = h.Read(seed1[:])
+	_, _ = h.Read(seed2[:])
+	dpk, dsk := mode3.NewKeyFromSeed(&seed1)
+	epair := ed25519.NewKeyFromSeed(seed2[:])
+
+	return &PublicKey{epair.GetPublic(), *dpk}, &PrivateKey{*epair, *dsk}
+}
+
+// SignTo signs the given message and writes the signature into signature.
+// It will panic if signature is not of length at least SignatureSize.
+func SignTo(sk *PrivateKey, msg []byte, signature []byte) {
+	mode3.SignTo(
+		&sk.d,
+		msg,
+		signature[:mode3.SignatureSize],
+	)
+	esig, _ := sk.e.SignPure(
+		msg,
+	)
+	copy(signature[mode3.SignatureSize:], esig[:])
+}
+
+// Verify checks whether the given signature by pk on msg is valid.
+func Verify(pk *PublicKey, msg []byte, signature []byte) bool {
+	if !mode3.Verify(
+		&pk.d,
+		msg,
+		signature[:mode3.SignatureSize],
+	) {
+		return false
+	}
+	if !ed25519.Verify(
+		pk.e,
+		msg,
+		signature[mode3.SignatureSize:],
+		crypto.Hash(0),
+	) {
+		return false
+	}
+	return true
+}
+
+// Unpack unpacks pk to the public key encoded in buf.
+func (pk *PublicKey) Unpack(buf *[PublicKeySize]byte) {
+	var tmp [mode3.PublicKeySize]byte
+	copy(tmp[:], buf[:mode3.PublicKeySize])
+	pk.d.Unpack(&tmp)
+	pk.e = make([]byte, ed25519.PublicKeySize)
+	copy(pk.e, buf[mode3.PublicKeySize:])
+}
+
+// Unpack sets sk to the private key encoded in buf.
+func (sk *PrivateKey) Unpack(buf *[PrivateKeySize]byte) {
+	var tmp [mode3.PrivateKeySize]byte
+	copy(tmp[:], buf[:mode3.PrivateKeySize])
+	sk.d.Unpack(&tmp)
+	sk.e = *ed25519.NewKeyFromSeed(buf[mode3.PrivateKeySize:])
+}
+
+// Pack packs the public key into buf.
+func (pk *PublicKey) Pack(buf *[PublicKeySize]byte) {
+	var tmp [mode3.PublicKeySize]byte
+	pk.d.Pack(&tmp)
+	copy(buf[:mode3.PublicKeySize], tmp[:])
+	copy(buf[mode3.PublicKeySize:], pk.e)
+}
+
+// Pack packs the private key into buf.
+func (sk *PrivateKey) Pack(buf *[PrivateKeySize]byte) {
+	var tmp [mode3.PrivateKeySize]byte
+	sk.d.Pack(&tmp)
+	copy(buf[:mode3.PrivateKeySize], tmp[:])
+	copy(buf[mode3.PrivateKeySize:], sk.e.Seed())
+}
+
+// Bytes packs the public key.
+func (pk *PublicKey) Bytes() []byte {
+	return append(pk.d.Bytes(), pk.e...)
+}
+
+// Bytes packs the private key.
+func (sk *PrivateKey) Bytes() []byte {
+	return append(sk.d.Bytes(), sk.e.Seed()...)
+}
+
+// MarshalBinary packs the public key.
+func (pk *PublicKey) MarshalBinary() ([]byte, error) {
+	return pk.Bytes(), nil
+}
+
+// MarshalBinary packs the private key.
+func (sk *PrivateKey) MarshalBinary() ([]byte, error) {
+	return sk.Bytes(), nil
+}
+
+// UnmarshalBinary the public key from data.
+func (pk *PublicKey) UnmarshalBinary(data []byte) error {
+	if len(data) != PublicKeySize {
+		return errors.New("packed public key must be of eddilithium3.PublicKeySize bytes")
+	}
+	var buf [PublicKeySize]byte
+	copy(buf[:], data)
+	pk.Unpack(&buf)
+	return nil
+}
+
+// UnmarshalBinary unpacks the private key from data.
+func (sk *PrivateKey) UnmarshalBinary(data []byte) error {
+	if len(data) != PrivateKeySize {
+		return errors.New("packed private key must be of eddilithium3.PrivateKeySize bytes")
+	}
+	var buf [PrivateKeySize]byte
+	copy(buf[:], data)
+	sk.Unpack(&buf)
+	return nil
+}
+
+// Sign signs the given message.
+//
+// opts.HashFunc() must return zero, which can be achieved by passing
+// crypto.Hash(0) for opts.  rand is ignored.  Will only return an error
+// if opts.HashFunc() is non-zero.
+//
+// This function is used to make PrivateKey implement the crypto.Signer
+// interface.  The package-level SignTo function might be more convenient
+// to use.
+func (sk *PrivateKey) Sign(rand io.Reader, msg []byte, opts crypto.SignerOpts) (
+	signature []byte, err error) {
+	var sig [SignatureSize]byte
+
+	if opts.HashFunc() != crypto.Hash(0) {
+		return nil, errors.New("eddilithium3: cannot sign hashed message")
+	}
+
+	SignTo(sk, msg, sig[:])
+	return sig[:], nil
+}
+
+// Public computes the public key corresponding to this private key.
+//
+// Returns a *PublicKey.  The type crypto.PublicKey is used to make
+// PrivateKey implement the crypto.Signer interface.
+func (sk *PrivateKey) Public() crypto.PublicKey {
+	return &PublicKey{
+		sk.e.GetPublic(),
+		*sk.d.Public().(*mode3.PublicKey),
+	}
+}
--- a/sign/eddilithium3/eddilithium_test.go
+++ b/sign/eddilithium3/eddilithium_test.go
@@ -1,0 +1,111 @@
+package eddilithium3_test
+
+import (
+	"github.com/cloudflare/circl/sign/eddilithium3"
+
+	"encoding/binary"
+	"testing"
+)
+
+func BenchmarkVerify(b *testing.B) {
+	// Note that Dilithium precomputes quite a bit during Unpacking/Keygen
+	// instead of at the moment of verification (as compared to the reference
+	// implementation.  A fair comparison thus should sum verification
+	// times with unpacking times.)
+	var seed [32]byte
+	var msg [8]byte
+	var sig [eddilithium3.SignatureSize]byte
+	pk, sk := eddilithium3.NewKeyFromSeed(&seed)
+	eddilithium3.SignTo(sk, msg[:], sig[:])
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
+		// We should generate a new signature for every verify attempt,
+		// as this influences the time a little bit.  This difference, however,
+		// is small and generating a new signature in between creates a lot
+		// pressure on the allocator which makes an accurate measurement hard.
+		eddilithium3.Verify(pk, msg[:], sig[:])
+	}
+}
+
+func BenchmarkSign(b *testing.B) {
+	// Note that Dilithium precomputes quite a bit during Unpacking/Keygen
+	// instead of at the moment of signing (as compared to the reference
+	// implementation.  A fair comparison thus should sum sign times with
+	// unpacking times.)
+	var seed [32]byte
+	var msg [8]byte
+	var sig [eddilithium3.SignatureSize]byte
+	_, sk := eddilithium3.NewKeyFromSeed(&seed)
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
+		binary.LittleEndian.PutUint64(msg[:], uint64(i))
+		eddilithium3.SignTo(sk, msg[:], sig[:])
+	}
+}
+
+func BenchmarkGenerateKey(b *testing.B) {
+	var seed [32]byte
+	for i := 0; i < b.N; i++ {
+		binary.LittleEndian.PutUint64(seed[:], uint64(i))
+		eddilithium3.NewKeyFromSeed(&seed)
+	}
+}
+
+func BenchmarkPublicFromPrivate(b *testing.B) {
+	var seed [32]byte
+	for i := 0; i < b.N; i++ {
+		b.StopTimer()
+		binary.LittleEndian.PutUint64(seed[:], uint64(i))
+		_, sk := eddilithium3.NewKeyFromSeed(&seed)
+		b.StartTimer()
+		sk.Public()
+	}
+}
+
+func TestSignThenVerifyAndPkSkPacking(t *testing.T) {
+	var seed [eddilithium3.SeedSize]byte
+	var sig [eddilithium3.SignatureSize]byte
+	var msg [8]byte
+	var pkb1, pkb2 [eddilithium3.PublicKeySize]byte
+	var skb1, skb2 [eddilithium3.PrivateKeySize]byte
+	var pk2 eddilithium3.PublicKey
+	var sk2 eddilithium3.PrivateKey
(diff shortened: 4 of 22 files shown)


bwesterb:
@claucece

claucece:
I'll take a look later today @bwesterb ;) Thanks for this!

sign/dilithium/gen.go
```diff
@@ -139,6 +143,9 @@ var (
 			DoubleEtaBits:  3,
 			Beta:           175,
 			Omega:          120,
+		}, {
+			Name:   "EdDilithium3",
```
bwesterb, line 147: Probably to  preempt your question, @armfazh, I guess you'd prefer Edwards25519-Dilithium3 here?

armfazh: why is needed a new implementation of Mode3?

bwesterb: It's not a full new implementation.  This extra line only tells `gen.go` to generate the file that registers a new mode.

sign/dilithium/dilithium.go
```diff
@@ -6,13 +6,20 @@
...
-// subpackge.  For instance, Dilithium III can be found in
+// subpackage.  For instance, Dilithium III can be found in
 //
 //  github.com/cloudflare/circl/sign/dilithium/mode3
 //
 // If your choice for mode is fixed compile-time, use the subpackages.
 // This package provides a convenient wrapper around all of the subpackages
 // so one can be chosen at runtime.
+//
+// This package also implements two hybrid modes: EdDilithium3 and EdDilithium4.
```
armfazh, line 17: This package implements only dilithium, no more than that.
I don't see the need to include the hybrid mode here.

sign/dilithium/gen.go
armfazh: The hybrid mode is external to the scope of this package. There is no need to mantain a register of external implementations.

sign/eddilithium3/eddilithium.go
```diff
@@ -0,0 +1,210 @@
...
+	if err != nil {
+		return nil, nil, err
+	}
+
+	pk, sk := NewKeyFromSeed(&seed)
+	return pk, sk, nil
+}
+
+// NewKeyFromSeed derives a public/private key pair using the given seed.
+func NewKeyFromSeed(seed *[SeedSize]byte) (*PublicKey, *PrivateKey) {
```
armfazh, line 58: In Ed22519, this seed is actually the private key. What are the implications of using the same seed for both?

sign/eddilithium3/eddilithium.go
```diff
@@ -0,0 +1,210 @@
...
+	var seed1 [32]byte
+	var seed2 [32]byte
+
+	h := shake.NewShake256()
+	_, _ = h.Write(seed[:])
+	_, _ = h.Read(seed1[:])
```
armfazh, line 64: What if you read the 96 bytes needed for dilithium?

sign/eddilithium3/eddilithium.go
```diff
@@ -0,0 +1,210 @@
...
+	)
+	esig, _ := sk.e.SignPure(
+		msg,
+	)
+	copy(signature[mode3.SignatureSize:], esig[:])
+}
+
+// Verify checks whether the given signature by pk on msg is valid.
+func Verify(pk *PublicKey, msg []byte, signature []byte) bool {
+	if !mode3.Verify(
```
armfazh, line 88:
```suggestion
	return  mode3.Verify() && ed25519.Verify()
```

sign/eddilithium3/eddilithium_test.go
```diff
@@ -0,0 +1,106 @@
...
+		// We should generate a new signature for every verify attempt,
+		// as this influences the time a little bit.  This difference, however,
+		// is small and generating a new signature in between creates a lot
+		// pressure on the allocator which makes an accurate measurement hard.
+		Verify(pk, msg[:], sig[:])
+	}
+}
+
+func BenchmarkSign(b *testing.B) {
+	// Note that the expansion of the matrix A is done at Unpacking/Keygen
```
armfazh, line 28: Let's hide the details of the implementation of dilithium.

sign/eddilithium3/eddilithium_test.go
```diff
@@ -0,0 +1,106 @@
...
+)
+
+func BenchmarkVerify(b *testing.B) {
+	// Note that the expansion of the matrix A is done at Unpacking/Keygen
+	// instead of at the moment of verification (as in the reference
+	// implementation.)
+	var seed [32]byte
+	var msg [8]byte
+	var sig [SignatureSize]byte
+	pk, sk := NewKeyFromSeed(&seed)
```
armfazh, line 15: Let's promote the use of `GenerateKey` instead of `NewKeyFromSeed`.

sign/eddilithium3/eddilithium_test.go
```diff
@@ -0,0 +1,106 @@
+package eddilithium3
```
armfazh, line 1: Since there are no tests specific to the internals of the package.

```suggestion
package eddilithium3_test
```

sign/eddilithium3/example_test.go
```diff
@@ -0,0 +1,45 @@
...
+	pk, sk, err := eddilithium3.GenerateKey(nil)
+	if err != nil {
+		panic(err)
+	}
+
+	// (Alternatively one can derive a keypair from a seed,
+	// see NewKeyFromSeed().)
+
+	// Packs public and private key
+	var packedSk [eddilithium3.PrivateKeySize]byte
```
armfazh, line 20: I think this example must show the minimum code for generating and verfiying a signature.
Additionally, it can be created other `Example_` functions for Pack/Unpack functions, if needed.

sign/eddilithium3/example_test.go
```diff
@@ -0,0 +1,45 @@
...
+
+import (
+	"fmt"
+
+	"github.com/cloudflare/circl/sign/eddilithium3"
+)
+
+func Example() {
+	// Generates a keypair.
+	pk, sk, err := eddilithium3.GenerateKey(nil)
```
armfazh, line 11:
```suggestion
	pk, sk, err := eddilithium3.GenerateKey(rand.Reader)
```

claucece:
I agree with @armfazh . I just added an small comment as well ;)

sign/dilithium/gen.go
bwesterb: I've considered putting the eddilithium3 as a subpackage under Dilithium.  The point is that in deployment you should definitely not  use Dilithium — you want to combine it with some classical algorithm.  Thus it makes sense from a usability perspective to include that immediately so that people don't have to write it themselves.

sign/dilithium/dilithium.go
bwesterb: Let's merge this discussion with the thread above where I try to argue that it should be logically part of this package.

sign/eddilithium3/eddilithium.go
bwesterb: We're not using the same seed for both: we're deriving two different seeds from a single seed.  Both Dilithium and Ed25519 don't use their respective seeds directly, but stretch them in a similar way to derive the full private keys.  Even if there is a catastrophic weakness in Dilithium which would allow one to obtain the real private key, then one still isn't able to derive the seed as that would require breaking Shake (which Dilithium uses internally.). Thus it would actually be safe to use literally the same seed (instead of deriving two subseeds,) but that would set a bad example if someone every copy and pasted this for other schemes which don't work like that. So, this is safe and actually a bit conservative.

sign/eddilithium3/eddilithium.go
bwesterb: That would break abstraction.  I don't mind doing that for a good reason, but the performance gain here is marginal.

sign/eddilithium3/eddilithium.go
bwesterb: I considered that, but it makes the line very long and harder to read.

sign/eddilithium3/eddilithium_test.go
bwesterb: I adjusted the note.

sign/eddilithium3/eddilithium_test.go
bwesterb: I agree for code examples, but this is a benchmark.  Why use `GenerateKey` here?

sign/eddilithium3/eddilithium_test.go
bwesterb: Sure.

sign/eddilithium3/example_test.go
bwesterb: Specifying the random source explicitly as `crypto/rand` can only lead to issues if `rand` happens to point to something else than `crypto/rand` and doesn't have any advantage as `crypto/rand` is the default anyway.  What am I missing?

sign/eddilithium3/eddilithium.go
```diff
@@ -0,0 +1,210 @@
...
+	) {
+		return false
+	}
+	if !ed25519.Verify(
+		pk.e,
+		msg,
+		signature[mode3.SignatureSize:],
+	) {
+		return false
+	}
```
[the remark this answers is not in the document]
bwesterb: Yes, using timing the attacker can see which of the signatures fails.  Dilithium and Ed25519 signature verification aren't constant time either themselves: they leak information about *which* signature is being verified, just like most signature schemes.  (But it doesn't leak information about the private key as it obviously doesn't come into play at all during verification.)

sign/eddilithium3/example_test.go
bwesterb: You never just generate a keypair, sign and then directly verify: you pack a key after generating and you unpack a key before verifying with it. It doesn't make sense to force the reader to look at a different place to learn about that when they need it anyway.

sign/eddilithium3/eddilithium.go
armfazh: I see, this argument should be documented somewhere.

sign/eddilithium3/eddilithium_test.go
armfazh: For the same reason, promoting the use of `GenerateKey`. In fact, I am not sure, if we need `NewKeyFromSeed`. Disclaimer: I added `NewKeyFromSeed` in Ed25519 to be compatible with stdlib, but I think the function `Public()` does the same job.

sign/eddilithium3/example_test.go
armfazh: Most of people just copy-paste whatever they found in the examples, without paying attention to the details.

sign/eddilithium3/example_test.go
armfazh: This is not about functionality, it's about best practices and educating people. 
Even though, the internals of `GenerateKey` replace nil by `crypto/rand`, putting nil in the example, gives a bad example.

sign/dilithium/gen.go
armfazh: Say some other deployment (not us) wants to use Dilithium+RSA, then following your argument, does every new deployent should modify the internals of Dilithium package? 
It isn't better to create a standalone package that fetches `crypto/rsa` and `circl/sign/dilithium` as dependencies.

bwesterb: My proposal to include the Ed25519-Dilithium3 mode inside the Dilithium package does not prevent a third party from adding RSA2048-Dilithium3 using their own package.  They won't have the convenience of using the Modes API in the Dilithium package, but we won't either if we don't include the Ed25519-Dilithium3 mode in there.  So I do not see a downside.

sign/eddilithium3/eddilithium_test.go
bwesterb: In some situations you want to deterministically generate a key pair from a base secret.  For instance if you want to have one secret and generate a new key pair for each website/contact you have to prevent linking.

sign/eddilithium3/example_test.go
bwesterb: Yes, I agree, and then it is nice that the packing/unpacking code is in there as well, isn't it?

sign/eddilithium3/example_test.go
bwesterb: Why is it bad?  I think `rand.Reader` is strictly worse than `nil` because `rand` could be something else than `crypto/rand`.

sign/eddilithium3/eddilithium.go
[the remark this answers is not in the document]
claucece: Got it! But I was thinking that maybe here we can always check `!mode3.Verify && !ed25519.Verify` at the end of the function and then return false..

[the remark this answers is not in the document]
bwesterb: The compiler will turn that into just checking mode3.Verify and then returning false immediately when that is false. ( https://godbolt.org/z/5HaWu7 )

To prevent this compiler optimisation we could add a variant of verify that returns uint64s instead of bools and use `mode3.VerifyI() | ed25519. VerifyI()`.  I'm not sure whether such a complication of the API is worth it.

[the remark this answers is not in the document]
claucece: Oh, right. Thanks for letting me now @bwesterb . Let's keep it the way you did it ;)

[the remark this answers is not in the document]
chris-wood: Or we could just use `ConstantTimeByteEq`, casting the inputs to `byte`s and output to a `bool`? I don't see how this complicates the API, given that this is internal to `Verify`.

sign/eddilithium3/example_test.go
chris-wood: I agree with @armfazh. Passing `crypto/rand` in is best practice, and we ought to make the test code reflect best practice.

sign/eddilithium3/example_test.go
```diff
@@ -0,0 +1,45 @@
+package eddilithium3_test
+
+import (
+	"fmt"
+
+	"github.com/cloudflare/circl/sign/eddilithium3"
+)
+
+func Example() {
```
chris-wood, line 9: Why is this not a "regular" unit test? (Passing in `t *testing.T`, and having a proper name.)

bwesterb: This is a package wide testable example that appears after the description in the godoc. See https://blog.golang.org/examples

sign/eddilithium3/example_test.go
bwesterb: Could you please explain *why* this is bad practice.

sign/eddilithium3/eddilithium.go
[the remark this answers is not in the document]
bwesterb: That might also work, but it doesn't change the fact that signature verification of Ed25519 and Dilithium aren't constant-time themselves.  Even if we make this check constant time, we'll still leak exactly the same fact because Dilithium and Ed25519 verification aren't constant time.

sign/dilithium/eddilithium3.go
```diff
@@ -0,0 +1,96 @@
...
+	"errors"
+	"fmt"
+	"io"
+
+	"github.com/cloudflare/circl/sign/dilithium/internal/common"
+	"github.com/cloudflare/circl/sign/eddilithium3"
+)
+
+// implEdMode3 implements the mode.Mode interface for EdDilithium3.
+type implEdMode3 struct{}
```
mmcloughlin, line 15: nit: not the biggest fan of `impl` in type names. In a case like this I would usually just call this `edMode3`

sign/dilithium/eddilithium3.go
```diff
@@ -0,0 +1,96 @@
...
+
+// EdMode3 is Dilithium in mode "EdDilithium3".
+var EdMode3 Mode = &implEdMode3{}
+
+func (m *implEdMode3) GenerateKey(rand io.Reader) (PublicKey,
+	PrivateKey, error) {
```
mmcloughlin, line 21: nit: unusual line break

sign/dilithium/eddilithium3.go
```diff
@@ -0,0 +1,96 @@
...
+
+func (m *implEdMode3) NewKeyFromExpandedSeed(seed *[96]byte) (PublicKey,
+	PrivateKey) {
+	panic("Not supported for hybrid modes")
+}
+
+func (m *implEdMode3) NewKeyFromSeed(seed []byte) (PublicKey,
+	PrivateKey) {
+	if len(seed) != common.SeedSize {
+		panic(fmt.Sprintf("seed must be of length %d", common.SeedSize))
```
mmcloughlin, line 33: `fmt.Errorf`?

sign/dilithium/templates/mode.templ.go
```diff
@@ -28,7 +32,11 @@ func (m *{{ .Impl }}) GenerateKey(rand io.Reader) (PublicKey,
 
 func (m *{{ .Impl }}) NewKeyFromExpandedSeed(seed *[96]byte) (PublicKey,
 	PrivateKey) {
+{{- if  .Hybrid }}
+	panic("Not supported for hybrid modes")
```
mmcloughlin, line 36: nit: lowercase error messages

https://github.com/golang/go/wiki/CodeReviewComments#error-strings

sign/eddilithium3/eddilithium.go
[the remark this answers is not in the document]
mmcloughlin: Yep, I was going to suggest using `subtle`, as @chris-wood recommends.

sign/eddilithium3/eddilithium.go
```diff
@@ -0,0 +1,210 @@
+// eddilithium3 implements the hybrid signature scheme Ed25519-Dilithium3.
+package eddilithium3
+
+import (
+	"crypto"
+	cryptoRand "crypto/rand"
```
mmcloughlin, line 6: nit: `cryptorand`

> They are lower case, with no under_scores or mixedCaps

https://blog.golang.org/package-names

sign/eddilithium3/eddilithium.go
```diff
@@ -0,0 +1,210 @@
...
+		pk.e,
+		msg,
+		signature[mode3.SignatureSize:],
+	) {
+		return false
+	}
+	return true
+}
+
+// Sets pk to the public key encoded in buf.
```
mmcloughlin, line 105: Doesn't conform to the "Comment Sentences" guideline:

https://github.com/golang/go/wiki/CodeReviewComments#comment-sentences

This applies to other comments in this file.

Would have expected a lint failure for this.

sign/dilithium/eddilithium3.go
bwesterb: It's because of the type `implMode3` in `sign/dilithium` (which is generated from the same file) which then would be `mode3` which conflicts with the subpackage `sign/dilithium/mode3`.

mmcloughlin: Got it. Yeah not a huge deal, I just don't like my Go looking like Java 😝

bwesterb: Yeah, I don't like it either, but at least it doesn't show up in the API :).

sign/eddilithium3/example_test.go
armfazh: Yes, you can do that using another example function.

bwesterb: But that would mean that the user would have to look in two different places.  I would prefer one example to be exhaustive for the common use case.  Packing and unpacking is required in the common use case.

sign/eddilithium3/example_test.go
armfazh: Some input https://checkmarx.gitbooks.io/go-scp/cryptographic-practices/pseudo-random-generators.html

sign/eddilithium3/eddilithium.go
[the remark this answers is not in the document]
armfazh: > Let's keep it the way you did it
 
 Agree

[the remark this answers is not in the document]
bwesterb: Again, using constant time operations here does not do anything to the overall behaviour of the function as ed25519.Verify and dilithium3.Verify are very much not constant time.

sign/eddilithium3/eddilithium_test.go
armfazh: > In some situations you want to deterministically generate a key pair from a base secret. For instance if you want to have one secret and generate a new key pair for each website/contact you have to prevent linking.

This is a use case that is not part of the algorithmic definition.

sign/eddilithium3/example_test.go
chris-wood: Got it. Do we have actual unit tests?

sign/eddilithium3/eddilithium.go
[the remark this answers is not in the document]
chris-wood: Is it possible for those functions to be done in constant time? If so, I see no reason to make this constant time on the assumption that these two functions might later change. The cost is basically nothing.

sign/dilithium/gen.go
armfazh: > They won't have the convenience of using the Modes API in the Dilithium package,

I didn't get this sentence. So, are you saying that having the hybrid mode inside the package has better usability than being in an external package? If so, that feels like the dilithium must improves its usability.

armfazh: In the meantime, could you please refactor to have the hybrid mode as an external package.

bwesterb: > > They won't have the convenience of using the Modes API in the Dilithium package,
> 
> I didn't get this sentence. So, are you saying that having the hybrid mode inside the package has better usability than being in an external package? If so, that feels like the dilithium must improves its usability.

I think it should be easy to switch between Ed25519-Dilithium3 and Dilithium3, do you agree?

bwesterb: > In the meantime, could you please refactor to have the hybrid mode as an external package.

It is already an external package.  What you're saying is that I should remove the ability to pick Ed25519-Dilithium3 via the dilithium.ModyByName API?

sign/eddilithium3/eddilithium.go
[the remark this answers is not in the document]
bwesterb: It is possible to create constant-time versions of dilithium3.Verify and ed25519.Verify, but they would be significantly slower.  The reason I don't like making this function here constant-time, is that it is a very clear statement "here is something special: here we should be constant time", whereas in fact that is useless action.  If it would be commonplace for signature schemes to have constant-time verification, I would think differently to prevent copy-paste errors, but that is not the case: signature schemes generally don't have constant-time verification operation.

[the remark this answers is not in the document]
bwesterb: So the cost is not in performance, but in ease of understanding for the reader.

sign/eddilithium3/eddilithium_test.go
bwesterb: I don't understand that?

Isn't what I describe is what is essentially used for U2F?

sign/eddilithium3/eddilithium.go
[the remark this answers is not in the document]
chris-wood: I disagree with the claim that this is a cost to the reader. It's quite a trivial change. I do buy that signature verification algorithms are not commonly constant-time. However, we also don't have much experience with hybrid signature schemes, so assuming whatever we did before maps perfectly to what we should do here now doesn't seem right.

Food for thought.

sign/eddilithium3/eddilithium_test.go
armfazh: I am not saying that `NewKeyFromSeed` is useles at all, am only pointing that l prefer to use GenerateKey.

sign/eddilithium3/example_test.go
bwesterb: Yes, `TestSignThenVerifyAndPkSkPacking` and `TestPublicFromPrivate` in the `_test.go`.

sign/dilithium/eddilithium3.go
bwesterb: In other places we raise a string.  Why be inconsistent and raise an error here instead of a string?

bwesterb: Ah, I see I'm mistaken: it's inconsistent already.  I'll choose one. Thanks.

sign/dilithium/templates/mode.templ.go
bwesterb: Thanks.

sign/eddilithium3/eddilithium.go
bwesterb: @armfazh What do you think?  I think it's worth here to make it easy to see this is `crypto`s `rand` sub package.  (We also do it like that in `ed25519` and `ed448`.)

sign/eddilithium3/eddilithium.go
bwesterb: Yes, it is unfortunately the standard, so we'll have to live with `Pack packs the public key`, `SignaturesSize is the signature size`, etc. Of course, in a creative mode I may write `Pack serializes the public key`, but was this worth the mental effort?  I changed the comments accordingly.

sign/eddilithium3/eddilithium.go
mmcloughlin: Just to be clear, I think the package alias is good, I'm just suggesting `cryptorand` not `cryptoRand`.

Yes, looks like it is done the same way in `ed25519` and `ed448`. I would make the same comment for those occurrences.

mmcloughlin: https://github.com/golang/go/blob/dd150176c3cc49da68c8179f740eadc79404d351/src/crypto/ed25519/ed25519.go#L22

sign/eddilithium3/eddilithium_test.go
bwesterb: Yes, I agree to use `NewKeyFromSeed` in code examples, but this is a benchmark.

sign/eddilithium3/eddilithium.go
```diff
@@ -0,0 +1,210 @@
...
+		pk.e,
+		msg,
+		signature[mode3.SignatureSize:],
+	) {
+		return false
+	}
+	return true
+}
+
+// Unpack unpacks pk to the public key encoded in buf.
```
mmcloughlin, line 105:
```suggestion
// Unpack pk to the public key encoded in buf.
```

sign/eddilithium3/eddilithium.go
```diff
@@ -0,0 +1,210 @@
...
+
+// Unpack sets sk to the private key encoded in buf.
+func (sk *PrivateKey) Unpack(buf *[PrivateKeySize]byte) {
+	var tmp [mode3.PrivateKeySize]byte
+	copy(tmp[:], buf[:mode3.PrivateKeySize])
+	sk.d.Unpack(&tmp)
+	sk.e = *ed25519.NewKeyFromSeed(buf[mode3.PrivateKeySize:])
+}
+
+// Pack packs the public key into buf.
```
mmcloughlin, line 122:
```suggestion
// Pack the public key into buf.
```

sign/eddilithium3/eddilithium.go
```diff
@@ -0,0 +1,210 @@
...
+func (pk *PublicKey) Pack(buf *[PublicKeySize]byte) {
+	var tmp [mode3.PublicKeySize]byte
+	pk.d.Pack(&tmp)
+	copy(buf[:mode3.PublicKeySize], tmp[:])
+	copy(buf[mode3.PublicKeySize:], pk.e)
+}
+
+// Pack packs the private key into buf.
```
mmcloughlin, line 130:
```suggestion
// Pack the private key into buf.
```

sign/eddilithium3/eddilithium.go
bwesterb: I prefer your suggestion, but the example from the code guidelines you linked is

```go
// Encode writes the JSON encoding of req to w.
func Encode(w io.Writer, req *Request) { ...
```

Wouldn't they have used `Encode req as JSON and write into w.` or something of that affect?

bwesterb: Also your sentence is an imperative.  Wouldn't `Unpacks pk ...` make more sense?

mmcloughlin: First word has to be `Unpack`, and it just doesn't read well to use `unpack` twice.

mmcloughlin: In the other example, at least `write` is actually a different word :joy: 

Anyway, the suggested edit I made is what I do in this annoying situation. I'll leave it up to you now, I don't want to bother you endlessly with small comments.

bwesterb: @armfazh What is your opinion?  Personally, I don't like to strictly adhere to this guideline strictly as it distract us with the occasional linguistic puzzle as above.

sign/eddilithium3/eddilithium.go
armfazh: they are using an alias for `crypto/rand` because there exists a variable called `rand` in one of the functions.
So renaming this variable allows to use `crypto/rand` without alias.

sign/eddilithium3/eddilithium.go
bwesterb: Done.

sign/eddilithium3/eddilithium.go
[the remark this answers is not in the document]
bwesterb: Yes, it's good to stay critical of these details!

sign/eddilithium3/eddilithium.go
mmcloughlin: Regardless of collisions, there is an argument for using `cryptorand` alias anyway, to avoid any accidental use of `math/rand` instead of `crypto/rand`.

This always makes me nervous... if you call `rand.Read(...)` then `goimports` could give you the wrong one, and you have to double-check it did the right thing.

sign/dilithium/gen.go
bwesterb: After discussion I've removed EdDilithium3 support for the Dilitihium mode API.  The easy switching benefit will return when we have a unified signature/verifier API for all schemes.

sign/eddilithium3/eddilithium.go
armfazh: I personally follow that guideline, the syntax is 

```
// XYZ <whatever> <PERIOD>
 function XYZ()
```

This works well for functions and variables, but the issue is when the name of the function is a verb, or when the function is a method as in object-oriented programming.

For `CIRCL`, I think it's enough to have a documentation string that is easy to read and talks about its input parameters and what it returns.

armfazh: Feel free to open an issue for discussing style and documentation, if needed.

sign/eddilithium3/eddilithium_test.go
armfazh: all right, let's move forward as is.

sign/dilithium/dilithium.go
```diff
@@ -6,13 +6,23 @@
...
-// subpackge.  For instance, Dilithium III can be found in
+// subpackage.  For instance, Dilithium III can be found in
 //
 //  github.com/cloudflare/circl/sign/dilithium/mode3
 //
 // If your choice for mode is fixed compile-time, use the subpackages.
 // This package provides a convenient wrapper around all of the subpackages
 // so one can be chosen at runtime.
+//
+// As we don't understand the security of Dilithium well yet, it is advisable
```
armfazh, line 17: From https://pq-crystals.org/dilithium/
```
For users who are interested in using Dilithium, we recommend the following:

Use Dilithium in a so-called hybrid mode in combination with an established "pre-quantum" signature scheme.
We recommend using the Dilithium-1280x1024 parameter set, which—according to a very conservative analysis—achieves around 128 bits of security against all known classical and quantum attacks.
```

sign/eddilithium3/example_test.go
armfazh: (already discussed internally)

sign/dilithium/dilithium.go
bwesterb: What about the current version?

armfazh: It's ok. Do the rebase with master and proceed to merge.

PR #136, opened 2020-07-07
Fix linting

claucece:
It is not encouraged anymore to use `enable-all`, as it will be deprecated and removed soon: https://golangci-lint.run/usage/configuration/

### DIFF
--- a/.etc/golangci.yml
+++ b/.etc/golangci.yml
@@ -17,7 +17,6 @@ linters:
     - govet
     - ineffassign
     - interfacer
-    - lll
     - misspell
     - nakedret
     - scopelint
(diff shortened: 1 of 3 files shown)


claucece:
@armfazh check this PR as well ;)

.etc/golangci.yml
```diff
@@ -1,15 +1,34 @@
...
-    - gocritic    # it can be improved
-    - gocognit    # it can be improved
-    - gochecknoglobals
-    - gochecknoinits
-    - gomnd
-    - dupl
-    - wsl
+  disable-all: true
+  enable:
+    - bodyclose
```
armfazh, line 4: Let's keep track of these linters, as a to-do task for improving the code.

```suggestion
#   - lll    
#   - gocritic    
#   - gocognit    
    - bodyclose
```

PR #138, opened 2020-07-08
sign: add hybrid of Dilithium4 and Ed448

bwesterb:
Almost the same as eddilithium3 except for generation of the seed (as Ed448 requires more entropy) and the API of ed448.  (ed448 and ed25519 aren't drop-in replacements of eachother, yet.)

### DIFF
--- a/sign/eddilithium4/eddilithium.go
+++ b/sign/eddilithium4/eddilithium.go
@@ -1,0 +1,213 @@
+// eddilithium4 implements the hybrid signature scheme Ed448-Dilithium4.
+package eddilithium4
+
+import (
+	"crypto"
+	cryptoRand "crypto/rand"
+	"errors"
+	"io"
+
+	"github.com/cloudflare/circl/internal/shake"
+	"github.com/cloudflare/circl/sign/dilithium/mode4"
+	"github.com/cloudflare/circl/sign/ed448"
+)
+
+const (
+	// Size of seed for NewKeyFromSeed
+	SeedSize = ed448.PrivateKeySize // > mode3.SeedSize
+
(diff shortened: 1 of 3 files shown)


bwesterb:
@claucece

sign/eddilithium4/eddilithium_test.go
```diff
@@ -0,0 +1,111 @@
...
+		eddilithium4.NewKeyFromSeed(&seed)
+	}
+}
+
+func BenchmarkPublicFromPrivate(b *testing.B) {
+	var seed [57]byte
+	for i := 0; i < b.N; i++ {
+		b.StopTimer()
+		binary.LittleEndian.PutUint64(seed[:], uint64(i))
+		_, sk := eddilithium4.NewKeyFromSeed(&seed)
```
armfazh, line 59: I think here we need to measure the function `NewKeyFromSeed`.  In EdDSA, the seed is the private key.

sign/eddilithium4/eddilithium.go
```diff
@@ -0,0 +1,213 @@
...
+	"io"
+
+	"github.com/cloudflare/circl/internal/shake"
+	"github.com/cloudflare/circl/sign/dilithium/mode4"
+	"github.com/cloudflare/circl/sign/ed448"
+)
+
+const (
+	// Size of seed for NewKeyFromSeed
+	SeedSize = ed448.PrivateKeySize // > mode3.SeedSize
```
armfazh, line 17:
```suggestion
	SeedSize = ed448.PrivateKeySize // > mode4.SeedSize
```

PR #139, opened 2020-07-08
Add a GitHub action to lint the Markdown.

jbampton:
Add a Markdownlint config file.

Rules linted:
MD001 - Header levels should only increment by one level at a time
MD012 - Multiple consecutive blank lines
MD022 - Headers should be surrounded by blank lines
MD030 Spaces after list markers

https://www.npmjs.com/package/markdownlint-cli

Trim single leading whitespace.
Remove unneeded "/" from HTML img tag.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,0 +1,15 @@
+name: Lint Markdown
+
+on: [push, pull_request]
+
+jobs:
+  test:
+    runs-on: ubuntu-18.04
+    steps:
+      - uses: actions/checkout@v2
+      - name: Use Node.js
+        uses: actions/setup-node@v1
+        with:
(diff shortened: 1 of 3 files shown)


armfazh:
Thanks, for your contribution. Any chance to use the github action https://github.com/marketplace/actions/markdownlint-cli .
Also, as this linter looks for errors in markdown files, I don't think is worth to trigger it on every push and pull_request.
Does it possible to trigger only if a markdown file was modified?. If this isn't possible, we can merge only the markdown files after linter.

armfazh:
The readme.md file was updated after an offline linter pass. 
As there are few markdown files in this project, I consider not going for the automated action, and close this PR.
Thanks for your contribution.

PR #140, opened 2020-07-09
Add eddsa for prehashed and context use for ed448 #109 

claucece:
Add the same API as ed25519 to ed448.

@bwesterb , I found that there were some style problems with the comments on the eddilithium4 and eddilithium3 code. I changed it. Hope that is ok. 

Here the explanations for comments in Golang: https://github.com/golang/go/wiki/CodeReviewComments#comment-sentences, https://golang.org/doc/effective_go.html#commentary


claucece:
This is now done @armfazh . Thanks for the review so far!

PR #141, opened 2020-07-14
Refactoring ed25519 for compatibility with stdlib

armfazh:
We want to make that circl/sign/ed25519 be compatible with methods currently in crypto/ed25519 standard library.

Here are the schemes supported by circl/sign/ed25519. 

| Scheme Name    | Sign Function     | Verification | Context            |
|----------------|-------------------|--------------|--------------------|
| Ed25519 (pure) | Sign              | Verify       | None               |
| Ed25519Ph      | SignPh            | VerifyPh     | Yes (can be empty) |
| Ed25519Ctx     | SignWithCtx           | VerifyWithCtx    | Yes (non-empty)    |
| All above      | (PrivateKey).Sign | VerifyAny    | As above           |

### DIFF
--- a/sign/ed25519/pubkey112.go
+++ b/sign/ed25519/pubkey112.go
@@ -1,4 +1,4 @@
-// +build go1.12
+// +build !go1.13
 
 package ed25519
 
--- a/sign/ed25519/pubkey112.go
+++ b/sign/ed25519/pubkey112.go
@@ -1,0 +1,20 @@
+// +build go1.12
+
+package ed25519
+
+import (
+	"bytes"
+	"crypto"
+)
+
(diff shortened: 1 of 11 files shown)


armfazh:
@bwesterb I've updated this PR with your comments.

sign/ed25519/pubkey112.go
```diff
@@ -0,0 +1,20 @@
+// +build go1.12
```
bwesterb, line 1: `go1.12` means Go 1.12 or later.  So we want to use `!go1.13` here, which effectively means Go 1.12 or earlier.

armfazh:
Closed as it was superseeded by #142 , which also includes these changes.

Issue #132, opened 2020-07-02
ed25519 is not a drop-in replacement of crypto/ed25519


bwesterb:
- [ ] GetPublic -> Public
- [ ] NewKeyFromSeed returns *KeyPair instead of PrivateKey
- [ ] (check for others.)

bwesterb:
@armfazh Fixing this would make it easier to copy code from Circl to our stdlib fork.

claucece:
Sort of already discussed here: https://github.com/cloudflare/circl/issues/109, and there is already a PR for some of this: https://github.com/cloudflare/circl/pull/123

bwesterb:
Would we want ed448 to be a drop-in replacement of ed25519 as well?

chris-wood:
@bwesterb can we close this issue?

bwesterb:
I just checked the API and it's a drop in replacement now!

PR #142, opened 2020-07-14
Make ed448 compatible with new ed25519 API

armfazh:
Mimics the API of circl/sign/ed25519.
Updates eddilithium4 accordingly.

Merge after #132

### DIFF
--- a/.github/workflows/ci-actions.yml
+++ b/.github/workflows/ci-actions.yml
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-18.04
     strategy:
       matrix:
-        GOVER: ['1.14', '1.13', '1.12']
+        GOVER: ['1.14', '1.13']
     steps:
       - name: Checkout
         uses: actions/checkout@v2
--- a/.github/workflows/ci-actions.yml
+++ b/.github/workflows/ci-actions.yml
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-18.04
     strategy:
       matrix:
-        GOVER: ['1.14', '1.13']
+        GOVER: ['1.14', '1.13', '1.12']
     steps:
       - name: Checkout
         uses: actions/checkout@v2
--- a/sign/ed25519/ed25519.go
+++ b/sign/ed25519/ed25519.go
@@ -38,7 +38,6 @@ package ed25519
 import (
 	"bytes"
 	"crypto"
-	cryptoEd25519 "crypto/ed25519"
 	cryptoRand "crypto/rand"
 	"crypto/sha512"
 	"crypto/subtle"
@@ -89,9 +88,6 @@ const (
 	ED25519Ph
 	ED25519Ctx
 )
-
-// PublicKey is the type of Ed25519 public keys.
-type PublicKey = cryptoEd25519.PublicKey
 
 // PrivateKey is the type of Ed25519 private keys. It implements crypto.Signer.
 type PrivateKey []byte
--- a/sign/ed25519/ed25519.go
+++ b/sign/ed25519/ed25519.go
@@ -1,6 +1,7 @@
 // Package ed25519 implements Ed25519 signature scheme as described in RFC-8032.
 //
-// This package implements the three signature variants.
+// This package provides optimized implementations of the three signature
+// variants and maintaining closer compatiblilty with crypto/ed25519.
 //
 //  | Scheme Name | Sign Function     | Verification  | Context           |
 //  |-------------|-------------------|---------------|-------------------|
@@ -16,7 +17,8 @@
 //
 // Signing with Ed25519Ph or Ed25519Ctx requires a context string for domain
 // separation. This parameter is passed using a SignerOptions struct defined
-// in this package.
+// in this package. While Ed25519Ph accepts an empty context, Ed25519Ctx
+// enforces non-empty context strings.
 //
 // Compatibility with crypto.ed25519
 //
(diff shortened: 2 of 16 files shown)


.github/workflows/ci-actions.yml
```diff
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-18.04
     strategy:
       matrix:
-        GOVER: ['1.14', '1.13', '1.12']
+        GOVER: ['1.14', '1.13']
```
bwesterb, line 15: Why are we removing support for 1.12?

(Latest Debian is still on 1.11.  Ubuntu 19.10 is on 1.12.  Ubuntu 20.04 is on 1.13.)

sign/ed25519/ed25519.go
```diff
@@ -1,144 +1,217 @@
...
+//  | All above   | (PrivateKey).Sign | VerifyAny     | As above          |
+//
+// Specific functions for sign and verify are defined. A generic signing
+// function for all schemes is available through the crypto.Signer interface,
+// which is implemented by the PrivateKey type. A correspond all-in-one
+// verification method is provided by the VerifyAny function.
+//
+// Signing with Ed25519Ph or Ed25519Ctx requires a context string for domain
+// separation. This parameter is passed using a SignerOptions struct defined
+// in this package.
```
bwesterb, line 19: Perhaps explain the difference between Ed25519Ph and Ed25519Ctx.

sign/ed25519/ed25519.go
```diff
@@ -1,144 +1,217 @@
...
+//
+// Compatibility with crypto.ed25519
+//
+// These functions are compatible with the “Ed25519” function defined in
+// RFC-8032. However, unlike RFC 8032's formulation, this package's private
+// key representation includes a public key suffix to make multiple signing
+// operations with the same key more efficient. This package refers to the
+// RFC-8032 private key as the “seed”.
```
bwesterb, line 27: Perhaps mention that this package can be used as a drop-in replacement of `crypto/ed25519`.

sign/ed25519/ed25519.go
```diff
@@ -1,144 +1,217 @@
...
-func (kp *KeyPair) Sign(rand io.Reader, message []byte, opts crypto.SignerOpts) ([]byte, error) {
+// Use a SignerOptions struct (defined in this package) to pass a context
+// string for signing.
+func (priv PrivateKey) Sign(
+	rand io.Reader,
+	message []byte,
+	opts crypto.SignerOpts) (signature []byte, err error) {
 	var ctx string
-	if o, ok := opts.(*Options); ok {
+	var scheme SchemeID
```
bwesterb, line 137: Perhaps explicitly set or note that the default is Ed25519.

sign/ed25519/ed25519_test.go
```diff
@@ -1,390 +1,89 @@
+// Copyright 2016 The Go Authors. All rights reserved.
```
bwesterb, line 1: Is this copyright notice really appropriate?

bwesterb:
@claucece Might want to take a look as well.

.github/workflows/ci-actions.yml
armfazh: The issue is that crypto/ed25519 appears inside stdlib as of 1.13. 
The public key is an alias of the stdlb crypto/ed25519. This alias plays nicely with type assertions, for example those found in x509.CreateCertificate.
An alternative for not removing support of 1.12 is welcomed.

bwesterb: We can control the definition of PublicKey by a build tag.  The "go1.13" build tag is satisfied on Go 1.13 and beyond.

armfazh: The issue with that is that golang.org/x/crypto/ed25519 becomes now a dependency, but only for 1.12, and not for 1.13.
how do we express this conditional dependencies in the go.mod file?

claucece:
Oh, nice! I'll take a look later today ;)

.github/workflows/ci-actions.yml
bwesterb: I meant: use crypto/ed25519.PublicKey for go1.13 and beyond and simply []byte in previous Go versions.  Go 1.12 and before doesn't support Ed25519 in crypto/x509 anyway.

mmcloughlin: Go support policy is the last two major releases, so I tend to think it's reasonable for libraries to follow the same rules.

https://golang.org/doc/devel/release.html#policy

As far as I am concerned Debian/Ubuntu are too far behind, and library authors should not have to accommodate them. People should be installing Go some other way on those platforms.

claucece: I'm actually pretty much in favor or removing 1.12, it creates some problems as the API changed a lot.

Debian 10 should have go 1.13: https://tecadmin.net/install-go-on-debian/ for what I understand.

claucece:
This looks very nice.. should we do the same for circl/ed25519?

.github/workflows/ci-actions.yml
bwesterb: The latest in Debian stable (Buster at the moment) is [1.11](https://packages.debian.org/buster/golang).

It's easy to install the latest Go version from source, but that does adds some friction to adoption.

I'm not hell bend on supporting very old version (also as we don't have a big user base yet), but if it's not too much effort, then it does ease adoption.

sign/ed25519/ed25519.go
```diff
@@ -1,144 +1,225 @@
 // Package ed25519 implements Ed25519 signature scheme as described in RFC-8032.
 //
-// References:
-//  - RFC8032 https://rfc-editor.org/rfc/rfc8032.txt
-//  - Ed25519 https://ed25519.cr.yp.to/
-//  - High-speed high-security signatures. https://doi.org/10.1007/s13389-012-0027-1
+// This package implements the three signature variants.
```
claucece, line 3: Just opening...

PR #143, opened 2020-07-15
Generic signature API

bwesterb:
Used to easily add support for new (PQ) signature schemes to [our fork of Go](https://github.com/cloudflare/go) and cfssl.


bwesterb:
Another reason is that I'm quickly prototyping integration into Go stdlib to find possible issues with the API.  I found one already: we can't use `PubicKeyAlgorithm() x509.PublicKeyAlgorithm` in our interface for that would cause an import cycle.

bwesterb:
@armfazh @claucece Have another look.

claucece:
I'll check this tomorrow ;)

sign/api.go
```diff
@@ -0,0 +1,413 @@
...
+		if tlsScheme.TLSIdentifier() == id {
+			return scheme
+		}
+	}
+	return nil
+}
+
+func UnmarshalPEMPublicKey(data []byte) (PublicKey, error) {
+	block, rest := pem.Decode(data)
+	if len(rest) != 0 {
```
armfazh, line 157: I think the checks are for the block, like in the [Decode example](https://golang.org/pkg/encoding/pem/#Decode)

```
        block, rest := pem.Decode(pubPEMData)
	if block == nil || block.Type != "PUBLIC KEY" {
		log.Fatal("failed to decode PEM block containing public key")
	}
```

sign/eddilithium3.go
```diff
@@ -0,0 +1,108 @@
...
+	"github.com/cloudflare/circl/sign/eddilithium3"
+
+	cryptoRand "crypto/rand"
+	"encoding/asn1"
+	"errors"
+)
+
+type edDilithium3Scheme struct{}
+
+var EdDilithium3 Scheme = &edDilithium3Scheme{}
```
armfazh, line 13: it does not require the address

sign/eddilithium3.go
```diff
@@ -0,0 +1,108 @@
...
+
+import (
+	"github.com/cloudflare/circl/sign/eddilithium3"
+
+	cryptoRand "crypto/rand"
+	"encoding/asn1"
+	"errors"
+)
+
+type edDilithium3Scheme struct{}
```
armfazh, line 11: All these definitions must live inside of `sign/eddilithium3` package,

armfazh, line 11: This is a zero-sized data, which is ok. What if this is an integer, which can also be used as identifier in the list of schemes.

claucece:
I think it is looking good. Maybe we should be more clear on some names and all, though

sign/eddilithium3.go
bwesterb: Try it without the `&`.

sign/eddilithium3.go
bwesterb: That is what I would like as well, but how do you do that without creating an import cycle?

sign/api.go
```diff
@@ -0,0 +1,413 @@
...
+	PublicKeySize() uint
+
+	// Size of binary marshalled public keys
+	PrivateKeySize() uint
+
+	// Name of the scheme
+	Name() string
+
+	// Size of signatures
+	SignatureSize() uint
```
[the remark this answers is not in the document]
bwesterb: Sizes are always positive.  However, it might indeed be better to use `int` as it removed the need for casting, say, `len(something)` to `uint`.

sign/api.go
```diff
@@ -0,0 +1,413 @@
...
+	// For compatibility with Go standard library
+	crypto.Signer
+
+	encoding.BinaryMarshaler
+}
+
+// A Scheme represents a specific instance of a signature scheme.
+type Scheme interface {
+	// GenerateKey creates a new key-pair.
+	GenerateKey() (PublicKey, PrivateKey, error)
```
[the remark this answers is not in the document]
bwesterb: The common case should be as easy and foolproof as possible and the common case is that people don't  need to use a custom random source.

sign/api.go
```diff
@@ -0,0 +1,413 @@
...
+}
+
+// SchemeByName returns the scheme with the given name and nil if it is not
+// supported.  Use ListSchemes() to list supported schemes.
+// Names are case insensitive.
+func SchemeByName(name string) Scheme {
+	// XXX add a (compile time?) lookup table
+	name = strings.ToLower(name)
+	for _, scheme := range schemes {
+		if strings.ToLower(scheme.Name()) == name {
```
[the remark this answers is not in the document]
bwesterb: If we would remove this `ToLower()` then we will never match any scheme that has an upper case character in the return of its `scheme.Name()`.

sign/eddilithium3.go
bwesterb: You can't use a specific integer as a type.

sign/api.go
```diff
@@ -0,0 +1,413 @@
...
+		data,
+	}
+
+	return asn1.Marshal(pkix)
+}
+
+// Go1.15 adds {PublicKey,PrivateKey}.Equal(). Until then, we can use that
+// we use this.
+
+func PublicKeysEqual(a, b PublicKey) bool {
```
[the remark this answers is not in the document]
bwesterb: I could use the interface that we expect Go1.15 to have, but I'd prefer to wait in case they change their mind.

sign/api.go
```diff
@@ -0,0 +1,413 @@
...
+	for _, scheme := range schemes {
+		if strings.ToLower(scheme.Name()) == name {
+			return scheme
+		}
+	}
+	return nil
+}
+
+// ListSchemeNames returns the names of all schemes supported.
+func ListSchemeNames() []string {
```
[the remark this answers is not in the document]
bwesterb: I'm thinking of adding such a function as well to return `[]Scheme`.

sign/eddilithium3.go
```diff
@@ -0,0 +1,108 @@
...
+	if err != nil {
+		return nil, nil, err
+	}
+	return wrapPublicKey(pk, EdDilithium3), wrapPrivateKey(sk, EdDilithium3), nil
+}
+
+func (s *edDilithium3Scheme) Sign(sk PrivateKey, message []byte,
+	opts *SignatureOpts) []byte {
+	sig := make([]byte, eddilithium3.SignatureSize)
+	if opts != nil && opts.Context != "" {
```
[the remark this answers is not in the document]
bwesterb: I don't understand.  For `opts.Context` to make sense, we must have `opts != nil`.

bwesterb:
> Maybe we should be more clear on some names and all, though

Sure, which?

sign/eddilithium3.go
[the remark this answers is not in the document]
armfazh: if any of the options is not used, then is ignored.

armfazh:
I just added a generic interface for four signatures, there are some methods marked with TODO.
Also I moved some methods into the `circl/pki` package.

sign/api.go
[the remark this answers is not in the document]
claucece: Yes, but perhaps the `scheme.Name()` should always be enforced to be lower case... what do you think?

sign/api.go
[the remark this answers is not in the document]
claucece: I agree now.

sign/api.go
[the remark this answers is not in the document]
claucece: Yeah.. I was thinking from the casting perspective that it is used many times...

claucece:
> Sure, which?

Let me thinking of a nice naming convention and propose something over here tomorrow.

sign/api.go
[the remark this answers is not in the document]
bwesterb: Ok, let's go for int!

sign/ed25519/signapi.go
```diff
@@ -0,0 +1,75 @@
...
+
+	"github.com/cloudflare/circl/sign"
+)
+
+// Scheme is
+const Scheme = scheme(sign.Ed25519)
+
+type scheme sign.SchemeID
+
+func (scheme) ID() sign.SchemeID    { return sign.SchemeID(Scheme) }
```
bwesterb, line 15: What is the advantage of this compared to:

```
type scheme struct{}
func (scheme) ID() sign.SchemeID    { return sign.Ed25519 }
```

sign/api/api.go
```diff
@@ -0,0 +1,33 @@
...
+
+var allSchemes [sign.SchemeCount]sign.Scheme
+var allSchemeNames map[string]sign.Scheme
+
+func init() {
+	allSchemeNames = make(map[string]sign.Scheme)
+	register(ed25519.Scheme)
+	register(ed448.Scheme)
+	register(eddilithium3.Scheme)
+	register(eddilithium4.Scheme)
```
bwesterb, line 20: Why not `register(sign.Ed25519, ed25519.Scheme)`?  Then there is no need for a public `ID()`, `SchemeID`, etc.

sign/api/api.go
```diff
@@ -0,0 +1,33 @@
...
+	allSchemes[s.ID()] = s
+	allSchemeNames[s.Name()] = s
+}
+
+// SchemeByName returns the scheme with the given name and nil if it is not
+// supported.
+func SchemeByName(name string) sign.Scheme { return allSchemeNames[name] }
+
+// AllSchemes returns all signature schemes supported.
+func AllSchemes() []sign.Scheme { a := allSchemes; return a[:] }
```
bwesterb, line 33: This allows the caller to manipulate the list. That's dangerous.

bwesterb:
I have a few issues with your changes @armfazh 

- I dislike that `SchemeByName` is now under the `github.com/cloudflare/circl/sign/api` package.
- Providing a context to a scheme that does not support it is completely ignored.  It should be an error.

pki/pki.go
```diff
@@ -0,0 +1,179 @@
...
+	"github.com/cloudflare/circl/sign"
+	"github.com/cloudflare/circl/sign/api"
+)
+
+var allSchemesByOID map[string]sign.Scheme
+var allSchemesByTLS map[uint]sign.Scheme
+
+func init() {
+	allSchemesByOID = make(map[string]sign.Scheme)
+	for _, scheme := range api.AllSchemes() {
```
bwesterb, line 19: @armfazh You can't rely on the right `init()` order.

bwesterb: https://medium.com/@markbates/go-init-order-dafa89fcef22

sign/api/api.go
armfazh: `a` is a copy, not a reference to the register

bwesterb: Yes, you're right. I didn't read it correctly.

pki/pki.go
mmcloughlin: > If a package has imports, the imported packages are initialized before initializing the package itself.

https://golang.org/ref/spec#Package_initialization

pki/pki_test.go
```diff
@@ -0,0 +1,54 @@
...
+
+import (
+	"testing"
+
+	"github.com/cloudflare/circl/pki"
+	"github.com/cloudflare/circl/sign/api"
+)
+
+func TestPEM(t *testing.T) {
+	for _, scheme := range api.AllSchemes() {
```
mmcloughlin, line 11: This seems like an appropriate use for subtests.

```go
func TestPEM(t *testing.T) {
	for _, scheme := range api.AllSchemes() {
		t.Run(scheme.Name(), func(t *testing.T) {
			// ...
		}
	}
}
```

sign/api/api_test.go
```diff
@@ -0,0 +1,94 @@
...
+	"fmt"
+	"testing"
+
+	"github.com/cloudflare/circl/sign"
+	"github.com/cloudflare/circl/sign/api"
+)
+
+func TestApi(t *testing.T) {
+	allSchemes := api.AllSchemes()
+	for _, scheme := range allSchemes {
```
mmcloughlin, line 14: Consider subtests?

sign/ed25519/ed25519_test.go
```diff
@@ -39,6 +39,24 @@ func TestMalleability(t *testing.T) {
...
+func TestPublic(t *testing.T) {
+	var zero zeroReader
+	pub, priv, err := ed25519.GenerateKey(zero)
+	if !priv.Equal(priv) {
+		t.Fatal()
+	}
+	if !pub.Equal(pub) {
+		t.Fatal()
+	}
+	if err != nil {
```
mmcloughlin, line 51: Why is the error check not the first assertion?

sign/ed25519/ed25519_test.go
```diff
@@ -39,6 +39,24 @@ func TestMalleability(t *testing.T) {
...
 }
 
+func TestPublic(t *testing.T) {
+	var zero zeroReader
+	pub, priv, err := ed25519.GenerateKey(zero)
+	if !priv.Equal(priv) {
+		t.Fatal()
+	}
+	if !pub.Equal(pub) {
+		t.Fatal()
```
mmcloughlin, line 49: Why no message with `Fatal`? If you intend to use no message, just use `FailNow()` directly.

https://golang.org/pkg/testing/#T.Fatal

sign/ed25519/signapi.go
```diff
@@ -0,0 +1,75 @@
+package ed25519
```
mmcloughlin, line 1: Nearly identical to the other `signapi.go` files (I know there are some differences). Is it generated? If so it needs a generated comment.

sign/ed25519/signapi.go
```diff
@@ -0,0 +1,75 @@
...
+}
+
+func (scheme) GenerateKey() (sign.PublicKey, sign.PrivateKey, error) {
+	return GenerateKey(rand.Reader)
+}
+
+func (scheme) Sign(
+	sk sign.PrivateKey,
+	message []byte,
+	opts *sign.SignatureOpts) []byte {
```
mmcloughlin, line 33: nit: my preference would be to use a trailing comma in multi-line signatures.

sign/sign.go
```diff
@@ -0,0 +1,107 @@
...
+
+	// Size of signatures
+	SignatureSize() int
+
+	// Size of seeds
+	SeedSize() int
+}
+
+var (
+	// ErrType is
```
mmcloughlin, line 99: FYI, if you don't want to document these individually, I _think_ the linter accepts just a comment on the entire `var` block.

sign/sign.go
```diff
@@ -0,0 +1,107 @@
...
+	ErrType = errors.New("types mismatch")
```
mmcloughlin, line 100: nit: `ErrTypeMismatch`?

bwesterb:
@armfazh Will you make changes according to the comments or shall I make them?

armfazh:
> @armfazh Will you make changes according to the comments or shall I make them?

Please, go forward with changes.

sign/api/api.go
bwesterb: @armfazh I removed `SchemeID` without changing the rest of the API at all.

pki/pki_test.go
bwesterb: Nice, didn't know about these.  Thanks!

sign/ed25519/ed25519_test.go
bwesterb: Does it really matter?  (Changed anyway.)

sign/ed25519/ed25519_test.go
bwesterb: That's a bug. Thanks for pointing that out.

sign/ed25519/signapi.go
bwesterb: No, it's not.

sign/ed25519/signapi.go
bwesterb: Yes, that's nicer.

bwesterb:
@armfazh 

> * I dislike that `SchemeByName` is now under the `github.com/cloudflare/circl/sign/api` package.

I don't see a way to solve this without making something else ugly.

> * Providing a context to a scheme that does not support it is completely ignored.  It should be an error.

I've changed that behaviour: provided a context where none is supported causes an error.

bwesterb:
I've also removed the `Hash` from the SignatureOpts as it's not used.

sign/api.go
```diff
@@ -0,0 +1,413 @@
...
+		&pem.Block{
+			Type:  fmt.Sprintf("%s PRIVATE KEY", sk.Scheme().Name()),
+			Bytes: data,
+		},
+	)
+	return str, nil
+}
+
+func UnmarshalPKIXPrivateKey(data []byte) (PrivateKey, error) {
+	var pkix struct {
```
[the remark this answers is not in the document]
bwesterb: For the private keys it is (and so I made a common type) — for the public keys it isn't.

sign/api.go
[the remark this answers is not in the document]
bwesterb: `api.AllSchemes()`

sign/api.go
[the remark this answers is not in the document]
bwesterb: @armfazh Removed case insensitivity again. Did you do that on purpose @armfazh ?

sign/eddilithium3.go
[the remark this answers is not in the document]
bwesterb: I don't understand — could you rephrase perhaps?

bwesterb:
@armfazh @claucece I've addressed all comments.  Before starting with final polish (i.e. better documentation & examples), I think we still need to discuss some API issues:

- [ ] Are we really happy with having `sign` package which defines `Scheme` and `sign/api` package which actually lists them?  Isn't `schemes` perhaps a better name than `api` as the `sign` package itself defines the API.  Then `schemes.All()` and `scheme.ByName()` would be nice to write.
- [ ] Do we want Scheme names to be case insensitive?  This was the case but @armfazh reversed it. This is required for easy integration with `cfssl`.
- [ ] Doesn't ed25519 have a variant that support contexts?  Should we add this to the Ed25519 scheme or have a separate Ed25519Ctx scheme?  I guess it's best to have separate schemes for these variants — @claucece what do you think?

armfazh:
> * [ ]  Are we really happy with having `sign` package which defines `Scheme` and `sign/api` package which actually lists them?  Isn't `schemes` perhaps a better name than `api` as the `sign` package itself defines the API.  Then `schemes.All()` and `scheme.ByName()` would be nice to write.

Agree on `s/api/schemes/` 

> * [ ]  Do we want Scheme names to be case insensitive?  This was the case but @armfazh reversed it. This is required for easy integration with `cfssl`.

I don't have a strong opinion on this. Feel free to bring back that code.

> * [ ]  Doesn't ed25519 have a variant that support contexts?  Should we add this to the Ed25519 scheme or have a separate Ed25519Ctx scheme?  I guess it's best to have separate schemes for these variants — @claucece what do you think?

That variant would be another named scheme `Ed25519Ctx`, which is separated from the current `Ed25519` scheme.

> I've also removed the Hash from the SignatureOpts as it's not used — we can add it back in if we have a scheme that actually uses it.

I added that the Hash field to make the structure compatible with the crypto.SignerOpts. So, we can receive a crypto.SignerOpts, and internally convert it to a different type.

> I dislike that SchemeByName is now under the github.com/cloudflare/circl/sign/api package.

I also didn't like it at all. But, that's the way I found for avoiding circle dependencies. Renaming sign/api to sign/schemes can alleviate a bit its usage.

sign/ed25519/signapi.go
```diff
@@ -0,0 +1,84 @@
+package ed25519
+
+import (
+	"crypto/rand"
+	"encoding/asn1"
+
+	"github.com/cloudflare/circl/sign"
+)
+
+var Scheme sign.Scheme = &scheme{}
```
armfazh, line 10: I don't understand why to take the address of a zero-sized value. You can implement the interface with `scheme` and not with `*scheme`.

sign/sign.go
```diff
@@ -0,0 +1,103 @@
...
+	PrivateKeySize() int
+
+	// Size of signatures
+	SignatureSize() int
+
+	// Size of seeds
+	SeedSize() int
+
+	// Returns whether contexts are supported
+	SupportsContext() bool
```
armfazh, line 80: I am not sure whether this method is required. As per the policy mentioned above, if context is provided and not required then it should fail.

bwesterb: It's nice to know beforehand whether a scheme will by useable when you require a context.  (I need this in the tests to check whether to provide a context or not.)

armfazh: Is this the only use case? I think this issue was one of the reasons of the previous policy. A scheme uses the context if the scheme supports it, otherwise it is simply ignored. 
Anyway, either policy works.

bwesterb: Ignoring context if the scheme doesn't support it is very dangerous: a protocol might essentially require different signatures for different contexts.  If we ignore the context for schemes that don't support it, we leave a foot gun.

sign/ed25519/signapi.go
bwesterb: Under the hood an instance of an interface is always a pair of a pointer and a concrete type, so instantiating it with a pointer is closer to what really happens.  The assembly generated btw is exactly [the same](https://godbolt.org/z/6zq31x).

armfazh: The difference is on the assembly calls.
     `call  AX` vs `call    "".(*Impl2).A(SB)`
one call is with a register and the other is with a pointer. 

I am not convinced why to use pointer when they are not needed.

bwesterb: Oops, I simplified my example too much.

[Take a look at this one](https://godbolt.org/z/KsaccE)

Creating an instance of an interface with either a empty struct or a pointer to an empty struct results in the pointer `runtime.zerobase`.  (The code of `alloc1` and `alloc2` is the same.  This time I checked properly.)

The generic way to run a function on an interface involves a virtual call (`call ax`).  So interestingly the Go compiler optimised the call with the pointer receiver, but not the call with the value receiver.  I don't see a reason why it couldn't optimise the one with the value receiver as well and perhaps it will in the future.

So using pointers or not does not actually change anything when using empty structs (and the calling code doesn't know the types.) And if it does it optimizes the pointer case better. That's not the reason I use pointers though — as I said I prefer putting them here as it's closer to what Go actually does under the hood: an instance of an interface is always a pointer.

bwesterb:
> Agree on `s/api/schemes/`

Done.  Updated this branch, Go fork and cfssl fork.

> I added that the Hash field to make the structure compatible with the crypto.SignerOpts. So, we can receive a crypto.SignerOpts, and internally convert it to a different type.

The type of `Scheme.Sign` is `Sign(sk PrivateKey, message []byte, opts *SignatureOpts) []byte` so that wouldn't accept a `crypto.SignerOpts` anyway now.  (Note that there are two sign functions: one on the scheme and one on the PrivateKey — the one on the private key does accept a `crypto.SignerOpts` but does not allow for the advanced options like `Context`.)

armfazh:
> the one on the private key does accept a crypto.SignerOpts but does not allow for the advanced options like Context.)

I had the same intuition, but here is the trick. Even if we announce to receive a `crypto.SignerOpts`, we could pass a `Sign.SignatureOpts` (which must have field Hash) and internally convert the parameter into a `Sign.SignatureOpts` struct. (I learn this trick from @claucece)

sign/sign.go
armfazh: good point.

sign/ed25519/signapi.go
armfazh: > So interestingly the Go compiler optimised the call with the pointer receiver, but not the call with the value receiver.

I want to learn more about this.

> So using pointers or not does not actually change anything when using empty structs.

Agree.

claucece:
>  Are we really happy with having sign package which defines Scheme and sign/api package which actually lists them? Isn't schemes perhaps a better name than api as the sign package itself defines the API. Then schemes.All() and scheme.ByName() would be nice to write.

Yes! 

> Do we want Scheme names to be case insensitive? This was the case but @armfazh reversed it. This is required for easy integration with cfssl.

Mmm... It will be better case insensitive.

> Doesn't ed25519 have a variant that support contexts? Should we add this to the Ed25519 scheme or have a separate Ed25519Ctx scheme? I guess it's best to have separate schemes for these variants — @claucece what do you think?

There are two: the ctx and the ph. The schemes should be there on the ed25519 API.

claucece:
> I had the same intuition, but here is the trick. Even if we announce to receive a crypto.SignerOpts, we could pass a Sign.SignatureOpts (which must have field Hash) and internally convert the parameter into a Sign.SignatureOpts struct. (I learn this trick from @claucece)

Exactly! Then, the "Golang-way" API does not change ;)

claucece:
@bwesterb check what I did with the ed448 and ed25519 API, specially the usage of this: https://github.com/cloudflare/circl/blob/master/sign/ed25519/ed25519.go#L69

sign/ed25519/ed25519_test.go
bwesterb: It does matter: FailNow doesn't show the line number when triggered and Fatal does.  Thus I reverted back to using Fatal.

bwesterb:
> I had the same intuition, but here is the trick. Even if we announce to receive a `crypto.SignerOpts`, we could pass a `Sign.SignatureOpts` (which must have field Hash) and internally convert the parameter into a `Sign.SignatureOpts` struct. (I learn this trick from @claucece)

Yes, that's for a method on the private key, but here we're dealing with a top level function.

(For the `Sign` method on the private key it's actually a bit tricky.  If we use the same type for an ed25519 and ed25519ctx private key, then we can't know whether it was used for the scheme ed25519 or ed25519ctx.  I'd like to propose to fix this in a coming pull request where we add schemes for the variants of ed*.)

bwesterb:
@armfazh @claucece I'm happy with the PR as it is now (assuming we'll be able to make changes to the API before we tag it.) Do you agree?

claucece:
> (For the Sign method on the private key it's actually a bit tricky. If we use the same type for an ed25519 and ed25519ctx private key, then we can't know whether it was used for the scheme ed25519 or ed25519ctx. I'd like to propose to fix this in a coming pull request where we add schemes for the variants of ed*.)

Mmm.. ok.

> Do you agree?

Sure!

pki/pki.go
```diff
@@ -0,0 +1,184 @@
...
+	})
+}
+
+func MarshalPEMPrivateKey(sk sign.PrivateKey) ([]byte, error) {
+	data, err := MarshalPKIXPrivateKey(sk)
+	if err != nil {
+		return nil, err
+	}
+	str := pem.EncodeToMemory(&pem.Block{
+		Type:  fmt.Sprintf("%s PRIVATE KEY", sk.Scheme().Name()),
```
armfazh, line 158:
```suggestion
		Type:  sk.Scheme().Name() + " PRIVATE KEY",
```

armfazh: During unmarshalling, does this name is verified that matches with the content?

pki/pki.go
```diff
@@ -0,0 +1,184 @@
...
+	return asn1.Marshal(struct {
+		pkix.AlgorithmIdentifier
+		asn1.BitString
+	}{
+		pkix.AlgorithmIdentifier{
+			Algorithm: scheme.(CertificateScheme).Oid(),
+		},
+		asn1.BitString{
+			Bytes:     data,
+			BitLength: len(data) * 8,
```
armfazh, line 147: Do we need to verify that`scheme.PublicKeySize() == len(data)` ?

pki/pki.go
```diff
@@ -0,0 +1,184 @@
...
+func init() {
+	allSchemesByOID = make(map[string]sign.Scheme)
+	for _, scheme := range schemes.All() {
+		if cert, ok := scheme.(CertificateScheme); ok {
+			allSchemesByOID[cert.Oid().String()] = scheme
+		}
+	}
+
+	allSchemesByTLS = make(map[uint]sign.Scheme)
+	for _, scheme := range schemes.All() {
```
armfazh, line 33: Merge for-loops.

pki/pki.go
bwesterb: > During unmarshalling, does this name is verified that matches with the content?

No.  The reason is that I'm not sure that all private key files actually contain the type in them.

pki/pki.go
bwesterb: I don't think it's necessary here.

bwesterb:
Thank you, I merged it.  I created an issue ( #147 ) to track desired improvements to this API.

PR #145, opened 2020-08-12
Go 1.15 support

bwesterb:


bwesterb:
@Lekensteyn Could you please do the honours and review this tiny PR now @armfazh is on PTO?

PR #154, opened 2020-09-29
Implement OPRF

claucece:


oprf/group/group.go
```diff
@@ -0,0 +1,117 @@
...
+// Ciphersuite object determines the prime-order group (pog) that is used for
+// performing the OPRF operations, along with the different hash function
+// definitions.
+// Should be created using New, using a CiphersuiteID constant.
+type Ciphersuite struct {
+	// name of the ciphersuite
+	name string
+
+	// context
+	ctx string
```
chris-wood, line 28: Why is `context` part of the Ciphersuite struct?

oprf/group/group.go
```diff
@@ -0,0 +1,117 @@
...
+	// group order.
+	RandomScalar() (*Scalar, error)
+
+	// Returns the ByteLength of objects associated with the Ciphersuite
+	ByteLength() int
+}
+
+// Point is the interface that represents points in a given
+// Ciphersuite instantiation.
+type Point interface {
```
chris-wood, line 75: Should this be `Element`, to better align with pogs?

oprf/group/group.go
```diff
@@ -0,0 +1,117 @@
...
+
+	// Performs a scalar multiplication of the Generator with some scalar
+	// input
+	ScalarMultBase(*Scalar) (*Point, error)
+
+	// Performs the group addition operation on the calling Point object
+	// along with a separate Point provided as input
+	Add(*Point) (*Point, error)
+
+	Neg(*Point) (*Point, error)
```
chris-wood, line 98: nit: missing documentation

oprf/group/group.go
```diff
@@ -0,0 +1,117 @@
...
+
+// Point is the interface that represents points in a given
+// Ciphersuite instantiation.
+type Point interface {
+	// New constructs a Point object for the associated Suite
+	// instantiation
+	New(Suite) *Point
+
+	// Returns a bool indicating that the Point is valid
+	IsValid() bool
```
chris-wood, line 81: I wonder if it makes sense to have this function in the first place. Can we just ensure that all points are valid? When would we need an invalid point?

oprf/group/group.go
```diff
@@ -0,0 +1,117 @@
...
+
+	// Performs a transformation to encode bytes as a Scalar object in the
+	// appropiate group.
+	HashToScalar([]byte) (Scalar, error)
+
+	// Samples a random scalar value from the field of scalars defined by the
+	// group order.
+	RandomScalar() (*Scalar, error)
+
+	// Returns the ByteLength of objects associated with the Ciphersuite
```
chris-wood, line 69: What is an "object" here? What is this function used for?

oprf/group/group.go
```diff
@@ -0,0 +1,117 @@
...
+
+	// Returns the order of the canonical generator in the group.
+	Order() *big.Int
+
+	// Base hash function used in conjunction with the Ciphersuite
+	Hash() hash.Hash
+
+	// Returns k * G, where G is the canonical generator of the group, and k is
+	// some scalar value provided as input.
+	ScalarMult(*Scalar) (*Point, error)
```
chris-wood, line 55: Why are these functions on the Suite interface? That seems a bit odd, no?

chris-wood (changes requested):
Good start!

oprf/group/group.go
claucece: Only this function or all of them? The generator, the order, the name, etc seem reasonable as they belong to the whole suite; but I agree that having the ScalarMult there is odd..

oprf/group/group.go
claucece: The byte length of Points and Scalars, which are the objects associated with a Ciphersuite. It returns the length of them, which is often useful to check for serialisation checks.

oprf/group/group.go
claucece: Might be. I guess that at some point you might receive a public key (which is a Point) from a Client, for example, that generated it not by using our library but some other library. As it is not possible at that point to know if the public key was correctly generated, this function can be used.

oprf/group/group.go
claucece: Yes. I'll like it to be aligned with the draft as well. But I haven't seen a cryptographic library that refers to points as elements.. so might be confusing.

oprf/group/group.go
claucece: The context is defined at the start of the protocol by the SetUp function, so I was thinking that it will be good to set it globally here, so it can be reached by the several functions that use it.

oprf/group/group.go
```diff
@@ -0,0 +1,118 @@
+package group
```
armfazh, line 1: group package must live outside oprf group.
/circl/group
/circl/oprf

oprf/group/group.go
armfazh: The fact to named it element is because `Group` is an abstraction from elliptic curves.
When we add other implementations of a `Group`, e.g. decaf, the elements are not exactly points.

oprf/group/group.go
```diff
@@ -0,0 +1,118 @@
...
+	New(CiphersuiteID) (Suite, error)
+
+	// Returns the identifying name of the group
+	Name() string
+
+	// Returns the canonical (fixed) generator for defined group
+	Generator() *Point
+
+	// Returns the order of the canonical generator in the group.
+	Order() *big.Int
```
armfazh, line 48: Returns a Scalar

oprf/group/group.go
claucece: OK ;) I'll rename it element ;)

oprf/group/group.go
claucece: Yes! Just what I was going to ask ;)

oprf/oprf.go
```diff
@@ -0,0 +1,55 @@
...
+// TODO: should we add: VerifyFinalize and EvaluateBatch?
+type Server interface {
+	// SetUp configures a new server by generating its context and keys
+	SetUp(group.CiphersuiteID) (string, PrivateKey)
+
+	// Evaluate evaluates the token
+	Evaluate(sk *PrivateKey, pk *PublicKey, b *BlindToken, e *Evaluation) *Evaluation
+
+	// Finalize outputs a byte array that corresponds to its input.
+	Finalize(t *Token, issuedT *IssuedToken, info []byte) []byte
```
chris-wood, line 55: This seems to be missing `FullEvaluate` (though I don't think that's needed right now).

oprf/oprf.go
```diff
@@ -0,0 +1,55 @@
...
+// Client is
+type Client interface {
+	// SetUp configures a new client by generating its context and keys
+	SetUp(group.CiphersuiteID) (string, PrivateKey)
+
+	// Blind generates a token and blinded data.
+	Blind(in []byte) (*Token, *BlindToken)
+
+	// Unblind unblinds the server response.
+	Unblind(pk *PublicKey, t *Token, b *BlindToken, e *Evaluation) *IssuedToken
```
chris-wood, line 41: This is missing `Finalize`, right?

oprf/oprf.go
```diff
@@ -0,0 +1,55 @@
...
+// PublicKey is a struct representing a public key
+type PublicKey *group.Point
+
+// PrivateKey is a struct representing a private key
+type PrivateKey *group.Scalar
+
+// Client is
+type Client interface {
+	// SetUp configures a new client by generating its context and keys
+	SetUp(group.CiphersuiteID) (string, PrivateKey)
```
chris-wood, line 35: Why not `New`?

chris-wood: Also, what are the output values? Why does this return a `string` and `PrivateKey`?

oprf/oprf.go
```diff
@@ -0,0 +1,55 @@
...
+
+	// Unblind unblinds the server response.
+	Unblind(pk *PublicKey, t *Token, b *BlindToken, e *Evaluation) *IssuedToken
+}
+
+// Server is
+// TODO: should we add: VerifyFinalize and EvaluateBatch?
+type Server interface {
+	// SetUp configures a new server by generating its context and keys
+	SetUp(group.CiphersuiteID) (string, PrivateKey)
```
chris-wood, line 49: As above, what are the output values?

oprf/oprf.go
```diff
@@ -0,0 +1,55 @@
...
+
+	// Blind generates a token and blinded data.
+	Blind(in []byte) (*Token, *BlindToken)
+
+	// Unblind unblinds the server response.
+	Unblind(pk *PublicKey, t *Token, b *BlindToken, e *Evaluation) *IssuedToken
+}
+
+// Server is
+// TODO: should we add: VerifyFinalize and EvaluateBatch?
```
chris-wood, line 45: Maybe in a future change!

chris-wood (changes requested):

oprf/oprf.go
claucece: In Golang, the return values are defined after the method declaration, so following the definition, they are:

`SetUp(group.CiphersuiteID) (string, PrivateKey)`, meaning they are: a string and a PrivateKey. 

In Golang, there is no way to name the return values, unless you use this: https://yourbasic.org/golang/named-return-values-parameters/#:~:text=In%20Go%20return%20parameters%20may,are%20used%20as%20return%20values.&text=Named%20return%20parameters%20are%20initialized,can%20make%20for%20good%20documentation.

So, Im not really sure what you mean by what are them. As defined per spec, they are: a context and the key.  If you can, please explain me what you mean. Thanks ;)

claucece: Copying the definition from the draft:

```
def SetupBaseServer(suite):
  (skS, _) = KeyGen(GG)
  contextString = I2OSP(modeBase, 1) || I2OSP(suite.ID, 2)
  return ServerContext(contextString, skS)
```

It returns a context, which is a string, and the private key.

oprf/oprf.go
claucece: Copying the definition from the draft:

```
def SetupBaseServer(suite):
  (skS, _) = KeyGen(GG)
  contextString = I2OSP(modeBase, 1) || I2OSP(suite.ID, 2)
  return ServerContext(contextString, skS)
```

It returns a context, which is a string, and the private key.

oprf/oprf.go
claucece: Yes, it is. As this was summited prior to the discussion on the chat. I'll wait to put it there once it definitely lands on the draft.

oprf/oprf.go
claucece: The draft calls it setup and that is what I was following. It can be renamed to new as well.

oprf/oprf.go
armfazh: You can return a `Server` that implements `Evaluate` and `FullEvaluate`, the server structure is composed of two fields: the contextString and a skS.
For this to work, `SetupServer` must be a simple function with no receiver.

claucece:
@armfazh @chris-wood check how everything is looking now. I have some unpushed tests as they fail. 
@armfazh want to have a meeting to check how will be good to move the group functionality outside? I have some thoughts now after checking the p384 implementation here..

oprf/group/group.go
chris-wood: Oh, my point was that we can simply fail deserializing such points, rather than deserialize them and then require applications to check "IsValid". Would that work?

oprf/group/group.go
```diff
@@ -0,0 +1,344 @@
...
+type Element interface {
+	// Returns a bool indicating that the Element is valid
+	IsValid() bool
+
+	// Returns a bool indicating that the Element is the identity
+	IsIdentity() bool
+
+	// Performs a scalar multiplication of the Element with some scalar
+	// input
+	ScalarMult(*Scalar) (*Point, error)
```
chris-wood, line 28: The name of the interface is `Element`, but here we return `Point`. Is there a reason for the discrepancy?

oprf/group/group.go
```diff
@@ -0,0 +1,344 @@
...
+	// TODO: should we keep it? Seems only be used on the Verified case..
+	HashToScalar([]byte) (Scalar, error)
+
+	// Samples a random scalar value from the field of scalars defined by the
+	// group order.
+	RandomScalar() (*Scalar, error)
+
+	// Performs a scalar multiplication of the Generator with some scalar
+	// input
+	ScalarMultBase(*Scalar) (*Point, error)
```
chris-wood: Why is this a function on the `Suite` interface?

oprf/group/group.go
chris-wood: Do these have the same lengths? Do you have an example for how this might be used?

oprf/group/suite.go
```diff
@@ -0,0 +1,201 @@
...
+	// TODO: should we keep it? Seems only be used on the Verified case..
+	HashToScalar([]byte) (Scalar, error)
+
+	// Samples a random scalar value from the field of scalars defined by the
+	// group order.
+	RandomScalar() (*Scalar, error)
+
+	// Performs a scalar multiplication of the Generator with some scalar
+	// input
+	ScalarMultBase(*Scalar) (*Point, error)
```
chris-wood, line 56: Why does this function exist on the `Suite` interface?

oprf/group/suite.go
```diff
@@ -0,0 +1,201 @@
...
+	Order() *Scalar
+
+	// Performs a transformation to encode bytes as a Element object in the
+	// group.
+	HashToGroup([]byte) (*Point, error)
+
+	// Performs a transformation to encode bytes as a Scalar object in the
+	// appropiate group.
+	// TODO: should we keep it? Seems only be used on the Verified case..
+	HashToScalar([]byte) (Scalar, error)
```
chris-wood, line 48: Let's add it when the time comes. We can remove it for now. :)

oprf/group/suite.go
```diff
@@ -0,0 +1,201 @@
...
+
+	"github.com/armfazh/h2c-go-ref"
+)
+
+// Ciphersuite corresponds to the OPRF ciphersuite that is chosen. The
+// Ciphersuite object determines the prime-order group (pog) that is used for
+// performing the OPRF operations, along with the different hash function
+// definitions.
+// Should be created using NewSuite, using an string.
+type Ciphersuite struct {
```
chris-wood, line 18: An OPRF ciphersuite also requires a specific hash function -- should we add that here?

oprf/group/suite.go
```diff
@@ -0,0 +1,201 @@
...
+
+// HashToScalar performs a transformation to encode bytes as a Scalar object in the
+// appropiate group.
+func (c *Ciphersuite) HashToScalar([]byte) (*Scalar, error) {
+	return &Scalar{}, nil
+}
+
+// RandomScalar samples a random scalar value from the field of scalars defined by the
+// group order.
+// TODO: not constant time
```
chris-wood, line 146: Do we intend to make it constant time soon?

oprf/group/suite.go
```diff
@@ -0,0 +1,201 @@
...
+}
+
+// NewSuite creates a new ciphersuite for the requested name.
+func NewSuite(name string) (*Ciphersuite, error) {
+	cSuite := &Ciphersuite{}
+
+	switch name {
+	case "P-256":
+		cSuite.name = "OPRFP256-SHA512-ELL2-RO"
+		cSuite.dst = "RFCXXXX-P256_XMD:SHA-512_SSWU_RO_"
```
chris-wood, line 194: I think there should be no trailing "_". Per the draft, the DST is "RFCXXXX-P256XMD:SHA-256_SSWU_RO".

oprf/oprf.go
```diff
@@ -0,0 +1,267 @@
...
+
+// ServerContext implements the functionality of a Server.
+// TODO: add FullEvaluate
+type ServerContext interface {
+	// Evaluate evaluates the token
+	Evaluate(b BlindToken) (*Evaluation, error)
+}
+
+func generateCtx(suiteID uint16) string {
+	ctx := "000"
```
chris-wood, line 95: nit: can we make this ("modeBase") a named constant?

oprf/oprf.go
```diff
@@ -0,0 +1,267 @@
...
+}
+
+// Unblind unblinds the server response.
+func (c *Client) Unblind(t *Token, e *Evaluation) (IssuedToken, error) {
+	p := group.NewPoint(c.suite)
+	p.Deserialize(e.element)
+
+	r := t.blind
+	n := c.suite.Order()
+	// TODO: as to not export the internal field of the scalar, this can be encapsulated
```
chris-wood, line 224: 👍 Yeah, great idea! Why not add a `ModInverse` function to the `Scalar` interface?

oprf/group/suite.go
chris-wood: (The Sage code is wrong here!)

oprf/group/suite.go
claucece: Ok ;)

oprf/oprf.go
claucece: Yes! Just did it.

oprf/group/suite.go
claucece: Yes.. I thought so as well. But it seems like they all use `SHA-512`?

oprf/group/group.go
claucece: Just changed it ;)

oprf/group/group.go
claucece: I thought of putting it on the Element one.. but I'm not sure if it belongs to Points or Scalars..

oprf/group/suite.go
chris-wood: For now :) We will likely change this in a version soon.

claucece: Ah, gotta it. I'll put it there ;)

oprf/group/suite.go
claucece: Not so soon. It is the rejection part that it is not constant.

oprf/group/suite.go
claucece: Oh! I'll change. Maybe tomorrow I'll change some parts of the sage code as well, that I found.

oprf/oprf.go
claucece: Sure!

oprf/group/suite.go
claucece: I thought of putting it on the Element one.. but I'm not sure if it belongs to Points or Scalars..

oprf/group/suite.go
claucece: Mmm.. so I find that the dst in the draft is: `RFCXXXX-P256_XMD:SHA-256_SSWU_RO_` with the trailing.. or maybe it is a different draft version?

oprf/group/group.go
claucece: Let's do this in the different PR as discussed.

oprf/oprf.go
```diff
@@ -0,0 +1,273 @@
...
+	return IssuedToken(iToken), nil
+}
+
+// Finalize outputs a byte array that corresponds to the client input.
+func (c *Client) Finalize(t *Token, issuedT IssuedToken, info []byte) []byte {
+	h := c.suite.Hash
+
+	lenBuf := make([]byte, 2)
+	binary.BigEndian.PutUint16(lenBuf, uint16(len(t.data)))
+	h.Write(lenBuf) //nolint:errcheck
```
armfazh, line 255: change for 
```suggestion
	_, _ = h.Write(lenBuf)
```

oprf/oprf_test.go
```diff
@@ -0,0 +1,168 @@
+package oprf
```
armfazh, line 1:
```suggestion
package oprf_test
```

oprf/group/group.go
```diff
@@ -0,0 +1,188 @@
...
+	// Returns a bool indicating that the Point is the identity
+	IsIdentity() bool
+
+	// Performs a scalar multiplication of the Point with some scalar
+	// input
+	ScalarMult(*Scalar) (*Point, error)
+
+	// Performs the addition operation on the calling Point object
+	// along with a separate Point provided as input
+	Add(*Point) (*Point, error)
```
armfazh, line 31: All these arithmetic operations never return error.
Only Deserialize could return an error.

oprf/group/group.go
```diff
@@ -0,0 +1,188 @@
...
+	q := NewPoint(p.c)
+	q.x, q.y = p.c.curve.ScalarMult(p.x, p.y, s.Serialize())
+
+	return q, nil
+}
+
+// Add performs the addition operation on the calling Point object
+// along with a separate Point provided as input
+func (p *Point) Add(q *Point) (*Point, error) {
+	if !p.IsValid() || !q.IsValid() {
```
armfazh, line 72: No need to check whether point is valid, since every instance of ~point~ element must be valid all the time.

oprf/group/group.go
```diff
@@ -0,0 +1,188 @@
...
+
+// Scalar is an struct representing a field element
+type Scalar struct {
+	c *Ciphersuite
+	x *big.Int
+}
+
+// FieldElement is the interface that represents scalars in a given
+// Ciphersuite instantiation.
+type FieldElement interface {
```
armfazh, line 143: This name is confusing, I suggest using Scalar for the interface and `scalar`  for the structure.

armfazh (approved):
I approve changes for OPRF, however, more work is needed in the group package. 
As discussed internally, let's move this forward, and submit a follow up PR addressing the comments about group package.

oprf/group/group.go
claucece: Changed ;)

oprf/group/group.go
claucece: Changed ;)

oprf/group/group.go
claucece: Changed ;)

oprf/oprf_test.go
claucece: Mmm.. but then the oprf functions will not be available to this package..

armfazh: You need to import the package, as any other user will do

claucece: Done.

oprf/oprf.go
```diff
@@ -0,0 +1,238 @@
...
+
+// Server is a representation of a Server during protocol execution.
+type Server struct {
+	suite *group.Ciphersuite
+	ctx   string
+	K     *KeyPair
+}
+
+// ServerContext implements the functionality of a Server.
+// TODO: add FullEvaluate
```
chris-wood, line 85: Can we remove this TODO?

oprf/oprf.go
```diff
@@ -0,0 +1,238 @@
...
+
+	// Finalize outputs a byte array that corresponds to its input.
+	Finalize(t *Token, issuedT IssuedToken, info []byte) []byte
+}
+
+// Server is a representation of a Server during protocol execution.
+type Server struct {
+	suite *group.Ciphersuite
+	ctx   string
+	K     *KeyPair
```
chris-wood, line 81: Do we need to export `K`? (Also, can we rename it to something more clear, perhaps `Key` (or `key`)?

oprf/oprf.go
```diff
@@ -0,0 +1,238 @@
...
+	pubK := suite.ScalarMultBase(privK)
+
+	return &KeyPair{pubK, privK}, nil
+}
+
+func suiteFromID(suiteID uint16) (*group.Ciphersuite, error) {
+	var err error
+	var suite *group.Ciphersuite
+
+	// TODO: add other suites
```
chris-wood, line 112: Can we remove this TODO?

oprf/oprf.go
```diff
@@ -0,0 +1,238 @@
...
+	"fmt"
+
+	"github.com/cloudflare/circl/oprf/group"
+)
+
+const (
+	// OPRFCurve25519 is the constant to represent the OPRF curve25519 with SHA-512 (ELL2-RO) group.
+	OPRFCurve25519 uint16 = 0x0001
+	// OPRFCurve448 is the constant to represent the OPRF curve448 with SHA-512 (ELL2-RO) group.
+	OPRFCurve448 uint16 = 0x0002
```
chris-wood, line 15:
```suggestion
```

(since these suites are not yet supported)

oprf/oprf.go
```diff
@@ -0,0 +1,238 @@
...
+	OPRFP256 uint16 = 0x0003
+	// OPRFP384 is the constant to represent the OPRF P-384 with SHA-512 (SSWU-RO) group.
+	OPRFP384 uint16 = 0x0004
+	// OPRFP521 is the constant to represent the OPRF P-521 with SHA-512 (SSWU-RO) group.
+	OPRFP521 uint16 = 0x0005
+)
+
+const (
+	// OPRFMode is the context string to define a OPRF.
+	OPRFMode string = "000"
```
chris-wood, line 26: Hmm, the mode string is supposed to be encoded in a single octet. Why are these three zeroes?

oprf/oprf.go
```diff
@@ -0,0 +1,238 @@
...
+}
+
+// Evaluation corresponds to the evaluation over a token.
+type Evaluation struct {
+	element []byte
+}
+
+// KeyPair is an struct containing a public and private key.
+type KeyPair struct {
+	PubK  *group.Element
```
chris-wood, line 55: Do these need to be exported?

oprf/oprf.go
```diff
@@ -0,0 +1,238 @@
...
+// ClientContext implements the functionality of a Client.
+type ClientContext interface {
+	// Blind generates a token and blinded data.
+	Blind(in []byte) (*Token, *BlindToken, error)
+
+	// Unblind unblinds the server response.
+	Unblind(t *Token, e *Evaluation) (IssuedToken, error)
+
+	// Finalize outputs a byte array that corresponds to its input.
+	Finalize(t *Token, issuedT IssuedToken, info []byte) []byte
```
chris-wood, line 74: Do these need to be two separate functions? Or can we merge them, e.g., into an API like so:

```
type ClientRequestContext interface {
   Finalize(e *Evaluation, info []byte) ([]byte, error)
}

type ClientContext interface {
   Request(in []byte) (*BlindToken, ClientRequestContext)
}
```

Clients would then use this like so:

```
input := ... // []byte
token, requestContext := client.Request(input)
// send token, get back `evaluation`
output, err := requestContext.Finalize(evaluation, []byte("info"))
// process err and output
```

I think this would be less error prone, since the implementation takes care of the OPRF details.

oprf/oprf.go
```diff
@@ -0,0 +1,238 @@
...
+// TODO: add FullEvaluate
+type ServerContext interface {
+	// Evaluate evaluates the token.
+	Evaluate(b BlindToken) *Evaluation
+}
+
+func generateCtx(suiteID uint16) string {
+	ctx := OPRFMode + fmt.Sprintf("%x", suiteID)
+
+	return ctx
```
chris-wood, line 94:
```suggestion
        return OPRFMode + fmt.Sprintf("%x", suiteID)
```

oprf/oprf.go
```diff
@@ -0,0 +1,238 @@
...
+}
+
+func generateKeys(suite *group.Ciphersuite) (*KeyPair, error) {
```
chris-wood, line 97:
```suggestion
func generateKeyPair(suite *group.Ciphersuite) (*KeyPair, error) {
```

oprf/oprf.go
```diff
@@ -0,0 +1,238 @@
...
+	privK, err := suite.RandomScalar()
+	if err != nil {
+		return nil, err
```
chris-wood, line 100: Should this be fallible? Why would we fail to generate a scalar here? Can we `panic` instead?

oprf/oprf.go
```diff
@@ -0,0 +1,238 @@
...
+func (s *Server) Evaluate(b BlindToken) *Evaluation {
+	p := group.NewElement(s.suite)
+	p.Deserialize(b)
+
+	z := p.ScalarMult(s.K.PrivK)
+
+	ser := z.Serialize()
+
+	eval := &Evaluation{ser}
+	return eval
```
chris-wood, line 162:
```suggestion
	return &Evaluation{ser}
```

oprf/oprf.go
```diff
@@ -0,0 +1,238 @@
...
+	server.ctx = ctx
+	server.K = keyPair
+
+	return server, nil
+}
+
+// Evaluate creates an evaluation of the blided token.
+func (s *Server) Evaluate(b BlindToken) *Evaluation {
+	p := group.NewElement(s.suite)
+	p.Deserialize(b)
```
chris-wood, line 155: What if deserialization fails? Should we not return an error?

oprf/oprf.go
```diff
@@ -0,0 +1,238 @@
...
+		return nil, err
+	}
+
+	ctx := generateCtx(suiteID)
+
+	client := &Client{}
+	client.suite = suite
+	client.ctx = ctx
+
+	return client, nil
```
chris-wood, line 178:
```suggestion
	return &Client{
	    suite: suite,
	    ctx: ctx,
	}, nil
```

oprf/oprf_test.go
```diff
@@ -0,0 +1,150 @@
+package oprf_test
```
chris-wood, line 1: Can we add tests against the test vectors?

chris-wood (changes requested):
I've not reviewed the Group API (since I understand it's subject to change). This is looking good! I left some suggestions for improvement. @armfazh, please have a look.

armfazh:
> This is looking good! I left some suggestions for improvement. @armfazh, please have a look.

Agree with most of your comments, few changes needed.

chris-wood:
@claucece please let us know if you need *any* help addressing this feedback!

oprf/oprf.go
claucece: Sure.

oprf/oprf.go
claucece: At some point, the HIBPvN library needed the serialized private key, so maybe it is needed. I haven't checked if OPAQUE needs the keys.

oprf/oprf.go
claucece: Sure.

oprf/oprf.go
claucece: I'm not sure. Do the OPAQUE or HIBPvN libraries will need them?

oprf/oprf.go
claucece: How will it fail?

oprf/oprf_test.go
claucece: I'll change the sage code first.

oprf/oprf.go
claucece: Ok, so I think I'm going to clarify here what i think this actually means. This is referring to the 3 methods of the ClientContext not only the 2 last ones. The idea is to change:

```
	// Blind generates a token and blinded data.
	Blind(in []byte) (*Token, *BlindToken, error)
```

by 

```
   Request(in []byte) (*Token, *BlindToken, ClientRequestContext)
```

and:

```
	Unblind(t *Token, e *Evaluation) (IssuedToken, error)

	// Finalize outputs a byte array that corresponds to its input.
	Finalize(t *Token, issuedT IssuedToken, info []byte) []byte
```

by 

```
type ClientRequestContext interface {
   Finalize(t *Token, e *Evaluation, info []byte) ([]byte, error)
}
```

is it correct? 

Note that `t *Token` is still needed..

oprf/oprf.go
claucece: Only when ReadFull fails. We can panic, yes.

claucece: Note that usually in the generation of keys in Golang, it errors not panic: https://golang.org/src/crypto/ed25519/ed25519.go?s=3257:3320#L88

oprf/oprf.go
```diff
@@ -0,0 +1,222 @@
...
+	}
+	if err != nil {
+		return nil, err
+	}
+
+	return suite, nil
+}
+
+// NewServer creates a new instantiation of a Server.
+func NewServer(suiteID uint16) (*Server, error) {
```
claucece, line 119: @tomrist see here.

oprf/oprf_test.go
claucece: The were just added. There is some difference, that I'll explain in a call.

oprf/oprf.go
claucece: Fixed now.

oprf/oprf.go
claucece: Fixed now.

oprf/oprf.go
claucece: I think I addressed this in a simpler way. But let me know what you think.

chris-wood:
cc @tomrist

oprf/oprf_test.go
```diff
@@ -0,0 +1,230 @@
...
+	}
+}
+
+func TestClientBlind(t *testing.T) {
+	client, err := NewClient(OPRFP256)
+	if err != nil {
+		t.Fatal("invalid setup of client: " + err.Error())
+	}
+
+	token, bToken, err := client.Blind([]byte{00})
```
armfazh, line 41: (just an idea) 
an interesting test would be to show the ability that client can do `unblind(blind(x)) == x` .
but it's not possible with the current api.

oprf/oprf_test.go
```diff
@@ -0,0 +1,230 @@
...
+		t.Fatal("invalid evaluation of server: no evaluation.")
+	}
+
+	info := []byte("test information")
+	iToken, h, _ := client.Finalize(token, eval, info)
+
+	// From the test vectors
+	testIToken, _ := hex.DecodeString("0277f2ded1d0cbbc9a71504f94dd0aa997709f7adb3368631392313164273eb340")
+
+	if !bytes.Equal(testIToken[:], iToken[:]) {
```
armfazh, line 221: not a must, but happy if the circl/internal/testing package is used for error reporting.

oprf/oprf.go
```diff
@@ -0,0 +1,229 @@
+package oprf
```
armfazh, line 1: Make sure to include a brief documentation of the package.

oprf/group/suite.go
armfazh: In a subsequent PR, we can use `hashToField` to sample scalars.

armfazh (commented):
Some nits on names only.

oprf/oprf.go
tomrist: integer IDs for ciphersuites sounds much better. Are these constant values defined in RFC?

oprf/oprf.go
```diff
@@ -0,0 +1,229 @@
...
+// Client is a representation of a Client during protocol execution.
+type Client struct {
+	suite *group.Ciphersuite
+	ctx   []byte
+}
+
+// ClientContext implements the functionality of a Client.
+type ClientContext interface {
+	// Request generates a token and blinded data.
+	Request(in []byte) (*Token, *BlindToken, error)
```
chris-wood, line 64: Why is this a function on `ClientContext`, instead of `Client`?

oprf/oprf.go
```diff
@@ -0,0 +1,229 @@
...
+		return nil, []byte{}, err
+	}
+
+	r := t.blind
+	rInv := r.Inv()
+
+	tt := p.ScalarMult(rInv)
+	iToken := tt.Serialize()
+
+	h := group.FinalizeHash(c.suite, t.data, iToken, info, c.ctx)
```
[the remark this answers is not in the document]
armfazh: group.FinalizeHash

oprf/oprf.go
claucece: Yes they are. They are on the suites section.

oprf/oprf.go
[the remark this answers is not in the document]
claucece: I agree. I have an idea of making the inside of this functions part of an 'internal' folder, so they are not exposed, but they don't belong to the group.

oprf/oprf.go
claucece: So, I'm making the pub part of the KeyPair private, but the private key should be able to be serialized.

oprf/oprf_test.go
claucece: Making it part of another package makes it difficult to test functions or parameters that are not exported. So, I'll prefer to keep it part of the same package.

oprf/oprf.go
```diff
@@ -0,0 +1,229 @@
...
+	if n == 0 || octetStringSize > n {
+		panic("I2OSP error: integer too large")
+	}
+
+	subtle.ConstantTimeCopy(1, result[:n-octetStringSize], result[:n-octetStringSize])
+	subtle.ConstantTimeCopy(1, result[n-octetStringSize:], octetString)
+	return result
+}
+
+func generateCtx(suiteID uint16) []byte {
```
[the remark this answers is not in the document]
claucece: Thanks!

oprf/oprf.go
```diff
@@ -0,0 +1,229 @@
...
+	"crypto/subtle"
+	"errors"
+	"math/big"
+
+	"github.com/cloudflare/circl/oprf/group"
+)
+
+const (
+	// OPRFP256 is the constant to represent the OPRF P-256 with SHA-512 (SSWU-RO) group.
+	OPRFP256 uint16 = 0x0003
```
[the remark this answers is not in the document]
claucece: Sure.

oprf/oprf.go
```diff
@@ -0,0 +1,229 @@
...
+	}
+	if err != nil {
+		return nil, err
+	}
+
+	return suite, nil
+}
+
+// NewServer creates a new instantiation of a Server.
+func NewServer(suiteID uint16) (*Server, error) {
```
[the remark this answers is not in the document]
claucece: As discussed, I'll better remove the interface, and keep the methods and structs ;)

oprf/oprf.go
```diff
@@ -0,0 +1,229 @@
...
+	client.ctx = ctx
+
+	return &Client{
+		suite: suite,
+		ctx:   ctx,
+	}, nil
+}
+
+// Blind generates a token and blinded data.
+func (c *Client) Blind(in []byte) (*Token, BlindToken, error) {
```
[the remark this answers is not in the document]
claucece: Changed ;)

oprf/oprf.go
```diff
@@ -0,0 +1,229 @@
...
+	OPRFP256 uint16 = 0x0003
+	// OPRFP384 is the constant to represent the OPRF P-384 with SHA-512 (SSWU-RO) group.
+	OPRFP384 uint16 = 0x0004
+	// OPRFP521 is the constant to represent the OPRF P-521 with SHA-512 (SSWU-RO) group.
+	OPRFP521 uint16 = 0x0005
+)
+
+var (
+	// OPRFMode is the context string to define a OPRF.
+	OPRFMode *big.Int = big.NewInt(0)
```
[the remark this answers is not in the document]
claucece: Changed now.

oprf/oprf.go
claucece: Mmm.. Client is the struct, which has the method. But this is the interface that defines the methods that can be used by other objects implementing it. But, I'll remove the interface all together ;)

oprf/oprf.go
```diff
@@ -0,0 +1,229 @@
...
+// NewServer creates a new instantiation of a Server.
+func NewServer(suiteID uint16) (*Server, error) {
+	suite, err := suiteFromID(suiteID)
+	if err != nil {
+		return nil, err
+	}
+
+	ctx := generateCtx(suiteID)
+
+	keyPair := generateKeyPair(suite)
```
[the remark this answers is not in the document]
claucece: Added.

oprf/oprf.go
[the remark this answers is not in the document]
claucece: I'll change this on the PR that takes care of the group.

oprf/oprf_test.go
claucece: Let me see how can I add it ;)

oprf/oprf.go
```diff
@@ -0,0 +1,229 @@
...
+	tmp := big.NewInt(int64(byte(suiteID)))
+	id := i2OSP(tmp, 2)
+
+	var ctx []byte
+	ctx = append(ctx, mode...)
+	ctx = append(ctx, id...)
+	return ctx
+}
+
+func generateKeyPair(suite *group.Ciphersuite) *KeyPair {
```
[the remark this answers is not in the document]
claucece: I added the ability to export now ;)

oprf/oprf_test.go
claucece: Using it now ;)

oprf/group/suite.go
claucece: This should be solved now.

oprf/oprf_test.go
claucece: I included VerifyFinalize now, which does a nicer check..

oprf/oprf.go
```diff
@@ -0,0 +1,229 @@
...
+type Server struct {
+	suite *group.Ciphersuite
+	ctx   []byte
+	Keys  *KeyPair
+}
+
+// ServerContext implements the functionality of a Server.
+type ServerContext interface {
+	// Evaluate evaluates the token.
+	Evaluate(b BlindToken) (*Evaluation, error)
```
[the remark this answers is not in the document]
claucece: Included now.

oprf/oprf.go
claucece: Added ;)

oprf/oprf.go
```diff
@@ -0,0 +1,229 @@
...
+type BlindToken []byte
+
+// IssuedToken corresponds to a token that has been issued.
+// Internally, it is a serialized Element.
+type IssuedToken []byte
+
+// Token is the object issuance of the protocol.
+type Token struct {
+	data  []byte
+	blind *group.Scalar
```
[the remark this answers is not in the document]
claucece: We can indeed store the inverse of the blind, rather of the blind itself for `Finalize`. But I liked the flow of `Finalize` being in charge of inverting it. What do you think @armfazh ?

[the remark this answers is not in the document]
armfazh: either way works. Note that calculating inverses is sometimes slow, so it might be good to perform that operation once the server has replied.

oprf/group/group.go
```diff
@@ -0,0 +1,206 @@
...
+	p := &Element{c, new(big.Int), new(big.Int), true}
+	return p
+}
+
+// IsValid checks that the given Element is a valid curve point
+func (p *Element) IsValid() bool {
+	return p.c.curve.IsOnCurve(p.x, p.y)
+}
+
+// ScalarMult multiplies a Element by the provided Scalar value
```
chris-wood, line 29:
```suggestion
// ScalarMult multiplies a Element by the provided Scalar value.
```

oprf/group/group.go
```diff
@@ -0,0 +1,206 @@
...
+	"math/big"
+)
+
+// Element is a representation of a group element. It has two coordinates of
+// *big.Int type.
+type Element struct {
+	c   *Ciphersuite
+	x   *big.Int
+	y   *big.Int
+	com bool // use compression when serialiazing, be default set to true
```
chris-wood, line 15: Where do we ever set this to `false`? Also, shouldn't it be a property of the ciphersuite? For example, [the draft](https://cfrg.github.io/draft-irtf-cfrg-voprf/draft-irtf-cfrg-voprf.html#name-oprfp-384-sha-512 specifies that serialized points for certain ciphersuites are compressed.

oprf/group/group.go
```diff
@@ -0,0 +1,206 @@
...
+	rInv := NewScalar(s.c)
+	rInv.x.Set(inv)
+
+	return rInv
+}
+
+// Serialize the Scalar into a byte slice.
+func (s *Scalar) Serialize() []byte {
+	l := s.c.ByteLength()
+	bytes := s.x.Bytes()
```
chris-wood, line 192: Why can't we just return `s.x.Bytes()`?

oprf/group/suite.go
```diff
@@ -0,0 +1,251 @@
...
+	"math/big"
+
+	"github.com/armfazh/h2c-go-ref"
+)
+
+// Ciphersuite corresponds to the OPRF ciphersuite that is chosen. The
+// Ciphersuite object determines the prime-order group (pog) that is used for
+// performing the OPRF operations, along with the different hash function
+// definitions.
+// Should be created using NewSuite, using an string.
```
chris-wood, line 21:
```suggestion
// Should be created using NewSuite, using a string.
```

oprf/group/suite.go
```diff
@@ -0,0 +1,251 @@
...
+
+	curve elliptic.Curve
+}
+
+// Suite is an interface that defines operations within additive
+// groups of prime order. This is the setting in which the OPRF operations
+// take place.
+type Suite interface {
+	// Returns the identifying name of the group
+	Name() string
```
chris-wood, line 40: Should we also export the Identifier, e.g.:

```
Identifier() uint16
```

oprf/group/suite.go
```diff
@@ -0,0 +1,251 @@
...
+	// hash defines the hash function to be used.
+	hash hash.Hash
+
+	curve elliptic.Curve
+}
+
+// Suite is an interface that defines operations within additive
+// groups of prime order. This is the setting in which the OPRF operations
+// take place.
+type Suite interface {
```
chris-wood, line 38: Why do we call this `Suite` instead of, say, `Group`? Most of the functions on this interface seem to apply to the group. It seems like a `Suite` should generically be an interface similar to something like so:

```
type Group interface {
   // All the group-related functions from below!
}

type Suite interface {
   Name() string
   Identifier() uint16
   Hash() hash.Hash
   Group() Group
}
```

oprf/group/suite.go
```diff
@@ -0,0 +1,251 @@
...
+
+	// Samples a random scalar value from the field of scalars defined by the
+	// group order.
+	RandomScalar() *Scalar
+
+	// Performs a scalar multiplication of the Generator with some scalar
+	// input
+	ScalarMultBase(*Scalar) *Element
+
+	// Returns the ByteLength of objects associated with the Ciphersuite
```
chris-wood, line 64: What is an "object" here? An element? Scalar? Something else?

oprf/group/suite.go
```diff
@@ -0,0 +1,251 @@
...
+	hasher, err := suite.Get(c.dst)
+	if err != nil {
+		return nil, err
+	}
+
+	return eccHasher{c, hasher}, nil
+}
+
+// HashToPoint produces a new point by hashing the input message.
+type HashToPoint interface {
```
chris-wood, line 106:
```suggestion
type HashToElement interface {
```

(To align with the Element terminology)

oprf/group/suite.go
```diff
@@ -0,0 +1,251 @@
...
+	Hash(msg []byte) (*Element, error)
+}
+
+type eccHasher struct {
+	suite  *Ciphersuite
+	hasher h2c.HashToPoint
+}
+
+func (h eccHasher) Hash(in []byte) (*Element, error) {
```
chris-wood, line 115: nit this exported function is missing documentation.

oprf/group/suite.go
```diff
@@ -0,0 +1,251 @@
...
+	return gen.ScalarMult(s)
+}
+
+// ByteLength returns the ByteLength of objects associated with the Ciphersuite.
+func (c *Ciphersuite) ByteLength() int {
+	return (c.curve.Params().BitSize + 7) / 8
+}
+
+// FinalizeHash computes the final hash for the suite
+func FinalizeHash(c *Ciphersuite, data, iToken, info, ctx []byte) []byte {
```
chris-wood, line 197: Why is this function in `suite.go`?

oprf/group/suite.go
```diff
@@ -0,0 +1,251 @@
...
+		cSuite.name = "OPRFP384-SHA512-ELL2-RO"
+		cSuite.dst = append(dst, ctx...)
+		cSuite.hash = sha512.New()
+		cSuite.curve = elliptic.P384()
+	case "P-521":
+		cSuite.name = "OPRFP521-SHA512-ELL2-RO"
+		cSuite.dst = append(dst, ctx...)
+		cSuite.hash = sha512.New()
+		cSuite.curve = elliptic.P521()
+	// TODO: what other libraries should be used?
```
chris-wood, line 245: Let's turn this into a TODO to support ristretto255 and decaf448 suites.

oprf/group/utils.go
```diff
@@ -0,0 +1,44 @@
...
+// and a if c = 0
+func cMov(a, b, c *big.Int) *big.Int {
+	s := new(big.Int).Sub(One, c)
+	m1 := new(big.Int).Mul(c, b)
+	m2 := new(big.Int).Mul(s, a)
+	return new(big.Int).Add(m1, m2)
+}
+
+// Sgn0 returns -1 if x is negative (in little-endian sense) and 1 if x is positive
+func Sgn0(x *big.Int) *big.Int {
```
chris-wood, line 36: Why is this function exported?

oprf/group/suite.go
```diff
@@ -0,0 +1,251 @@
...
+
+	binary.BigEndian.PutUint16(lenBuf, uint16(len(dst)))
+	_, _ = h.Write(lenBuf)
+	_, _ = h.Write(dst)
+
+	return h.Sum(nil)
+}
+
+// NewSuite creates a new ciphersuite for the requested name.
+func NewSuite(name string, ctx []byte) (*Ciphersuite, error) {
```
chris-wood, line 224: Why don't we create suites with `SuiteID` (uint16) values?

oprf/oprf.go
```diff
@@ -0,0 +1,277 @@
...
+}
+
+// Server is a representation of a Server during protocol execution.
+type Server struct {
+	suite *group.Ciphersuite
+	ctx   []byte
+	Kp    *KeyPair
+}
+
+func generateCtx(id SuiteID) []byte {
```
chris-wood, line 85: nit: can we spell out `Context` rather than abbreviate to `Ctx`?

oprf/oprf.go
```diff
@@ -0,0 +1,277 @@
...
+	pub := group.NewElement(suite)
+	err := pub.Deserialize(pubK)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+//GenerateKeyPair generates a KeyPair in accordance with the group.
```
chris-wood, line 113:
```suggestion
// GenerateKeyPair generates a KeyPair in accordance with the group.
```

oprf/oprf.go
```diff
@@ -0,0 +1,277 @@
...
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	return &Server{
+			suite: suite,
+			ctx:   ctx,
+			Kp:    keyPair},
+		nil
```
chris-wood, line 176: Can we run `go fmt` on this? (This formatting seems off.)

oprf/oprf.go
```diff
@@ -0,0 +1,277 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return suite, err
+}
+
+// NewServer creates a new instantiation of a Server. It can create
+// a server with existing keys or use pre-generated keys.
+func NewServer(id SuiteID, privK, pubK []byte) (*Server, error) {
```
chris-wood, line 154: Is it more idiomatic in Go to support optional parameters, or to have separate functions with different parameter lists? For example, rather than make these optional, we could have two functions like so:

```
func NewServer(id SuiteID) (*Server, error) // generate a key pair
func NewServerWithKeyPair(id SuiteID, privK, pubK []byte) (*Server, error) // assign the key pair
```

oprf/oprf_test.go
```diff
@@ -0,0 +1,387 @@
...
+		}
+
+		testOutput, _ := hex.DecodeString(j.Output)
+		if !bytes.Equal(testOutput[:], h[:]) {
+			test.ReportError(t, h[:], testOutput[:], "finalize")
+		}
+	}
+}
+
+func (v *Vectors) runP521(t *testing.T) {
```
chris-wood, line 325: nit: this is unused!

oprf/oprf_test.go
```diff
@@ -0,0 +1,387 @@
...
+	// Test vectors from draft-05
+	var v Vectors
+
+	v.readFile(t, "testdata/256_vectors.json")
+	t.Run("ORPF-P256", v.runP256)
+
+	v.readFile(t, "testdata/384_vectors.json")
+	t.Run("ORPF-P384", v.runP384)
+
+	//v.readFile(t, "testdata/521_vectors.json")
```
chris-wood, line 385: Why is this commented out?

oprf/oprf_test.go
```diff
@@ -0,0 +1,387 @@
...
+			test.ReportError(t, h[:], testOutput[:], "finalize")
+		}
+	}
+}
+
+func TestDraftVectors(t *testing.T) {
+	// Test vectors from draft-05
+	var v Vectors
+
+	v.readFile(t, "testdata/256_vectors.json")
```
chris-wood, line 379: How did we create these test vectors? Why are they separate for each suite? Shouldn't all vectors belong to a single JSON file?

chris-wood (changes requested):
Looking good! I left some suggestions to improve the API, along with minor style nits.

oprf/oprf.go
```diff
@@ -0,0 +1,277 @@
...
+	client.ctx = ctx
+
+	return &Client{
+		suite: suite,
+		ctx:   ctx,
+	}, nil
+}
+
+// Request generates a token and its blinded version.
+func (c *Client) Request(in []byte) (*Token, BlindToken, error) {
```
chris-wood, line 246: Hmm, is it best that the `Request` function output the `Token` and `BlindToken` separately? This seems like an easy way for applications to misuse the API, e.g., by using the blindToken from one Request() call in a different Finalize() call. I think we should encapsulate these in a new type, and then make `Finalize` a function on that type. For example:

```go
type ClientRequest struct {
   input []byte
   token *Token
   blindToken BlindToken
}

func (c *Client) Request(in []byte) (ClientRequest, error) {
   // return a ClientRequest
}

func (r ClientRequest) Finalize(e * Evaluation) ([]byte, error) {
  // compute and return OPRF output
}
```

Also, why does `Finalize` need to return the `IssuedToken`? That doesn't seem like useful output for the protocol.

oprf/group/group.go
claucece: The idea is to move the group functionality eventually outside (in a subsequent PR), so it can be reused for different things besides OPRF.

claucece: To see when it is defined if it is compressed or not, there is the `isCompressed` function.

claucece: and also on every call to `NewElement` is set to true.

oprf/group/suite.go
claucece: Clarified now.

oprf/group/suite.go
claucece: No, it needs to be called like that as the h2c requires that name for the interface.

claucece: Sorry, my bad. I thought this was something else. Changed now.

oprf/group/utils.go
claucece: Unexported now.

oprf/oprf_test.go
claucece: I found a bug on 521 last night. So I have to check those vectors.

oprf/oprf_test.go
claucece: They are taken from the sage poc, with modifications. I thought it will be clearer to have them on separate files.

oprf/oprf.go
claucece: That is the output of `go fmt`.

claucece: I can put the nil on the same line, so it looks more to what you are accustomed to. But, if the second param to return is on the subsequent line, it looks like this ;)

oprf/oprf.go
claucece: Neither optional parameters (in its full PL definition) nor function overloading are supported in Go: https://golang.org/doc/faq#overloading, https://github.com/golang/go/wiki/GoForCPPProgrammers#top 
I could use this: Using Variadic Functions with Option Functions from https://medium.com/@petomalina/dealing-with-optional-parameters-in-go-9780f9bfbd1d

oprf/group/group.go
chris-wood: Sorry, my point is that if the value is never set to `false`, why make it a parameter at all?

oprf/oprf.go
chris-wood: Weird!

oprf/oprf.go
chris-wood: I didn't overload the functions above -- I used separate names. I don't think variadic functions is the right answer here.

oprf/oprf_test.go
chris-wood: Okay, thanks!

oprf/oprf_test.go
chris-wood: What modifications? Ideally, the output of Sage is read directly by the test code here, with *no* modifications, so we reduce the chance of errors.

claucece: Sage outputs the parameters with a `0x`, which golang `hex.DecodeString` cannot parse, as it does not expect the '0x'.

chris-wood: So let's update Sage to drop that prefix! Simple fix. :)

claucece: Sure.

oprf/oprf_test.go
claucece: Fixed now.

oprf/oprf_test.go
claucece: Used now.

oprf/group/group.go
```diff
@@ -0,0 +1,206 @@
...
+
+		p.com = true
+	}
+
+	return nil
+}
+
+// Equal returns a bool indicating whether two Elements are equal.
+func (p *Element) Equal(q *Element) bool {
+	return (p.x.Cmp(q.x) == 0) && (p.y.Cmp(q.y) == 0)
```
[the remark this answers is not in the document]
claucece: Mmm.. the `com` field is irrelevant to compare as it can be set to true or to false, and yet the internal representation of the Element will be equal, as compressed or uncompressed it will be the same Element. The `com` field is only for serialization or deserialization purposes. The Ciphersuite is just for compatibility for OPRF, but we can check it (although in the future for the Group API this will not be included).

oprf/group/group.go
claucece: Because the group functionality will eventually live outside the OPRF to be used be functionality that might need it to be set false.

oprf/group/group.go
```diff
@@ -0,0 +1,206 @@
...
+// *big.Int type.
+type Element struct {
+	c   *Ciphersuite
+	x   *big.Int
+	y   *big.Int
+	com bool // use compression when serialiazing, be default set to true
+}
+
+// NewElement generates a new point.
+func NewElement(c *Ciphersuite) *Element {
```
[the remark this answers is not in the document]
claucece: No, it is not the same. I changed it now.

oprf/group/group.go
```diff
@@ -0,0 +1,206 @@
...
+	return p
+}
+
+// IsValid checks that the given Element is a valid curve point
+func (p *Element) IsValid() bool {
+	return p.c.curve.IsOnCurve(p.x, p.y)
+}
+
+// ScalarMult multiplies a Element by the provided Scalar value
+func (p *Element) ScalarMult(s *Scalar) *Element {
```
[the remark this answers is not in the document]
claucece: That function is exposed as part of the `Suite` interface, as it is not a function of the Element group, but rather of an specific "Element", the generator.

oprf/group/group.go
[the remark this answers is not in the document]
claucece: With the new change, I don't think it is needed to compare the interface `elliptic.curve`.

oprf/group/suite.go
claucece: Done ;)

oprf/group/suite.go
claucece: Mmm.. yes, that looks good. But at some point, I've been thinking of why we need the interfaces themselves. I doesn't seem as anyone else will instantiate them with a different object than `CipherSuite`.

oprf/group/suite.go
claucece: Done ;)

oprf/group/suite.go
claucece: Sure! I looks a little weird, but it works.

oprf/oprf.go
claucece: Sure.

oprf/oprf_test.go
armfazh: I will go with the opposite, make Go skip the 0x prefix on the call to `hex.DecodeString`

oprf/oprf.go
armfazh: I like this idea, also we can extend the `ClientRequest` structure to support a list of tokens per request.
This is more useful when implementing the verifiable mode.

> Also, why does Finalize need to return the IssuedToken? 

I don't think so, unless for some reason the user wants to store them.

oprf/oprf_test.go
claucece: ok.. I'll change them here as well.

oprf/oprf.go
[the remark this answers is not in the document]
claucece: Yeah. I actually do like it more after the Server has replied. It is also nicer for the tests.

oprf/group/suite.go
claucece: I didn't want to expose the internal hash to the outside oprf.

oprf/group/group.go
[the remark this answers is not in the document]
chris-wood: Hmm, it seems odd to make that a function of the suite.

oprf/oprf.go
chris-wood: Yeah, I think the Finalize output should only be the OPRF output, and nothing more.

oprf/oprf.go
claucece: Sure. Using that then.

oprf/group/group.go
claucece: I think this is needed for p521, but I have to check.

oprf/oprf.go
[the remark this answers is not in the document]
chris-wood: Hmm, that's an interesting point. Let's resolve this for now, but maybe consider it later!

oprf/group/group.go
chris-wood: That's fine, but it's *not used today*, and I don't think we should be adding code that's not used or required for a specific purpose.

oprf/group/suite.go
chris-wood: I'm not sure I followed :-\

oprf/group/suite.go
chris-wood: But it is already exported (capital name), right? Or maybe you meant something else. I was basically asking why the function lived in suite.go rather than, say, oprf.go.

oprf/group/suite.go
claucece: I removed the interface all together.

oprf/group/suite.go
claucece: The function is exported, but not the internal hash as part of the Ciphersuite. I put it there, so I didn't have to export the hash field.

claucece: I'll move it  to the outside.

oprf/oprf.go
claucece: Mmm.. Finalize does not ask as a parameter for a the BlindToken, and the BlindToken is of a different type than the Token (one is an struct and the other is a byte array), so there is no way that the BlindToken can be used as a parameter for `Finalize`. 

```
type ClientRequest struct {
   input []byte
   token *Token
   blindToken BlindToken
}
```

Why you need the input if that is already part of the Token struct?

The struct defined here (`ClientRequest`) does not have access to the Ciphersuite (needed to access the type of hash defined by the ciphersuite, needed to do the finalizeHash call), which is accessible in the Client struct. Those parameters will need to be copied form the `Client` struct toe the `ClientRequest` struct. 

I don't really follow the misuse case.

claucece: Should be added now, though.

oprf/group/group.go
claucece: Sure. I'll remove.

oprf/group/group.go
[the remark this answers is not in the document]
claucece: Sure. I'll put as part of Element as wanted.

oprf/oprf.go
chris-wood: We can always change what goes in `ClientRequest` :-) To elaborate on the misuse, consider the following: 

```go
inputA := []byte("inputA")
inputB := []byte("inputB")

tokenA, blindTokenA, errA := c.Request(inputA)
tokenB, blindTokenB, errB := c.Request(inputB)

// ... send off blindTokenA and blindTokenB to get evaluationA and evaluationB

outputA, errA := c.Finalize(tokenA, evaluationB, "info")
outputB, errB := c.Finalize(tokenB, evaluationA, "info") // oops -- I swapped these things around!
```

```go
inputA := []byte("inputA")
inputB := []byte("inputB")

requestA, errA := c.Request(inputA)
requestB, errB := c.Request(inputB)

// ... send off blindTokenA (from requestA) and blindTokenB (from requestB) to get evaluationA and evaluationB

outputA, errA := requestA.Finalize(evaluationA, "info")
outputB, errB := requestB.Finalize(evaluationB, "info")
```

The potential misuse doesn't go away (obviously), but now the caller only needs to match one request to one evaluation, which seems simpler to me. It also lets us batch up multiple requests in a single object, which is nice.

oprf/group/group.go
claucece: Removed as I couldn't find the reference. If later there is some problems, I'm sure they can be reported.

oprf/oprf.go
claucece: Sure. Added that way.

claucece:
cc./ @chris-wood @armfazh you can review again now. Thanks.

oprf/oprf.go
```diff
@@ -0,0 +1,323 @@
...
+	ctx := generateContext(id)
+
+	suite, err := suiteFromID(id, ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	client := &Client{}
+	client.suite = suite
+	client.ctx = ctx
```
armfazh, line 275:
```suggestion

```

oprf/oprf.go
```diff
@@ -0,0 +1,323 @@
...
+	return &ClientRequest{c.suite, c.ctx, tk, bToken}, nil
+}
+
+// Finalize computes the signed token from the server Evaluation and returns
+// the output of the OPRF protocol.
+func (cr *ClientRequest) Finalize(e *Evaluation, info []byte) ([]byte, error) {
+	p := group.NewElement(cr.suite.Curve)
+	err := p.Deserialize(e.element)
+	if err != nil {
+		return []byte{}, err
```
armfazh, line 312:
```suggestion
		return nil, err
```

oprf/oprf.go
```diff
@@ -0,0 +1,323 @@
...
+		suite: suite,
+		ctx:   ctx}, nil
+}
+
+// ClientRequest is a structure to encapsulate the output of a Request call.
+type ClientRequest struct {
+	suite  *group.Ciphersuite
+	ctx    []byte
+	token  *Token
+	bToken BlindToken
```
armfazh, line 287: a user-client cannot send the blindToken because it is not exposed.

armfazh (approved):
few comments.

armfazh:
I messed the CI with a merge commit, sorry :(

@claucece, could you please squash the commits and make ci happy.

chris-wood (approved):

PR #166, opened 2020-10-23
kyber: bugfix: compute A^T when unpacking

bwesterb:
This wasn't hit in the tests, because the tests of unpacking didn't check
whether the cached A^Ts were the same.


claucece:
LGTM.

claucece (approved):

PR #170, opened 2020-10-26
Adds hpke support

armfazh:
Includes an implementation of [hpke](https://www.ietf.org/archive/id/draft-irtf-cfrg-hpke-06.html) v06.


hpke/hpke.go
```diff
@@ -0,0 +1,173 @@
...
+}
+
+type Sender struct {
+	state
+	pkR  crypto.PublicKey
+	info []byte
+	seed []byte
+}
+
+func (s Suite) NewSender(pkR crypto.PublicKey, info, seed []byte) *Sender {
```
cjpatton, line 59: Why use `crypto.PublicKey` instead of `kem.PublicKey"

cjpatton (changes requested):
In my first round I looked at the API changes and additions in the KEM package, beginning with `kem/short`. I have two high-level suggestions:
1. We should be more explicit about which KEMs are specified for use with HPKE. One can imagine alternative KEMs that are based on the same primitives and have similar security properties but aren't wire-compatible.
1. `New()` in `kem/short` exposes a "domain-separation tag" parameter, presumably to enable support for multiple HPKE drafts. (a) IIRC the changes from HPKE-05 to HPKE-06 are not wire compatible, so I'm not sure the code is actually backwards compatible (b) We won't need backwards compatibility with HPKE-05 for ECH. Based on these observations, I'd suggest removing this parameter from the API. When the time comes, we can move from HPKE-06 to RFC with a PR.

kem/short/doc.go
```diff
@@ -0,0 +1,47 @@
...
+// New returns an authenticaed KEM based on a short Weierstrass curve and HKDF
+// as key derivation function. Optionally, a domain-separation tag can be provided.
+func New(id KemID, dst []byte) kem.AuthScheme {
+	var c elliptic.Curve
+	var h crypto.Hash
+	switch id {
+	case KemP256Sha256:
+		c, h = elliptic.P256(), crypto.SHA256
+	case KemP384Sha384:
+		c, h = elliptic.P384(), crypto.SHA384
```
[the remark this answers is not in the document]
armfazh: good catch!

kem/short/doc.go
```diff
@@ -0,0 +1,47 @@
...
+	_ "crypto/sha512" // linking packages
+
+	"github.com/cloudflare/circl/kem"
+)
+
+// KemID is used to identify a KEM instance.
+type KemID = uint16
+
+const (
+	KemP256Sha256 KemID = iota + 0x0010
```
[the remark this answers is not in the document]
armfazh: take a look on the new naming

kem/short/encap.go
```diff
@@ -0,0 +1,101 @@
...
+}
+
+func (s short) encap(pkr kem.PublicKey, pke crypto.PublicKey, ske crypto.PrivateKey) (ct []byte, ss []byte) {
+	dh := make([]byte, s.SharedKeySize())
+	enc, kemCtx := s.coreEncap(dh, pkr, ske, pke)
+	ss = s.extractExpand(dh, kemCtx)
+	return enc, ss
+}
+
+func (s short) authEncap(
```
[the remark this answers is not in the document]
armfazh: no special style guide is enforced other than gofmt

kem/short/doc.go
```diff
@@ -0,0 +1,47 @@
...
+
+var names map[KemID]string
+
+func init() {
+	names = make(map[KemID]string)
+	names[KemP256Sha256] = "KemP256Sha256"
+	names[KemP384Sha384] = "KemP384Sha384"
+	names[KemP521Sha512] = "KemP521Sha512"
+}
+
```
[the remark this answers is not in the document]
armfazh: domain separation tag removed and fixed to `HPKEv06`

hpke/hpke.go
armfazh: We plan to use crypto keys (from `"crypto"`) in the kem too.

armfazh:
@cjpatton, I have made changes based on your previous suggestions. Go forward with the second wave.

kem/shortkem/doc.go
```diff
@@ -0,0 +1,25 @@
...
+	"crypto"
+	"crypto/elliptic"
+
+	"github.com/cloudflare/circl/ecc/p384"
+	"github.com/cloudflare/circl/kem"
+)
+
+// P256HkdfSha256 is a KEM using P256 curve with HKDF based on SHA-256.
+func P256HkdfSha256() kem.AuthScheme {
+	return short{elliptic.P256(), 0x10, "KemP256HkdfSha256", crypto.SHA256}
```
cjpatton, line 14: The codepoint shouldn't be hard-coded here. Instead, I would move the HPKE KEA code points from `circl/hpke` into `circl/kem`. (Here and below.)

cjpatton (changes requested):
Coming along nicely!
1. In general, documentation is missing. Please add comments to interfaces, constructors, and so on.
1. Overall the `hpke` API looks good, but it needs some work.
1. The new KEMs in the `kem` package present themselves as general-purpose KEMs, but are actually specific to HPKE. I think this is likely to get us into trouble in the future. Here's what I suggest we do:
* Merge `kem/shortkem` and `kem/xkem` into one package, call it `kem/hpke`. This is where all HPKE-specific KEMs will live.
* Rename `P256HkdfSha256()` to `Hpke[Dh]P256HkdfSha256()`, `X25519HkdfSha256()` to `Hpke[Dh]X25519HkdfSha256()`, and so on.
* Move the KEM code points defined in `hpke` into `kem/hpke`.

chris-wood (changes requested):
This is really nice work! A full implementation, with not "much" code. What concerns me now is the API. I agree with much of what @cjpatton said, and have some separate questions and suggestions of my own. I think it's nearly ready to go after they are resolved.

kem/xkem/doc.go
```diff
@@ -0,0 +1,20 @@
...
+	"crypto"
+
+	"github.com/cloudflare/circl/dh/x25519"
+	"github.com/cloudflare/circl/dh/x448"
+	"github.com/cloudflare/circl/kem"
+)
+
+// X25519HkdfSha256 is a KEM using X25519 Diffie-Hellman function with HKDF based on SHA-256.
+func X25519HkdfSha256() kem.AuthScheme {
+	return xkem{x25519.Size, 0x20, "KemX25519HkdfSha256", crypto.SHA256}
```
bwesterb, line 14: (For KEM TLS) we need that for two different PublicKeys of the same scheme that `pk1.Scheme() == pk2.Scheme()`.  That's a pointer comparison.  If they were created using different invocations of `X25519HkdfSha256()` this won't hold.  I would suggest:

```go
var x25519HkdfSha256 kem.AuthScheme = &xkem{x25519.Size, 0x20, "KemX25519HkdfSha256", crypto.SHA256}
func X25519HkdfSha256() kem.AuthScheme { return x25519HkdfSha256 }
```

kem/shortkem/key.go
```diff
@@ -0,0 +1,104 @@
...
+}
+func (k *shortPrivKey) Public() shortPubKey {
+	if k.pub == nil {
+		x, y := k.c.ScalarBaseMult(k.k)
+		k.pub = &shortPubKey{k.c, x, y}
+	}
+	return *k.pub
+}
+
+func (s short) UnmarshalBinaryPrivateKey(data []byte) (kem.PrivateKey, error) {
```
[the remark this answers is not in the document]
armfazh: This is go parlance 
https://golang.org/pkg/encoding/#BinaryMarshaler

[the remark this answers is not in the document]
armfazh: but, I will rethink about this

kem/kem.go
```diff
@@ -81,6 +81,14 @@ type Scheme interface {
 	EncapsulationSeedSize() int
 }
 
+// AuthScheme represents a KEM that supports authenticated key encapsulation.
+type AuthScheme interface {
+	Scheme
+	AuthEncapsulate(pkr PublicKey, sks PrivateKey) (ct []byte, ss []byte, err error)
+	AuthDecapsulate(skr PrivateKey, ct []byte, pks PublicKey) ([]byte, error)
+	AuthEncapsulateDeterministically(pk PublicKey, seed []byte, sks PrivateKey) (ct, ss []byte, err error)
```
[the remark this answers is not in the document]
bwesterb: It's needed in KEM TLS.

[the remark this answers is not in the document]
chris-wood: Oh? Can you point to where it's needed there?

hpke/hpke.go
```diff
@@ -0,0 +1,179 @@
...
```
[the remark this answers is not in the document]
armfazh: > * I don't like the idea of the behavior of `Setup` depending on whether `Seed` is provided as input. I would suggest splitting `Setup()` into two functions: (1) `Setup()`, which generates its own coins; and `SetupDeterministic(seed []byte)`, which uses `seed` as the coins. This has the added benefit of aligning with the KEM API.

Splitting in Setup and SetupDeterministic will double from 4 to 8 functions for Sender. It's that what is suggested?

hpke/kem.go
```diff
@@ -0,0 +1,88 @@
...
+)
+
+func (s *Sender) encap(pk crypto.PublicKey) (ct []byte, ss []byte, err error) {
+	pub, ok := pk.(kem.PublicKey)
+	if !ok {
+		return nil, nil, kem.ErrTypeMismatch
+	}
+
+	k := s.getKem()
+	if s.seed == nil {
```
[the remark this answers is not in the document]
armfazh: what is the consensus ?

README.md
```diff
@@ -39,6 +39,8 @@ Version numbers are [Semvers](https://semver.org/). We release a minor version f
 | Key Exchange | FourQ | One of the fastest elliptic curves at 128-bit security level. | Experimental for key agreement and digital signatures. |
 | Key Exchange / Digital signatures | P-384 | Our optimizations reduce the burden when moving from P-256 to P-384. |  ECDSA and ECDH using Suite B at top secret level. |
 | Digital Signatures | Ed25519, Ed448 | RFC-8032 provides new signature schemes based on Edwards curves. | Digital certificates and authentication. |
+| Key Encapsulation | P-256, P-384, P-521, X25519 and X448 | Key encapsulation methods based on Diffie-Hellman. | HPKE |
+| Hybrid Public-Key Encryption | Base, Auth, PSK, AuthPSK | [HPKE](https://datatracker.ietf.org/doc/draft-irtf-cfrg-hpke) is a combination of KEM and Authenticated Encryption. | TLS |
```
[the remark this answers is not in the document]
armfazh: done

hpke/hpke.go
```diff
@@ -0,0 +1,179 @@
...
+// encryption scheme with additional data (AEAD).
+//
+// Specification in https://datatracker.ietf.org/doc/draft-irtf-cfrg-hpke
+package hpke
+
+import "crypto"
+
+const versionLabel = "HPKE-06"
+
+// Exporter is
```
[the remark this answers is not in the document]
armfazh: updated

hpke/hpke.go
```diff
@@ -0,0 +1,179 @@
...
+	Seal(pt, aad []byte) (ct []byte, err error)
+}
+
+// Opener is
+type Opener interface {
+	Exporter
+	Open(ct, aad []byte) (pt []byte, err error)
+}
+
+// ModeID represents an HPKE variant.
```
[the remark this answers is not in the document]
armfazh: unexported variants are also not exposed now

hpke/hpke.go
```diff
@@ -0,0 +1,179 @@
...
+type ModeID = uint8
+
+const (
+	// Base provides hybrid public-key encryption to a public key.
+	Base ModeID = 0x00
+	// PSK provides hybrid public-key encryption with authentication using a pre-shared key.
```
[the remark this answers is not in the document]
armfazh: done in hpke

hpke/algs.go
```diff
@@ -0,0 +1,74 @@
...
+	kdfParams[HkdfSha384] = crypto.SHA384
+	kdfParams[HkdfSha512] = crypto.SHA512
+
+	aeadParams = make(map[AeadID]aeadInfo)
+	aeadParams[AeadAES128GCM] = aeadInfo{aesGCM, 16}
+	aeadParams[AeadAES256GCM] = aeadInfo{aesGCM, 32}
+	aeadParams[AeadCC20P1305] = aeadInfo{chacha20poly1305.New, chacha20poly1305.KeySize}
+}
+
+type aeadInfo struct {
```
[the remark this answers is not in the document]
armfazh: the AEAD interface already provides these methods, so we rely on them to expose those parameters.

hpke/aead.go
```diff
@@ -0,0 +1,68 @@
...
+}
+type sealCxt struct{ *encdecCxt }
+type openCxt struct{ *encdecCxt }
+
+func (c *encdecCxt) Export(expCtx []byte, len uint16) []byte {
+	return c.s.labeledExpand(c.exporterSecret, []byte("sec"), expCtx, len)
+}
+
+func (c *encdecCxt) calcNonce() []byte {
+	out := make([]byte, len(c.seq))
```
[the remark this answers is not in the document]
armfazh: As mentioned above, the nonceSize varies depending on AEAD scheme. and this function is called one per de/encryption

hpke/algs.go
```diff
@@ -0,0 +1,74 @@
...
+
+	"github.com/cloudflare/circl/kem"
+	"github.com/cloudflare/circl/kem/shortkem"
+	"github.com/cloudflare/circl/kem/xkem"
+	"golang.org/x/crypto/chacha20poly1305"
+)
+
+type KemID = uint16
+
+const (
```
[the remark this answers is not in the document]
armfazh: hpke-specific kems moved back to hpke package

kem/shortkem/key.go
```diff
@@ -0,0 +1,104 @@
...
+
+func (s short) UnmarshalBinaryPublicKey(data []byte) (kem.PublicKey, error) {
+	x, y := elliptic.Unmarshal(s, data)
+	if x == nil {
+		return nil, errors.New("invalid public key")
+	}
+	return shortPubKey{s, x, y}, nil
+}
+
+func (s short) DeriveKey(seed []byte) (kem.PublicKey, kem.PrivateKey) {
```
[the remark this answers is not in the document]
armfazh: this requires to change PQ- Kems, what is the proposed name for it?

kem/kem.go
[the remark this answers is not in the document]
armfazh: in HPKE: (modeAuth and modeAuthPSK) + (provided seed) uses `AuthEncapsulateDeterministically`

hpke/kem.go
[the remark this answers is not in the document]
cjpatton: I'd be fine with removing it.

hpke/hpke.go
```diff
@@ -0,0 +1,245 @@
...
+
+import (
+	"errors"
+
+	"github.com/cloudflare/circl/kem"
+)
+
+const versionLabel = "HPKE-06"
+
+// Exporter allows exporting secrets from the encryption Context using a
```
cjpatton, line 18: s/the encryption Context/an HPKE context/

hpke/hpke.go
```diff
@@ -0,0 +1,245 @@
...
+// variable-length PRF. Export takes as input a context string expCtx and a
+// desired length (in bytes), and produces a secret derived from the internal
+// exporter secret using the corresponding KDF Expand function.
+type Exporter interface {
+	Export(expCtx []byte, len uint16) []byte
+}
+
+// Sealer encrypts a plaintext using AEAD encryption. Optionally, this
```
cjpatton, line 26:
```suggestion
// Sealer encrypts a plaintext using an AEAD encryption. Optionally, this
```

kem/kem.go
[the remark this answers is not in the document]
chris-wood: HPKE does not need a deterministic variant of this function.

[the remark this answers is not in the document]
bwesterb: > Oh? Can you point to where it's needed there?

Ok, not precisely.  KEM TLS doesn't use authenticated KEMs.  In the Go TLS implementation test vectors are generated by replacing the random generator by something deterministic. To accommodate this KEM TLS has to use EncapsulateDeterministically.  I can imagine that other applications like to have this as well.

Furthermore, it's not uncommon in schemes to deterministically derive several primitives from a single base key.

hpke/kemBase.go
```diff
@@ -0,0 +1,241 @@
...
+	"golang.org/x/crypto/hkdf"
+)
+
+type dhKEM interface {
+	sizeDH() int
+	calcDH(dh []byte, sk kem.PrivateKey, pk kem.PublicKey) error
+	GenerateKey() (kem.PublicKey, kem.PrivateKey, error)
+	DeriveKey(seed []byte) (kem.PublicKey, kem.PrivateKey)
+	UnmarshalBinaryPrivateKey(data []byte) (kem.PrivateKey, error)
+	UnmarshalBinaryPublicKey(data []byte) (kem.PublicKey, error)
```
cjpatton, line 21: It seems like what we want is for objects of type `dhKEM` is to implement the `kem.Scheme` interface, plus some other DH-specific stuff. Why not replace this with `kem.Scheme`?

hpke/kemBase.go
```diff
@@ -0,0 +1,241 @@
...
+}
+
+type kemBase struct {
+	id   KemID
+	name string
+	crypto.Hash
+	dh dhKEM
```
cjpatton, line 28: `dh dhKEM` -> `dhKEM`? In other words, I think what we want to do is require `kemBase` to implement the `dhKEM` interface.

hpke/algs.go
```diff
@@ -0,0 +1,149 @@
...
+	case HkdfSha256:
+		return crypto.SHA256
+	case HkdfSha384:
+		return crypto.SHA384
+	case HkdfSha512:
+		return crypto.SHA512
+	default:
+		return crypto.Hash(0)
+	}
+}
```
cjpatton, line 71: Would you mind adding the `Extract` and `Expand` operations associated with the KdfID? E.g.,
```
func (k KdfID) Extract(salt, ikm []byte) []byte {
   ...
}

func (k KdfID) Expand(prk, info []byte, outLen uint) []byte {
   ...
}
```

This will be needed for ECH.

cjpatton (changes requested):
Still working through my review, but at this point the API looks good to me. I'll make sure I can integrate it into ECH before approving.

High level comments:
1. I think that all errors should be handled. If an error would truly be the result of programmer error (i.e., independent of the inputs), then it's appropriate to `panic()` I think.
1. The KEMs defined in the hpke package should implement the `kem.Scheme` interface.

Just a heads up that we'll need two additional features for ECH. We can add them here, or in a follow-up PR if you prefer.

**Serialization of the HPKE context.** Currently, our ECH implementation expects to be able to serialize/deserialize the context as follows:
```
enum { sender(0), receiver(1) } HpkeRole; 
struct {                                  
    HpkeRole role;                        
    HpkeKemId kem_id;   // as defined in draft-irtf-cfrg-hpke-05
    HpkeKdfId kdf_id;   // as defined in draft-irtf-cfrg-hpke-05
    HpkeAeadId aead_id; // as defined in draft-irtf-cfrg-hpke-05
    opaque exporter_secret<0..255>;       
    opaque key<0..255>;                   
    opaque base_nonce<0..255>;                 
    uint64 seq;                           
} HpkeContext;   
```
This format isn't standardized, so we can change it if we want to go with something more compact. For example, it would suffice to marshal the role, cipher suite, sequence number, and whatever intermediate value is used to derive the exproter_secret, key, and base_nonce.

**KDF Extract() / Expand() operations.** Not totally essential, but it would be nice to get these API calls for ECH. Alternatively, for HKDF-based KDFs we could `KdfID.Hash()` and pass it to `HKDF` ourselves, but it would be nice not to have to break this abstraction boundary if, in the future, someone wants to use a non-KDF based HKDF for ECH. Unlikely, but it's cheap to future-prove this problem.

kem/shortkem/key.go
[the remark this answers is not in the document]
cjpatton: One reason to leave it as-is is that the same name used by the `kem.Scheme` interface.

kem/shortkem/key.go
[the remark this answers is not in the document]
cjpatton: Something like `DeriveKeyPair`. However, I would be inclined to leave as-is, since this is the name used by the `kem.Scheme` interface.

hpke/kem.go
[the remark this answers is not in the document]
cjpatton: I would say remove the seed until we have a use case for it.

hpke/hpke_test.go
```diff
@@ -0,0 +1,43 @@
+package hpke_test
+
+import (
+	"bytes"
+	"fmt"
+
+	"github.com/cloudflare/circl/hpke"
+)
+
+func Example_hpke() {
```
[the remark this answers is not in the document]
armfazh: using the underscore makes that the function be marked as a package Example.

hpke/kemBase.go
armfazh: there are two clases of kem in this package, both must implement `kem.Scheme` .
the dhKEM interface summarizes what is not common from these two implementations.
`  kem.Scheme = kemBase + dhKem `

hpke/kemBase.go
armfazh: we want kemBase implements what is common between xkem and shortkem.
the dh field in kemBase (the parent) allows to call child functions.

hpke/kemBase.go
```diff
@@ -0,0 +1,241 @@
...
+	)
+	binary.BigEndian.PutUint16(labeledInfo[0:2], l)
+	labeledInfo = append(append(append(append(labeledInfo,
+		versionLabel...),
+		suiteID[:]...),
+		label...),
+		info...)
+	b := make([]byte, l)
+	rd := hkdf.Expand(k.New, prk, labeledInfo)
+	_, _ = io.ReadFull(rd, b)
```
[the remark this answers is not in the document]
armfazh: > -Why not just return rd[:l] here?
rd is a Reader not a slice

hpke/shortkem.go
```diff
@@ -0,0 +1,143 @@
...
+	PK, ok := pk.(*shortPubKey)
+	if !ok {
+		return kem.ErrTypeMismatch
+	}
+	SK, ok := sk.(*shortPrivKey)
+	if !ok {
+		return kem.ErrTypeMismatch
+	}
+	l := len(dh)
+	x, _ := s.ScalarMult(PK.x, PK.y, SK.k)
```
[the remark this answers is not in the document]
armfazh: it's not an error, it's the y-coordinate

hpke/algs.go
armfazh: the implementation of these functions is different from stdlib?

cjpatton: Nope. The ask here is to provide convenient wrappers around these operations. As I mentioned in the top comment, it's not necessarily the case that all KDFs for HPKE will be HKDF based.

hpke/kemBase.go
cjpatton: But unless you explicitly declare it, and object that implements `dhKEM` doesn't actually implement `kem.Scheme`. Put another way, I believe you can't coerce a `dhKEM` implementor into a `kem.Scheme` implementor.

hpke/hpke.go
```diff
@@ -15,16 +15,16 @@ import (
...
 // desired length (in bytes), and produces a secret derived from the internal
 // exporter secret using the corresponding KDF Expand function.
 type Exporter interface {
 	Export(expCtx []byte, len uint16) []byte
 }
 
-// Sealer encrypts a plaintext using AEAD encryption. Optionally, this
-// scheme allows to include additional data.
+// Sealer encrypts a plaintext using an AEAD encryption. The caller supplies the plaintext and
+// associated data; the nonce is stored internally by the sealer and incremented after each call.
```
cjpatton, line 27: nit: Break this comment block at 80 characters

hpke/hpke.go
```diff
@@ -1,4 +1,4 @@
-// Package hpke implements hybrid public/private-key encryption.
+// Package hpke implements the Hybrid Public Key Encryption (HPKE) as specified by draft-irtf-cfrg-hpke-06.
```
cjpatton, line 1: nit: break this comment at 80 characters.

armfazh:
> 1 I think that all errors should be handled

most errors are handled, (let me know if I miss someone)

> 2 The KEMs defined in the hpke package should implement the kem.Scheme interface.

All hpke-kems implement kem.Scheme and are in the kem registry. (see /circl/kem/schemes)

> 1. Remove seed altogether and just use internal coins.

Done, I removed the seed from the external API. However, we still require deterministic methods for 1) internal testing and 2) to satisfy `kem.Scheme` interface.

> KDF Extract() / Expand() operations

Since we only have KDF using hkdf alg, we only expose the internal hash, so users should use hkdf package.

> HPKE context serialization.

not covered in this PR.

cjpatton:
I'm on PTO through Monday. I'll aim to have my next round of comments ready on Tuesday.

hpke/hpke.go
```diff
@@ -0,0 +1,253 @@
+// Package hpke implements the Hybrid Public Key Encryption (HPKE) as specified
```
cjpatton, line 1:
```suggestion
// Package hpke implements the Hybrid Public Key Encryption (HPKE) standard specified
```

hpke/hpke.go
```diff
@@ -0,0 +1,253 @@
...
+// by draft-irtf-cfrg-hpke-06.
+//
+// HPKE works for any combination of an asymmetric-key encapsulation mechanism
```
cjpatton, line 4: nit: s/asymmetric-key/public-key. (Here and elsewhere.)

hpke/hpke.go
```diff
@@ -0,0 +1,253 @@
...
+// (KEM), a key derivation function (KDF), and an authenticated symmetric-key
```
cjpatton, line 5: nit: "symmetric-key" seems redundant in this context.

hpke/hpke.go
```diff
@@ -0,0 +1,253 @@
...
+// encryption scheme with additional data (AEAD).
+//
+// Specification in
+// https://www.ietf.org/archive/id/draft-irtf-cfrg-hpke-06.html
```
cjpatton, line 9: I would link to the datatracker page here. You already mentioned that draft-06 is implemented above.

hpke/hpke.go
```diff
@@ -0,0 +1,253 @@
...
+
+	"github.com/cloudflare/circl/kem"
+)
+
+const versionLabel = "HPKE-06"
+
+// Exporter allows exporting secrets from an HPKE context using a
+// variable-length PRF. Export takes as input a context string expCtx and a
+// desired length (in bytes), and produces a secret derived from the internal
+// exporter secret using the corresponding KDF Expand function.
```
cjpatton, line 23: What are the limits on `len`? Might an implementation panic if `len` is too long? If so, this should be documented here.

hpke/hpke.go
```diff
@@ -0,0 +1,253 @@
...
+type Exporter interface {
+	Export(expCtx []byte, len uint16) []byte
```
cjpatton, line 25: Elsewhere the type of a length parameter is `int` rather than `uint16`. I think we should be consistent about the type of length parameters.

hpke/hpke.go
```diff
@@ -0,0 +1,253 @@
...
+type modeID = uint8
+
+const (
+	// modeBase provides hybrid public-key encryption.
+	modeBase modeID = 0x00
+	// modePSK provides hybrid public-key encryption with authentication using a
+	// pre-shared key.
+	modePSK modeID = 0x01
+	// modeAuth provides hybrid public-key encryption with authentication using
+	// the sender's secret key.
```
cjpatton, line 54:
```suggestion
	// the sender's private key.
```
Elsewhere we use the term "private key". (This might have been my fault.)

hpke/hpke.go
```diff
@@ -0,0 +1,253 @@
...
+	modeAuth modeID = 0x02
+	// modeAuthPSK provides hybrid public-key encryption with authentication
+	// using both a pre-shared key and an asymmetric key.
```
cjpatton, line 57:
```suggestion
	// using both a pre-shared key and the sender's private key.
```

hpke/hpke.go
```diff
@@ -0,0 +1,253 @@
...
+}
+
+// Receiver performs hybrid public-key decryption.
+type Receiver struct {
+	state
+	skR kem.PrivateKey
+	enc []byte
+}
+
+// NewReceiver creates a Receiver with knwoledge of a private-key.
```
cjpatton, line 158:
```suggestion
// NewReceiver creates a Receiver with knwoledge of a private key.
```

hpke/hpke.go
```diff
@@ -0,0 +1,253 @@
...
+}
+
+// NewSender creates a Sender with knowledge of the receiver's public-key.
+func (suite Suite) NewSender(pkR kem.PublicKey, info []byte) (*Sender, error) {
+	if !suite.isValid() {
+		return nil, errors.New("invalid suite")
+	}
+	return &Sender{
+		state: state{Suite: suite, info: info},
+		pkR:   pkR,
```
cjpatton, line 91: As long as we're enforcing the validity of the cipher suite, we should also make sure that `pkR` is the correct type. (Here and below.)

hpke/shortkem.go
```diff
@@ -0,0 +1,147 @@
...
+func (s shortKem) PublicKeySize() int         { return 1 + 2*s.byteSize() }
+func (s shortKem) EncapsulationSeedSize() int { return s.byteSize() }
+
+func (s shortKem) byteSize() int { return (s.Params().BitSize + 7) / 8 }
+
+func (s shortKem) sizeDH() int { return s.byteSize() }
+func (s shortKem) calcDH(dh []byte, sk kem.PrivateKey, pk kem.PublicKey) error {
+	PK, ok := pk.(*shortPubKey)
+	if !ok {
+		return kem.ErrTypeMismatch
```
cjpatton, line 33: It might be better to handle this error earlier, e.g., when the Sender and Receiver are constructed.

hpke/kembase.go
```diff
@@ -0,0 +1,234 @@
...
+	)
+	binary.BigEndian.PutUint16(labeledInfo[0:2], l)
+	labeledInfo = append(append(append(append(labeledInfo,
+		versionLabel...),
+		suiteID[:]...),
+		label...),
+		info...)
+	b := make([]byte, l)
+	rd := hkdf.Expand(k.New, prk, labeledInfo)
+	if n, err := io.ReadFull(rd, b); err != nil || n != int(l) {
```
cjpatton, line 76: I would handle `err != nil` and `n != int(l)` separately. If the second happens, then the `panic()` isn't very useful.

hpke/shortkem.go
```diff
@@ -0,0 +1,147 @@
...
+			uint16(s.byteSize()),
+		)
+		bytes[0] &= bitmask
+		skBig.SetBytes(bytes)
+	}
+	l := s.PrivateKeySize()
+	sk := &shortPrivKey{s, make([]byte, l), nil}
+	copy(sk.k[l-len(bytes):], bytes)
+	return sk.Public(), sk
+}
```
cjpatton, line 78: Add space.

hpke/vectors_test.go
```diff
@@ -0,0 +1,246 @@
...
+	sender, err := s.NewSender(pkR, info)
+	test.CheckNoErr(t, err, h+"err sender")
+
+	recv, err := s.NewReceiver(skR, info)
+	test.CheckNoErr(t, err, h+"err receiver")
+
+	return sender, recv
+}
+
+type SeededKem struct {
```
cjpatton, line 65: I don't think this type should be exported.

hpke/kembase.go
```diff
@@ -0,0 +1,234 @@
...
+	if err != nil {
+		return nil, nil, err
+	}
+	return k.authEncap(pkr, sks, pke, ske)
+}
+
+func (k kemBase) Encapsulate(
+	pkr kem.PublicKey,
+) (ct []byte, ss []byte, err error) {
+	pke, ske, err := k.dh.GenerateKey()
```
cjpatton, line 96: The test vectors validate `EncapsulateDeterminstically()`, but not `Encapsulate()`. To increase test coverage, I would do the following. Instead of calling `GenerateKey()` here, I would generate a seed and pass it to `DeriveKey()`, similar to `EncapsulateDeterministically` below.

kem/schemes/schemes.go
```diff
@@ -1,9 +1,20 @@
 // Package schemes contains a register of KEM schemes.
+//
+// Schemes Implemented
+//
+// Based on standard elliptic curves:
+//  HpkeDHKemP256HkdfSha256, HpkeDHKemP384HkdfSha384, HpkeDHKemP521HkdfSha512
```
cjpatton, line 6: This "Hpke" prefix doesn't match the KEMs exported by `circl/hpke`. Either (1) remove the "Hpke" prefix (here and below) or (2) ename the KEMs in `circl/hpke` to match their names here. My preference would be option (2), since the existing names are somewhat ambiguous in this context.

kem/schemes/schemes_test.go
```diff
@@ -123,3 +124,23 @@ func TestApi(t *testing.T) {
...
 }
+
+func Example_schemes() {
+	// import "github.com/cloudflare/circl/kem/schemes"
+
+	for _, sch := range schemes.All() {
+		fmt.Println(sch.Name())
+	}
+	// Output:
+	// HpkeDHKemP256HkdfSha256
```
cjpatton, line 135: Don't forget to correct these names according to the above change.

cjpatton (changes requested):
This is really close, just a few more small changes. I'm fine with the decision about KDF wrappers and HPKE context serialization. I'll submit a PR for the latter shortly after this one lands.

hpke/algs.go
```diff
@@ -0,0 +1,151 @@
...
+
+type AeadID uint16
+
+const (
+	// AES-128 block cipher in Galois Counter Mode (GCM).
+	AeadAES128GCM AeadID = 0x01
```
cjpatton, line 79: It's odd that abbreviations are capitalized here but not elsewhere. E.g., "HkdfSha256" above and "AES128GCM" here. I suggest changing to "Aes128Gcm" for consistency. Here and below.

hpke/aead.go
```diff
@@ -0,0 +1,68 @@
...
+		return errors.New("seq overflow")
+	}
+	return nil
+}
+
+func (c *sealCxt) Seal(pt, aad []byte) ([]byte, error) {
+	ct := c.AEAD.Seal(nil, c.calcNonce(), pt, aad)
+	err := c.inc()
+	if err != nil {
+		return nil, err
```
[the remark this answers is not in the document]
armfazh:
```python
def Context.IncrementSeq():
     if self.seq >= (1 << (8*Nn)) - 1:
       raise NonceOverflowError
     self.seq += 1

def Context.Seal(aad, pt):
     ct = Seal(self.key, self.ComputeNonce(self.seq), aad, pt)
     self.IncrementSeq()
     return ct
```

Assume Nn = 1, the sequence number ranges from 0 to 255 (a total of 256 encryptions).
However, `IncrementSeq` raise an error if seq=255 preventing the increment to occur.
Hence, during Seal even though the last encryption is performend correctly, the `IncrementSeq` function will trigger exception. Thus, the last allowed encryption is never completed (hence, only 255 encryptions).

Is this the expected behaviour?  @chris-wood @cjpatton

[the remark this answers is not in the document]
cjpatton: Yup, I think that's the correct interpretation of the spec. @chris-wood should chime in to decide if that's the right interpretation.

hpke/kem.go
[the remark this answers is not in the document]
chris-wood: Yes, let's remove the seed absent a use case. I think we should strive to keep the API as minimal and lightweight as possible where possible.

hpke/aead.go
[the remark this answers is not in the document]
chris-wood: Yep, that's correct.

kem/schemes/schemes.go
armfazh:
```go
	dhkemP256HkdfSha256.kemBase.id = DHKemP256HkdfSha256
	dhkemP256HkdfSha256.kemBase.name = "HpkeDHKemP256HkdfSha256"
```

KEM _names_ match, what do not match are Identifiers.

However, using `hpke.HpkeDHKemP256HkdfSha256` makes linter unhappy, it argues that `hpke.DHKemP256HkdfSha256` is clear enough.

hpke/hpke.go
armfazh: added some a check, doc, and test

hpke/hpke.go
armfazh: I think len should be non-negative. `int` is idiomatic in go, though.

hpke/hpke.go
armfazh: changed

hpke/hpke.go
armfazh: great observation, added verification

hpke/kembase.go
armfazh: updated

hpke/shortkem.go
armfazh: added some verification in earlier steps

kem/schemes/schemes.go
cjpatton: Ack, let's keep the linter happy.

cjpatton (approved):
LGTM modulo some comments about error handling. Thanks for your hard work, @armfazh.

hpke/xkem.go
```diff
@@ -0,0 +1,144 @@
...
+		x448.Shared(&ss, &sKey, &pKey)
+		copy(dh, ss[:])
+	}
+	return nil
+}
+func (x xkem) DeriveKey(seed []byte) (kem.PublicKey, kem.PrivateKey) {
+	// Implementation based on
+	// https://www.ietf.org/archive/id/draft-irtf-cfrg-hpke-06.html#name-derivekeypair
+	if len(seed) != x.SeedSize() {
+		panic(kem.ErrSeedSize)
```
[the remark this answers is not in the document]
cjpatton: Eh, on second thought I think this is fine. However, we need to document this behavior.

hpke/aead.go
```diff
@@ -0,0 +1,80 @@
...
+	exporterSecret []byte
+}
+
+type sealCtx struct{ *encdecCtx }
+type openCtx struct{ *encdecCtx }
+
+func (c *encdecCtx) Export(expCtx []byte, length uint) []byte {
+	maxLength := uint(255 * c.KdfID.Hash().Size())
+	if length > maxLength {
+		panic(fmt.Errorf("size greater than %v", maxLength))
```
cjpatton, line 23: This behavior needs to be documented. E.g., "Export panics if length is larger than ...."

armfazh: This is documented in the `hpke.Exporter` interface.

hpke/xkem.go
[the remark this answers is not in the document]
armfazh: yes, it satisy the `kem.Scheme` interface, which announces this panic.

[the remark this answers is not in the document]
cjpatton: It's also worth mentioning here. Imagine you were looking at the go docs for this function -- I don't think it would be immediately apparent that this object implements `kem.Scheme`.

cjpatton (approved):

hpke/shortkem.go
```diff
@@ -0,0 +1,142 @@
...
+	PK := pk.(*shortPubKey)
+	SK := sk.(*shortPrivKey)
+	l := len(dh)
+	x, _ := s.ScalarMult(PK.x, PK.y, SK.k) // only x-coordinate is used.
+	b := x.Bytes()
+	copy(dh[l-len(b):l], b)
+	return nil
+}
+
+func (s shortKem) DeriveKey(seed []byte) (kem.PublicKey, kem.PrivateKey) {
```
chris-wood, line 40: This derives a pair of keys, not a key.

```suggestion
func (s shortKem) DeriveKeyPair(seed []byte) (kem.PublicKey, kem.PrivateKey) {
```

hpke/util.go
```diff
@@ -0,0 +1,119 @@
...
+		make([]byte, Nn),
+		exporterSecret,
+	}, nil
+}
+
+func (st state) verifyPSKInputs(psk, pskID []byte) error {
+	gotPSK := psk != nil
+	gotPSKID := pskID != nil
+	if gotPSK != gotPSKID {
+		return errors.New("inconsistent psk inputs")
```
chris-wood, line 55:
```suggestion
		return errors.New("inconsistent PSK inputs")
```

hpke/util.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+	switch st.modeID {
+	case modeBase | modeAuth:
+		if gotPSK {
+			return errors.New("psk input provided when not needed")
```
chris-wood, line 60:
```suggestion
			return errors.New("PSK input provided when not needed")
```

hpke/util.go
```diff
@@ -0,0 +1,119 @@
...
+		}
+	case modePSK | modeAuthPSK:
+		if !gotPSK {
+			return errors.New("missing required psk input")
```
chris-wood, line 64:
```suggestion
			return errors.New("missing required PSK input")
```

hpke/xkem.go
```diff
@@ -0,0 +1,144 @@
...
+	bytes := x.labeledExpand(
+		dkpPrk,
+		[]byte("sk"),
+		nil,
+		uint16(x.PrivateKeySize()),
+	)
+	copy(sk.k, bytes)
+	return sk.Public(), sk
+}
+func (x xkem) GenerateKey() (kem.PublicKey, kem.PrivateKey, error) {
```
chris-wood, line 66:
```suggestion
func (x xkem) GenerateKeyPair() (kem.PublicKey, kem.PrivateKey, error) {
```

chris-wood (changes requested):
Nearly there! My two main questions are:

1) Are we validating keys correctly?
2) Why are we exporting a deterministic variant of encap? (@bwesterb, can you point to the use case?)

hpke/algs.go
```diff
@@ -0,0 +1,179 @@
...
+		return dhkemX25519HkdfSha256
+	case DHKemX448HkdfSha512:
+		return dhkemX448HkdfSha512
+	default:
+		return nil
+	}
+}
+
+func (k KemID) validatePublicKey(pk kem.PublicKey) bool {
+	var ok = false
```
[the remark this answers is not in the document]
armfazh: it needs early checks, good catch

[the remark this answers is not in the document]
armfazh: updated

hpke/hpke.go
```diff
@@ -0,0 +1,272 @@
...
+
+// Sender performs hybrid public-key encryption.
+type Sender struct {
+	state
+	pkR kem.PublicKey
+}
+
+// NewSender creates a Sender with knowledge of the receiver's public-key.
+func (suite Suite) NewSender(pkR kem.PublicKey, info []byte) (*Sender, error) {
+	if !suite.isValid() {
```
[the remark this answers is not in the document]
armfazh: added a function to always create valid Suites

hpke/algs.go
```diff
@@ -0,0 +1,179 @@
...
+	switch k {
+	case DHKemP256HkdfSha256, DHKemP384HkdfSha384, DHKemP521HkdfSha512:
+		_, ok = pk.(*shortPubKey)
```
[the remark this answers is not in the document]
armfazh: modified

hpke/aead.go
```diff
@@ -0,0 +1,80 @@
...
+	}
+	return c.labeledExpand(c.exporterSecret, []byte("sec"), expCtx, uint16(length))
+}
+
+func (c *encdecCtx) calcNonce() []byte {
+	nonce := (&[12]byte{})[:]
```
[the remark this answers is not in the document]
armfazh: could you please elaborate a bit.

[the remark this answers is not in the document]
chris-wood: Sure. This assumes that the nonce length for this and all future AEADs is 12 bytes. That won't always be the case. I'm suggesting that the nonce length is a function of the AEAD in use. (Does that clarify?)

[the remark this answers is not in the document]
armfazh:
```
	nonce := (&[12]byte{})[:]
	if len := c.NonceSize(); len != 12 {
		nonce = make([]byte, len)
	}
```
We fixed to 12 because it works for current AEADs and doesn't requires dinamic allocation. Also, the next line (if-then) prevents against the future AEADs with different nonce sizes.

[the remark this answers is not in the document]
chris-wood: Sorry, I don't think I'm being clear. I'm saying that `12` should not be a magic number. It *depends on the AEAD in use*. (You can also statically allocate up to 16 bytes on the stack and then use however many are needed based on the AEAD, if you're worried about mallocs minimization.)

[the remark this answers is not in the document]
armfazh: I think I got the idea, take a look on the changes.

hpke/aead.go
```diff
@@ -2,65 +2,74 @@ package hpke
...
+	return c.labeledExpand(c.exporterSecret, []byte("sec"),
+		exporterContext, uint16(length))
 }
 
-func (c *encdecCtx) calcNonce() []byte {
-	nonce := (&[12]byte{})[:]
-	if len := c.NonceSize(); len != 12 {
-		nonce = make([]byte, len)
+func (c *encdecContext) calcNonce() []byte {
+	nonce := (&[16]byte{})[:]
```
chris-wood, line 33: ❤️

hpke/shortkem.go
```diff
@@ -30,7 +30,10 @@ func (s shortKem) calcDH(dh []byte, sk kem.PrivateKey, pk kem.PublicKey) error {
 	PK := pk.(*shortPubKey)
 	SK := sk.(*shortPrivKey)
 	l := len(dh)
-	x, _ := s.ScalarMult(PK.x, PK.y, SK.k) // only x-coordinate is used.
+	x, _ := s.ScalarMult(PK.x, PK.y, SK.priv) // only x-coordinate is used.
+	if x.Sign() == 0 {
```
chris-wood, line 34: ❤️

chris-wood (approved):
This looks great! Thanks for taking the time to go through these suggestions. The end result is really quite nice. :-)

Issue #158, opened 2020-10-12
AVX2 optimisations to Kyber


bwesterb:
- [x] Simple arithmetic
- [x] NTT
- [x] InvNTT
- [ ] Sampling
- [x] NTT-domain multiplication
- [ ] (De)compression of polynomials
- [ ] (Un)packing

bwesterb:
It's fast enough already.

PR #173, opened 2020-11-03
kyber: AVX2-optimized MulHat

bwesterb:
kyber: AVX2-optimized MulHat
    
        BenchmarkMulHat-4           10229090               116 ns/op
        BenchmarkMulHatGeneric-4     1536634               781 ns/op

See #158


pke/kyber/internal/common/asm/src.go
```diff
@@ -24,7 +24,7 @@ func barrettReduceX16(x, q, num, t Op) {
 	//  x - int16((int32(x)*20159)>>26)*q
 
 	VPMULHW(num, x, t)   // t := (int32(x) * 20158) >> 16
-	VPSRAW(U8(10), t, t) // int16(t) >>= 10 so that t = (int32(x) * 20158) >> 26
+	VPSRAW(U8(10), t, t) // int16(t) >⩾ 10 so that t = (int32(x) * 20158) >> 26
```
armfazh, line 27: \>>

armfazh (approved):

pke/kyber/internal/common/asm/src.go
bwesterb: Good catch.

PR #177, opened 2020-11-04
kem: add x448

bwesterb:


armfazh:
This change is already part of https://github.com/cloudflare/circl/pull/170 .

@bwesterb please revert the changes for kem/x25519.
New feature must be discussed before be included.

bwesterb:
Oops, totally missed that.  Will do.

PR #178, opened 2020-11-04
kem: return error instead of panic()ing

bwesterb:
This reverts the revert.

We don't want that third party input (cipher text for instance) is able to cause a panic.

### DIFF
--- a/kem/kyber/kyber512/kyber.go
+++ b/kem/kyber/kyber512/kyber.go
@@ -331,22 +331,22 @@ func (*scheme) DeriveKey(seed []byte) (kem.PublicKey, kem.PrivateKey) {
 	return NewKeyFromSeed(seed[:])
 }
 
-func (*scheme) Encapsulate(pk kem.PublicKey) (ct []byte, ss []byte) {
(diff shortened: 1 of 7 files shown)


kem/kyber/kyber512/kyber.go
```diff
@@ -331,47 +331,47 @@ func (*scheme) DeriveKey(seed []byte) (kem.PublicKey, kem.PrivateKey) {
...
 
 	ct = make([]byte, CiphertextSize)
 	ss = make([]byte, SharedKeySize)
 
 	pub, ok := pk.(*PublicKey)
 	if !ok {
-		panic(kem.ErrTypeMismatch)
+		return nil, nil, kem.ErrTypeMismatch
 	}
 	pub.EncapsulateTo(ct, ss, seed)
```
armfazh, line 359: EncapsulateTo (and other functions) also panics internally.
It seems that returning errors must go deeper, The package pke/kyber also panics.

bwesterb: Those panics can't be triggered when called via the generic API.

armfazh (approved):

PR #179, opened 2020-11-05
kem: add SIKE to uniform API

bwesterb:


armfazh:
I like this, but what about moving `/circl/dh/sidh/sike.go` to `/circl/kem/sike/sike.go` to avoid the boilerplate ?

bwesterb:
> I like this, but what about moving `/circl/dh/sidh/sike.go` to `/circl/kem/sike/sike.go` to avoid the boilerplate ?

It's not that easy: a private/public key needs to know its Scheme and that didn't seem straight forward with the existing types.

armfazh (approved):

PR #181, opened 2020-11-06
Export some missing OPRF pieces. Clean up unit tests.

chris-wood:
cc @armfazh @claucece

### DIFF
--- a/oprf/oprf.go
+++ b/oprf/oprf.go
@@ -1,6 +1,5 @@
-//
 // Package oprf provides an implementation of Oblivious Pseudorandom Functions
-// (OPRFs), as defined on draft-irtf-cfrg-voprf.
+// (OPRFs), as defined on draft-irtf-cfrg-voprf: https://datatracker.ietf.org/doc/draft-irtf-cfrg-voprf/
 // It implements:
 // For a Client:
 //   - Blind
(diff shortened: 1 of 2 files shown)


oprf/oprf_test.go
```diff
@@ -36,23 +36,7 @@ func TestClientSetUp(t *testing.T) {
 	}
 }
 
-func TestClientRequest(t *testing.T) {
```
chris-wood, line 39: This was redundant with the test(s) below.

oprf/oprf_test.go
```diff
@@ -66,97 +50,41 @@ func TestServerEvaluation(t *testing.T) {
...
-	if err != nil {
-		t.Fatal("invalid evaluation of server: " + err.Error())
-	}
-
-	if eval == nil {
-		t.Fatal("invalid evaluation of server: no evaluation.")
-	}
-}
-
-func TestClientFinalize(t *testing.T) {
```
chris-wood, line 84: This was redundant with the test(s) below.

armfazh (approved):

oprf/oprf.go
```diff
@@ -195,8 +193,18 @@ func (s *Server) Evaluate(b BlindToken) (*Evaluation, error) {
 	return &Evaluation{ser}, nil
 }
 
+func mustWrite(h io.Writer, data []byte) {
+	dataLen, err := h.Write(data)
+	if err != nil {
```
claucece, line 198: Write from Hash will never error, why is this checked then?

https://golang.org/pkg/hash/#Hash

```
    // Write (via the embedded io.Writer interface) adds more data to the running hash.
    // It never returns an error.
    io.Writer
```

chris-wood: Defensive programing, in case it does. :)

PR #182, opened 2020-11-06
kyber: more avx2 optimisations

bwesterb:


pke/kyber/internal/common/poly_test.go
```diff
@@ -1,11 +1,18 @@
 package common
 
 import (
-	"crypto/rand"
+	cryptoRand "crypto/rand"
 	"fmt"
+	"math/rand"
```
armfazh, line 6: it would be good if we get ride from math/rand. I think mathRand must be explicitly used too.

```suggestion
	mathRand "math/rand"
```

armfazh (approved):
It would be good if crypto/rand is used instead of math/rand.
So no need to tag `nolint:gosec` in the code.

PR #183, opened 2020-11-07
Refactor the OPRF KeyPair interfaces, and add a serialization test.

chris-wood:
While trying to integrate this API with another library, I found some things that made it somewhat painful to use. I tried to clear those up here. I also added a test to exercise the `KeyPair` serialization logic.

cc @armfazh, @claucece

### DIFF
--- a/oprf/group/group.go
+++ b/oprf/group/group.go
@@ -159,5 +159,5 @@ func (s *Scalar) Serialize() []byte {
 // Deserialize an octet-string into a valid Scalar object.
 func (s *Scalar) Deserialize(in []byte) {
 	byteLength := (s.c.Params().BitSize + 7) / 8
-	s.x = new(big.Int).SetBytes(in[1 : byteLength+1])
+	s.x = new(big.Int).SetBytes(in[:byteLength])
 }
--- a/oprf/oprf.go
+++ b/oprf/oprf.go
@@ -67,8 +67,8 @@ type Evaluation struct {
 
 // KeyPair is an struct containing a public and private key.
 type KeyPair struct {
-	PublicKey  *group.Element
-	PrivateKey *group.Scalar
+	publicKey  *group.Element
+	privateKey *group.Scalar
 }
 
 // Client is a representation of a Client during protocol execution.
@@ -92,7 +92,7 @@ func generateContext(id SuiteID) []byte {
 
 // Serialize serializes a KeyPair elements into byte arrays.
 func (kp *KeyPair) Serialize() []byte {
-	return kp.PrivateKey.Serialize()
+	return kp.privateKey.Serialize()
 }
 
 // Deserialize deserializes a KeyPair into an element and field element of the group.
@@ -101,8 +101,8 @@ func (kp *KeyPair) Deserialize(suite *group.Ciphersuite, encoded []byte) error {
 	privateKey.Deserialize(encoded)
 	publicKey := suite.Generator().ScalarBaseMult(privateKey)
 
-	kp.PublicKey = publicKey
-	kp.PrivateKey = privateKey
+	kp.publicKey = publicKey
+	kp.privateKey = privateKey
 
 	return nil
 }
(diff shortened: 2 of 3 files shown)


oprf/group/group.go
```diff
@@ -159,5 +159,5 @@ func (s *Scalar) Serialize() []byte {
 // Deserialize an octet-string into a valid Scalar object.
 func (s *Scalar) Deserialize(in []byte) {
 	byteLength := (s.c.Params().BitSize + 7) / 8
-	s.x = new(big.Int).SetBytes(in[1 : byteLength+1])
+	s.x = new(big.Int).SetBytes(in[:byteLength])
```
chris-wood, line 162: Perhaps this 1-byte offset was important, somehow, but I couldn't see why. @claucece, do you know?

oprf/oprf.go
```diff
@@ -91,43 +91,28 @@ func generateContext(id SuiteID) []byte {
 }
 
 // Serialize serializes a KeyPair elements into byte arrays.
-func (kp *KeyPair) Serialize() ([]byte, []byte) {
-	pubK := kp.pubK.Serialize()
-	privK := kp.PrivK.Serialize()
-
-	return pubK, privK
+func (kp *KeyPair) Serialize() []byte {
+	return kp.PrivateKey.Serialize()
```
claucece, line 95: Why does it only serialise the Private Key now?

claucece, line 95: Why does it only serialise the Private Key now?

oprf/oprf.go
```diff
@@ -62,13 +62,13 @@ type Token struct {
...
 type Evaluation struct {
-	element []byte
+	Element []byte
 }
 
 // KeyPair is an struct containing a public and private key.
 type KeyPair struct {
-	pubK  *group.Element
-	PrivK *group.Scalar
+	PublicKey  *group.Element
```
claucece, line 70: It was previously argued that the Public Key did not need to be exported. Why is it exported now?

oprf/oprf.go
```diff
@@ -81,7 +81,7 @@ type Client struct {
 type Server struct {
 	suite   *group.Ciphersuite
 	context []byte
-	Kp      *KeyPair
+	Kp      KeyPair
```
claucece, line 84: Why is this not anymore a pointer if the methods will ask for a pointer?

oprf/oprf_test.go
```diff
@@ -12,6 +12,46 @@ import (
...
+	keyPair := GenerateKeyPair(suite)
+
+	server, err := NewServerWithKeyPair(OPRFP256, keyPair)
+	if err != nil {
+		t.Fatal("invalid setup of server: " + err.Error())
+	}
+
+	input := []byte("hello world")
+	info := []byte("info")
+	outputA, errA := server.FullEvaluate(input, info)
```
claucece, line 29: Why is this using a new error variable and not reusing the previous one?

oprf/group/group.go
claucece: Yeah, that is not needed.

claucece (changes requested):

oprf/oprf.go
bwesterb: Reduces allocation because the pointer can point into the instance of Server.

oprf/oprf.go
chris-wood: Because the public key is (already) derived from the private key, so serializing it was unnecessary and probably harmful, especially since the previous deserialization code didn't check to see if the derived public key matched that which was provided as input to `Deserialize`.

oprf/oprf.go
chris-wood: Hmm, was it? I don't remember why. I can unexport it.

oprf/oprf_test.go
chris-wood: I tend to think it's more readable code this way.

chris-wood:
@armfazh, can you please review?

armfazh (approved):

claucece:
Well, it is already merged. But I was going to say that I approve ;)

PR #184, opened 2020-11-10
Adds (V)OPRF and group packages

### LINKED ISSUE #155: Create a unified Group API
cc./ @armfazh

armfazh:
Adds a new pakcage for group operations.
Updates oprf for using group.

Closes #155


group/group.go
```diff
@@ -0,0 +1,74 @@
...
+var (
+	// P256 is the group generated by P-256 elliptic curve.
+	P256 Group = wG{elliptic.P256()}
+	// P384 is the group generated by P-384 elliptic curve.
+	P384 Group = wG{p384.P384()}
+	// P521 is the group generated by P-521 elliptic curve.
+	P521 Group = wG{elliptic.P521()}
+)
+
+// Group si
```
claucece, line 21: We need the full comment here ;)

group/group.go
```diff
@@ -0,0 +1,74 @@
...
+type Group interface {
+	NewElt() Element
```
claucece, line 23: While I like the simplicity of this, it might be better to actually say 'NewElement'. The same for 'NewScalar'.

group/group.go
claucece: Same on some other places.

group/group.go
```diff
@@ -0,0 +1,74 @@
...
+	NewScl() Scalar
+	Identity() Element
+	Generator() Element
+	Order() Scalar
+	Hashes([]byte) Hasher
```
claucece, line 28: Mmm... not really sure this is the best way. 

It is difficult to see here:

```
	h2c := s.suite.Group.Hashes(dst)
	p := h2c.Hash(in)
```

what it is actually happening. Can this be refactored into one call to a `HashToGroup` func?

group/group.go
```diff
@@ -0,0 +1,74 @@
...
+	Random() Randomizer
```
claucece, line 29: I'm not sure why we need this...

oprf/oprf.go
```diff
@@ -67,108 +72,142 @@ type Evaluation struct {
...
+	return
+}
+
 // GenerateKeyPair generates a KeyPair in accordance with the group.
-func GenerateKeyPair(suite *group.Ciphersuite) KeyPair {
-	privateKey := suite.RandomScalar()
-	publicKey := suite.Generator().ScalarBaseMult(privateKey)
+func GenerateKeyPair(s *suite) KeyPair {
+	r := s.Random()
+	privateKey := r.RndScl(rand.Reader)
```
claucece, line 151: I don't really see why we need to call an interface and then the method itself.

oprf/oprf.go
```diff
@@ -67,108 +72,142 @@ type Evaluation struct {
...
 		return nil, err
 	}
+	context := generateContext(id)
 	keyPair := GenerateKeyPair(suite)
 
 	return &Server{
 		suite:   suite,
-		context: context,
-		Kp:      keyPair}, nil
+		context: context[:],
```
claucece, line 191: If the context will not be a variable size, why not have the field be: [3]byte?

oprf/oprf.go
```diff
@@ -67,108 +72,142 @@ type Evaluation struct {
...
 
-func generateContext(id SuiteID) []byte {
-	context := [3]byte{OPRFMode, 0, byte(id)}
+type suite struct {
+	group.Group
+	crypto.Hash
+}
 
-	return context[:]
+func (id SuiteID) String() string {
```
claucece, line 99: This seems to be used only in the tests. Maybe we can move it there ;)

claucece (changes requested):

claucece:
Thank you for this @armfazh ;)

group/group.go
armfazh: Here, I followed the pattern of crypto.Hash, in which a hash function is first created, and then it can be used and reused in many invocations. This specific line, it uses h2c once.

oprf/oprf.go
armfazh: This is because I implemented random using randomBytes+hashing.
There maybe other ways to implement random that be not so expensive.

armfazh: refactoring ideas are welcomed.

armfazh:
This PR now includes group + OPRF + VOPRF + batched.
Test vectors of v06 are passing.

This is ready for review and comments.

group/group.go
```diff
@@ -0,0 +1,75 @@
...
+	Mul(x, y Scalar) Scalar
+	Neg(Scalar) Scalar
+	Inv(Scalar) Scalar
+	encoding.BinaryMarshaler
+	encoding.BinaryUnmarshaler
+}
+
+// Randomizer allows to get random group elements and scalars.
+type Randomizer interface {
+	RndElt(io.Reader) Element
```
chris-wood, line 61: Can we please spell these out? `RandomElement` and `RandomScalar`? (Also, `Randomizer` seems like a misleading name. Could we move these back to the `Group` interface?)

group/group.go
```diff
@@ -0,0 +1,75 @@
...
+	RndScl(io.Reader) Scalar
+}
+
+// Hasher allows to hash byte strings into group elements and scalars.
+type Hasher interface {
+	Encode([]byte) Element
+	Hash([]byte) Element
+	ScHash([]byte) Scalar
```
chris-wood, line 69:
```suggestion
	HashToScalar([]byte) Scalar
```

group/group.go
```diff
@@ -0,0 +1,75 @@
...
+// Randomizer allows to get random group elements and scalars.
+type Randomizer interface {
+	RndElt(io.Reader) Element
+	RndScl(io.Reader) Scalar
+}
+
+// Hasher allows to hash byte strings into group elements and scalars.
+type Hasher interface {
+	Encode([]byte) Element
+	Hash([]byte) Element
```
chris-wood, line 68:
```suggestion
	HashToElement([]byte) Element
```

group/group.go
```diff
@@ -0,0 +1,75 @@
...
+
+// Randomizer allows to get random group elements and scalars.
+type Randomizer interface {
+	RndElt(io.Reader) Element
+	RndScl(io.Reader) Scalar
+}
+
+// Hasher allows to hash byte strings into group elements and scalars.
+type Hasher interface {
+	Encode([]byte) Element
```
chris-wood, line 67: Why is this on the `Hasher` interface?

group/hash.go
```diff
@@ -0,0 +1,89 @@
+package group
+
+import (
+	"crypto"
+	"math/big"
+)
+
+// GFHasher allows to hash byte strings producing one or more field elements.
+type GFHasher interface {
```
chris-wood, line 9: Does this need to be exported? Also, perhaps this should be `FieldHasher`, rather than `GFHasher`?

group/group.go
chris-wood: It's also unclear why this function is called `Hashes` -- what's the rationale?

oprf/key.go
```diff
@@ -0,0 +1,43 @@
+package oprf
```
chris-wood, line 1: Should this file be "key_pair.go"?

oprf/oprf.go
```diff
@@ -1,307 +1,248 @@
...
+type Blind group.Scalar
+type Serialized = []byte
+type Blinded = []byte
 
-// Token is the object issuance of the protocol.
-type Token struct {
-	data  []byte
-	blind *group.Scalar
+type Proof struct {
+	PublicKey Serialized
```
chris-wood, line 64: Why does the proof carry `PublicKey`?

oprf/oprf.go
```diff
@@ -1,307 +1,248 @@
...
-// Internally, it is a serialized Element.
-type BlindToken []byte
+// ErrUnsupportedSuite is thrown when requesting a non-supported suite.
+var ErrUnsupportedSuite = errors.New("non-supported suite")
 
-// IssuedToken corresponds to a token that has been issued.
-// Internally, it is a serialized Element.
-type IssuedToken []byte
+type Blind group.Scalar
+type Serialized = []byte
```
chris-wood, line 60: `SerializedElement`?

oprf/oprf.go
armfazh: This might not be the better place. 
At which point the server's public key must be known by the client.? a) at client's inception  b) only during Finalize step.

oprf/oprf.go
```diff
@@ -139,24 +146,32 @@ func generateContext(id SuiteID) (ctx [3]byte) {
...
+		return nil, err
+	}
+	return suite.generateKeyPair(), nil
 }
 
 func suiteFromID(id SuiteID) (*suite, error) {
 	switch id {
 	case OPRFP256:
-		return &suite{group.P256, crypto.SHA256}, nil
+		return &suite{id, group.P256, crypto.SHA256}, nil
```
claucece: Mmm.. why do we need to pass the id if we already know the group and the hash?

oprf/oprf.go
```diff
@@ -243,7 +228,7 @@ func mustWrite(h io.Writer, data []byte) {
 }
 
 // FinalizeHash computes the final hash for the suite.
-func finalizeHash(s *suite, data, iToken, info, context []byte) []byte {
+func (s *suite) finalizeHash(data, iToken, info []byte) []byte {
```
claucece: Mmm.. why do we make this a method of the suite if it is more an auxiliary func?

claucece (changes requested):

claucece:
Thank you for this @armfazh . I agree with @chris-wood  comments ;)

group/group.go
armfazh: refactored.

group/group.go
armfazh: changed.

group/group.go
armfazh: removed

oprf/oprf.go
armfazh: I opted by Option B.

oprf/oprf.go
armfazh: it's carried to generate context_string later.

oprf/oprf.go
armfazh: either way can work, having it as a method reduces the number of parameters passed to it

armfazh:
I did some light refactoring. Please re-review, and merge if agree the changes.
@claucece @chris-wood

oprf/client.go
```diff
@@ -0,0 +1,156 @@
...
+			return nil, err
+		}
+	}
+	return &ClientRequest{inputs, blinds, blindedElements}, nil
+}
+
+// Finalize computes the signed token from the server Evaluation and returns
+// the output of the OPRF protocol. The function uses server's public key
+// to verify the proof in verifiable mode.
+func (c *Client) Finalize(r *ClientRequest, e *Evaluation, info []byte, pkS *PublicKey) ([][]byte, error) {
```
chris-wood, line 59: So, in non-verifiable mode, `pkS` is ignored? Could we store it in `Client` instead?

oprf/client.go
```diff
@@ -0,0 +1,156 @@
...
+
+// ClientRequest is a structure to encapsulate the output of a Request call.
+type ClientRequest struct {
+	inputs          [][]byte
+	blinds          []Blind
+	BlindedElements []Blinded
+}
+
+// NewClient creates a new instantiation of a Client.
+func NewClient(id SuiteID, m Mode) (*Client, error) {
```
chris-wood, line 19: Should we pass in the public key here, in verifiable mode? (And perhaps default to base mode if the public key is omitted?)

```suggestion
func NewClient(id SuiteID) (*Client, error) {
   // base mode
}

func NewVerifiableClient(id SuiteID, pks *PublicKey) (*Client, error) {
   // verifiable mode
}
```

oprf/server.go
```diff
@@ -0,0 +1,144 @@
...
+	suite, err := suiteFromID(id, m)
+	if err != nil {
+		return nil, err
+	}
+
+	return &Server{*suite, *skS}, nil
+}
+
+// NewServer creates a new instantiation of a Server.
+func NewServer(id SuiteID, m Mode) (*Server, error) {
```
chris-wood, line 32: Likewise for the client, should we split this up into two modes?

```suggestion
func NewServer(id SuiteID) (*Server, error) {
   // base mode
}
func NewVerifiableServer(id SuiteID, skS *PrivateKey) (*Server, error) {
   // verifiable mode
}   
```

chris-wood (commented):
This LGTM (from an API perspective), modulo some comments about how verifiable and base modes are represented. 

@armfazh, do you also need a review of the internals?

oprf/client.go
armfazh: done

oprf/oprf.go
armfazh: after Chris' review, we opted by option B.

oprf/oprf.go
armfazh: this was refactored

armfazh:
@chris-wood all your comments are now addressed.

chris-wood (approved):

chris-wood:
@claucece, is this ready to go?

claucece:
LGTM.

claucece (approved):

PR #190, opened 2020-11-19
Use functions for exposing kem and signature schemes

armfazh:
The registry of algorithms must be read-only, otherwise: 
```go
package main
import (
	"fmt"
	"github.com/cloudflare/circl/kem/kyber/kyber1024"
	"github.com/cloudflare/circl/kem/kyber/kyber512"
)
func main() {
	fmt.Println(kyber1024.Scheme.Name())
	kyber1024.Scheme = kyber512.Scheme
	fmt.Println(kyber1024.Scheme.Name())
}
```
Outputs 
```sh
Kyber1024
Kyber512
```

cc: @bwesterb

### DIFF
--- a/kem/kyber/kyber1024/kyber.go
+++ b/kem/kyber/kyber1024/kyber.go
@@ -263,7 +263,10 @@ func (pk *PublicKey) Unpack(buf []byte) {
 
 type scheme struct{}
 
-var Scheme kem.Scheme = &scheme{}
+var sch kem.Scheme = &scheme{}
+
+// Scheme returns a KEM interface.
+func Scheme() kem.Scheme { return sch }
 
 func (*scheme) Name() string               { return "Kyber1024" }
 func (*scheme) PublicKeySize() int         { return PublicKeySize }
@@ -273,8 +276,8 @@ func (*scheme) SharedKeySize() int         { return SharedKeySize }
 func (*scheme) CiphertextSize() int        { return CiphertextSize }
 func (*scheme) EncapsulationSeedSize() int { return EncapsulationSeedSize }
 
-func (sk *PrivateKey) Scheme() kem.Scheme { return Scheme }
-func (pk *PublicKey) Scheme() kem.Scheme  { return Scheme }
+func (sk *PrivateKey) Scheme() kem.Scheme { return sch }
+func (pk *PublicKey) Scheme() kem.Scheme  { return sch }
 
 func (sk *PrivateKey) MarshalBinary() ([]byte, error) {
 	var ret [PrivateKeySize]byte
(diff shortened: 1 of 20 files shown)


kem/kyber/kyber1024/kyber.go
```diff
@@ -273,8 +273,8 @@ func (*scheme) SharedKeySize() int         { return SharedKeySize }
 func (*scheme) CiphertextSize() int        { return CiphertextSize }
 func (*scheme) EncapsulationSeedSize() int { return EncapsulationSeedSize }
 
-func (sk *PrivateKey) Scheme() kem.Scheme { return Scheme }
-func (pk *PublicKey) Scheme() kem.Scheme  { return Scheme }
+func (sk *PrivateKey) Scheme() kem.Scheme { return Scheme() }
```
bwesterb, line 276: Here the call to Scheme() is not necessary.

kem/schemes/schemes.go
```diff
@@ -13,7 +13,7 @@ import (
 	"github.com/cloudflare/circl/kem/sike/sikep751"
 )
 
-var allSchemes = [...]kem.Scheme{
+var allSchemes = []func() kem.Scheme{
```
bwesterb, line 16: Using func() here is not necessary.

bwesterb (changes requested):
Good point.  With your changes you're being too cautious now.  I only gave a few places.

armfazh:
@bwesterb please take another look

kem/kyber/kyber1024/kyber.go
```diff
@@ -273,8 +273,8 @@ func (*scheme) SharedKeySize() int         { return SharedKeySize }
 func (*scheme) CiphertextSize() int        { return CiphertextSize }
 func (*scheme) EncapsulationSeedSize() int { return EncapsulationSeedSize }
 
-func (sk *PrivateKey) Scheme() kem.Scheme { return Scheme }
-func (pk *PublicKey) Scheme() kem.Scheme  { return Scheme }
+func (sk *PrivateKey) Scheme() kem.Scheme { return &scheme{} }
+func (pk *PublicKey) Scheme() kem.Scheme  { return &scheme{} }
```
bwesterb, line 277: This is still no good as `pk.Scheme() == pk.Scheme()` will fail.

I'd suggest:

```go
var sch kem.Scheme = &scheme{}
func Scheme() kem.Scheme { return sch }
func (sk *PrivateKey) Scheme() kem.Scheme { return sch }
// ...
```

Now the user can't modify the scheme and `pk.Scheme() == pk.Scheme()` still works.

armfazh: Suggestion applied

bwesterb (approved):

PR #191, opened 2020-11-20
kem: add hybrid KEM support

bwesterb:
We will use this to add the hybrid KEM X25519+Kyber512 later for KEM TLS.

A hybrid KEM is constructed simply by concatenating the shared secrets, cipher texts, etc, as suggested by https://datatracker.ietf.org/doc/draft-ietf-tls-hybrid-design/ for TLS and approved by https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-56Cr2.pdf

The disadvantage of this is that if one of the KEMs is weak, that then a part of the shared secret might be weak as well.  So one has to be careful when using the shared secret: it can't be used as the output of a hash.

To derive key pairs deterministically and to encapsulate deterministically, the seeds are derived using SHAKE256 from a seed that has as size the maximum required for the combined schemes.  This is to prevent a non-uniform seed (like such a shared secret could be) to impact just one of the KEMs.

This also adds one useless hybrid KEM: SIKE+Kyber.  This is just to demonstrate the code and we'll remove it once support for X25519 lands. (#170 )


bwesterb:
@armfazh @claucece @thomwiggers

I lost write access to the repository now I'm not part of the Cloudflare GitHub group anymore.  So I can't assign reviewers.  Could one of you review this?

kem/hybrid/hybrid.go
```diff
@@ -0,0 +1,292 @@
...
+
+var ErrUninitialized = errors.New("public or private key not initialized")
+
+// Temporary Hybrid scheme for testing purposes.
+func KyberSike() kem.Scheme {
+	return kyberSike
+}
+
+var kyberSike kem.Scheme = &scheme{
+	"Temporary-Kyber-Sike",
```
thomwiggers, line 37: I assume temporary because `x25519` still needs to be specified as a KEM?

kem/hybrid/hybrid.go
```diff
@@ -0,0 +1,292 @@
...
+}
+
+func (sch *scheme) EncapsulateDeterministically(pk kem.PublicKey, seed []byte) (
+	ct, ss []byte, err error) {
+	if len(seed) != sch.EncapsulationSeedSize() {
+		return nil, nil, kem.ErrSeedSize
+	}
+
+	h := sha3.NewShake256()
+	_, _ = h.Write(seed)
```
thomwiggers, line 218: considering you're overruling the seed _anyway_ by deriving it through `Shake`, shouldn't you just hardcode the seed size (in `EncapsulationSeedSize()`) of the hybrid scheme to 32 for Shake256? Why allow smaller ones?

kem/hybrid/hybrid.go
```diff
@@ -0,0 +1,292 @@
...
+		return nil, kem.ErrCiphertextSize
+	}
+
+	priv, ok := sk.(*privateKey)
+	if !ok {
+		return nil, kem.ErrTypeMismatch
+	}
+
+	firstSize := sch.first.CiphertextSize()
+	ss1, err := sch.first.Decapsulate(priv.first, ct[:firstSize])
```
thomwiggers, line 251: IDK how Go treats list indexing, but could `ct[:firstSize]` or `ct[firstSize:]` do an index-out-of-bounds?

thomwiggers (changes requested):

kem/hybrid/hybrid.go
bwesterb: Yes.

kem/hybrid/hybrid.go
bwesterb: If you combine two schemes which have 128 bit seeds, then you don't really need a 256 bit seed in total.

kem/hybrid/hybrid.go
bwesterb: At the start of the function we check that the size of `ct` is correct.

thomwiggers: Right, I missed that :+1:

claucece:
@bwesterb I added the reviewers. Pinging @chris-wood also here ;)

claucece:
LGTM. @bwesterb could you add a test just to show how the functionality can be used? 

@armfazh do you think we can merge x25519 as part of a different PR than #170.. just to see if we can play with it in KEM TLS?

claucece (changes requested):
Add a test for showing how the functionality works.

bwesterb:
> LGTM. @bwesterb could you add a test just to show how the functionality can be used?

It's already in there: the temporary scheme `hybrid.KyberSike()` demonstrates how to use it and as it's added to the list of schemes it's included in the tests run for every scheme.  (See schemes_test.go.)

claucece:
ah perfect @bwesterb

claucece (approved):

armfazh:
> @bwesterb 
> This is just to demonstrate the code and we'll remove it once support for X25519 lands. (#170 )

> @armfazh do you think we can merge x25519 as part of a different PR than #170.. just to see if we can play with it in KEM TLS?

It happens that the KEMs in hpke (170) are too specfic to the application. So, we need to bring back the previous PR (that I declined) for including kems based on X255/X448. 
Also, these new kems will be specific to the application KEMTLS, so let make sure the documentation states these differences.

kem/hybrid/hybrid.go
```diff
@@ -0,0 +1,292 @@
+// Package hybrid defines several hybrid classical/quantum KEMs.
```
armfazh, line 1: Is it possible to convert this package in a KEM concatenator, rather than listing specific kems inside of it
otherwise, lets rename package `pqhybrid`

kem/schemes/schemes.go
```diff
@@ -20,6 +21,7 @@ var allSchemes = [...]kem.Scheme{
 	sikep434.Scheme(),
 	sikep503.Scheme(),
 	sikep751.Scheme(),
+	hybrid.KyberSike(), // XXX temporary for testing
```
armfazh, line 24: In a side note, I noticed that the documentation doesn't shows what is registered, only if someone opens the source file.

armfazh:
I also added method Public() in the definition of PrivateKey, (you may want to rebase on top of 170)


```
type PrivateKey interface {
	// Returns the scheme for this private key
	Scheme() Scheme

	encoding.BinaryMarshaler
	Equal(PrivateKey) bool
	Public() PublicKey
}
```

kem/hybrid/hybrid.go
bwesterb: I don't think the user should be able to concatenate any KEMs themselves.  This package will list some of them.  I don't see why we have to reflect the fact that they're all PQ in the name.  And why is `pqhybrid` better than `classichybrid`? `classicpqhybrid` is too long.  Just `hybrid` is clear enough I think.

kem/schemes/schemes.go
bwesterb: I don't understand what you mean.

bwesterb:
> I also added method Public() in the definition of PrivateKey, (you may want to rebase on top of 170)
> 
> ```go
> type PrivateKey interface {
> 	// Returns the scheme for this private key
> 	Scheme() Scheme
> 
> 	encoding.BinaryMarshaler
> 	Equal(PrivateKey) bool
> 	Public() PublicKey
> }
> ```

I'd rather make the changes after #170 is merged (and so doesn't change anymore) or have this one merged first.

kem/schemes/schemes.go
armfazh: run godoc command and see the documentation of scheme package.

bwesterb: It's because `allSchemes` is private.  We could add a testable example for `All()` which prints the names of all schemes.  That would show up in the online documentation (but `go doc` doesn't seem to list examples.)

armfazh: I suggest to add a list in the documentation showing schemes. I can include this in the other PR.

go doc  != godoc

bwesterb: Then a testable example which outputs the list would be better because that'll force it to keep in synch, whereas a simple list will get out of date.

bwesterb:
@armfazh Updated to API changes.  Ready to merge.

armfazh (approved):

PR #193, opened 2020-11-25
Add +build ignore tag to ASM generators

cjpatton:


armfazh (approved):

bwesterb:
Now `go generate` does not call the ASM generators anymore.

armfazh:
not even before, I tried to generate files using ` $ go generate -x ./... ` . 
this command generates everything recursively, except it didn't enter in those folders with a go.mod inside. (I don't know why)
could you please reproduce this.

Issue #120, opened 2020-05-27
Develop generic implementation of circl/ecc/p384


armfazh:
The packge `circl/ecc/p384` contains optimized code for AMD64 and ARM64, but lacks of a generic implementation using `math/bits` package.

PR #194, opened 2020-11-25
Allows to compile p384 with arch others than arm64 and amd64.

armfazh:
Fallback to stdlib when compiling circl/ecc/p384 package with archs different from arm64 and amd64.

no functional changes, other than moving files.

Related to #120


claucece (approved):
Thanks!

PR #196, opened 2020-11-25
simd: add two-way f1600 using SHA3 extensions for arm64

bwesterb:
Requires Go 1.16 for the SHA3 instructions and [this cpu feature detection bug.](https://github.com/golang/go/issues/42747)


bwesterb:
@armfazh

armfazh (approved):
nice job!

PR #197, opened 2020-12-03
hpke: Add context serialization

cjpatton:
Adds support for HPKE context serialization as required by https://github.com/cloudflare/go/pull/38. The changes are summarized as follows:
- Renames Exporter to Context.
- Adds Suite() and Marshal() to Context interface.
- Adds unmarshalers for the Sealer and Opener implementations.
- Adds validators for cipher suite codepoints.

### DIFF
--- a/hpke/aead.go
+++ b/hpke/aead.go
@@ -6,11 +6,12 @@ import (
 )
 
 type encdecContext struct {
-	Suite
 	cipher.AEAD
+	suite          Suite
+	exporterSecret []byte
+	key            []byte
 	baseNonce      []byte
 	sequenceNumber []byte
-	exporterSecret []byte
 }
 
 type sealContext struct{ *encdecContext }
@@ -21,12 +22,16 @@ type openContext struct{ *encdecContext }
 // using the corresponding KDF Expand function. It panics if length is greater
 // than 255*N bytes, where N is the size (in bytes) of the KDF's output.
 func (c *encdecContext) Export(exporterContext []byte, length uint) []byte {
-	maxLength := uint(255 * c.KdfID.Hash().Size())
+	maxLength := uint(255 * c.suite.KdfID.Hash().Size())
 	if length > maxLength {
 		panic(fmt.Errorf("size greater than %v", maxLength))
 	}
-	return c.labeledExpand(c.exporterSecret, []byte("sec"),
+	return c.suite.labeledExpand(c.exporterSecret, []byte("sec"),
 		exporterContext, uint16(length))
+}
+
+func (c *encdecContext) Suite() Suite {
+	return c.suite
 }
 
 func (c *encdecContext) calcNonce() []byte {
--- a/hpke/aead_test.go
+++ b/hpke/aead_test.go
@@ -9,7 +9,7 @@ import (
 
 func TestAeadExporter(t *testing.T) {
 	suite := Suite{KdfID: HkdfSha256, AeadID: AeadAes128Gcm}
-	exporter := &encdecContext{Suite: suite}
+	exporter := &encdecContext{suite: suite}
 	maxLength := uint(255 * suite.KdfID.Hash().Size())
 
 	err := test.CheckPanic(func() {
@@ -29,8 +29,8 @@ func TestAeadSeqOverflow(t *testing.T) {
 	Nn := aead.NonceSize()
 	nonce := make([]byte, Nn)
 	_, _ = rand.Read(nonce)
-	sealer := &sealContext{&encdecContext{suite, aead, nonce, make([]byte, Nn), nil}}
-	opener := &openContext{&encdecContext{suite, aead, nonce, make([]byte, Nn), nil}}
+	sealer := &sealContext{&encdecContext{aead, suite, nil, nil, nonce, make([]byte, Nn)}}
+	opener := &openContext{&encdecContext{aead, suite, nil, nil, nonce, make([]byte, Nn)}}
 
 	pt := []byte("plaintext")
 	aad := []byte("aad")
--- a/hpke/algs.go
+++ b/hpke/algs.go
@@ -33,6 +33,19 @@ const (
 	DHKemX448HkdfSha512 KemID = 0x21
 )
 
+func (k KemID) IsValid() bool {
+	switch k {
+	case DHKemP256HkdfSha256,
+		DHKemP384HkdfSha384,
+		DHKemP521HkdfSha512,
+		DHKemX25519HkdfSha256,
+		DHKemX448HkdfSha512:
+		return true
+	default:
+		return false
+	}
+}
+
 func (k KemID) Scheme() kem.AuthScheme {
 	switch k {
 	case DHKemP256HkdfSha256:
@@ -87,6 +100,17 @@ const (
 	HkdfSha512 KdfID = 0x03
 )
 
+func (k KdfID) IsValid() bool {
+	switch k {
+	case HkdfSha256,
+		HkdfSha384,
+		HkdfSha512:
+		return true
+	default:
+		return false
+	}
+}
+
 func (k KdfID) Hash() crypto.Hash {
 	switch k {
 	case HkdfSha256:
@@ -121,6 +145,17 @@ func (a AeadID) New(key []byte) (cipher.AEAD, error) {
 		return chacha20poly1305.New(key)
 	default:
 		panic("invalid AeadID")
+	}
+}
+
+func (a AeadID) IsValid() bool {
+	switch a {
+	case AeadAes128Gcm,
+		AeadAes256Gcm,
+		AeadChaCha20Poly1305:
+		return true
+	default:
+		return false
 	}
 }
 
--- a/hpke/hpke.go
+++ b/hpke/hpke.go
@@ -11,6 +11,7 @@ package hpke
 
 import (
 	"crypto/rand"
+	"encoding"
 	"errors"
 	"io"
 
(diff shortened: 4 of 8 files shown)


hpke/aead.go
```diff
@@ -59,6 +153,46 @@ func (c *sealCtx) Seal(pt, aad []byte) ([]byte, error) {
...
+func (c *sealCtx) Marshal() ([]byte, error) {
+	rawCtx, err := c.encdecCtx.marshal()
+	if err != nil {
+		return nil, err
+	}
+	return append([]byte{0}, rawCtx...), nil
+}
+
+// UnmarshalSealer parses an HPKE sealer.
+func UnmarshalSealer(raw []byte) (Sealer, error) {
```
armfazh, line 185: What about using BinaryMarshaler std interface.

https://golang.org/pkg/encoding/#BinaryUnmarshaler

hpke/aead.go
```diff
@@ -3,21 +3,115 @@ package hpke
...
+	b.AddUint8LengthPrefixed(func(b *cryptobyte.Builder) {
+		b.AddBytes(c.key)
+	})
+	b.AddUint8LengthPrefixed(func(b *cryptobyte.Builder) {
+		b.AddBytes(c.baseNonce)
+	})
+	b.AddUint8LengthPrefixed(func(b *cryptobyte.Builder) {
+		b.AddBytes(c.seq)
+	})
+	return b.Bytes()
```
armfazh, line 57: the raw field must be updated with the result.

hpke/algs.go
```diff
@@ -31,6 +31,19 @@ const (
 	DHKemX448HkdfSha512 KemID = 0x21
 )
 
+func (k KemID) IsValid() bool {
```
armfazh, line 34: These functions are useful to implement `func (suite Suite) isValid() bool`

hpke/aead.go
cjpatton: We could use the `BinaryMarshaler` interface, but we can't use `BinaryUnmarshaler` without additional changes. In particular, using it would require the ability to implement the following pattern:
```
var sealer sealCtx
if err := &sealer.UnmarshalBinary(rawSealer); err != nil {
    // ...
}
```
This doesn't work because `sealCtx` isn't exported. The `Sealer` interface is exported, which is why the `UnmarshalSealer` pattern works.

If you'd like I can at least have `Export` inherit `encoding.BinaryMarshaler`.

hpke/aead.go
cjpatton: Good catch! The fix is a little different though. I removed `raw` from `encdecCtx` and moved it to `sealCtx` and `openCtx`.

hpke/algs.go
cjpatton: Done.

hpke/aead.go
armfazh: That is true, let's keep it as is.

>  I can at least have Export inherit encoding.BinaryMarshaler

You can implement this interface without enforicing it in the Exporter interface.

cjpatton: Up to you, I'm happy to do whatever!

armfazh: Choose what you think is best.

hpke/hpke.go
```diff
@@ -19,27 +19,30 @@ import (
...
+type Context interface {
 	// Export takes a context string expCtx and a desired length (in bytes),
 	// and produces a secret derived from the internal exporter secret using
 	// the corresponding KDF Expand function. It panics if length is greater
 	// than 255*N bytes, where N is the size (in bytes) of the KDF's output.
 	Export(expCtx []byte, length uint) []byte
+	// Suite returns the cipher suite corresponding to this context.
+	Suite() Suite
+	// Marshal returns the binary representation of the context.
+	Marshal() ([]byte, error)
```
armfazh, line 32: I see what you did, so yes you can enforce `binary.MarshalBinary`

hpke/aead.go
cjpatton: Ack, leaving as-is.

cjpatton: For future reference, I like the idea of using BinaryMarshaler and BinaryUnmarshaler interfaces, as the latter saves a malloc in some cases.

hpke/aead.go
```diff
@@ -67,6 +164,47 @@ func (c *sealCtx) Seal(pt, aad []byte) ([]byte, error) {
...
+// struct {
+//   HpkeRole role = 0; // sealer
+//   HpkeContext context;
+// } HpkeSealer;
+func (c *sealCtx) Marshal() ([]byte, error) {
+	rawCtx, err := c.encdecCtx.marshal()
+	if err != nil {
+		return nil, err
+	}
+	c.raw = append([]byte{0}, rawCtx...)
```
armfazh, line 192: Here, you are caching the `raw` value, but not using it in a subsequent call to Marshal.

Also, you can still use the `raw` variable inside the `encdecCtx` object. Right now, you have two `raw` variables.

 However, if you cache at time T, then after doing some encryptions, the serialized value must be updated anyway. So you might not need `raw` variable at all.

armfazh (commented):
Some comments

hpke/aead.go
cjpatton: Removed `raw` from `encdecContext`, `sealContext`, and `openContext`.

I realized caching raw is a bad idea, since Seal() and Open() advance the sequence number.

hpke/hpke.go
cjpatton: Done.

hpke/marshal.go
```diff
@@ -0,0 +1,154 @@
...
+		!s.ReadUint8LengthPrefixed(&t) ||
+		!t.ReadBytes(&c.sequenceNumber, len(t)) {
+		return nil, errors.New("failed to parse context")
+	}
+
+	c.suite.KemID = KemID(kemID)
+	c.suite.KdfID = KdfID(kdfID)
+	c.suite.AeadID = AeadID(aeadID)
+	if !c.suite.isValid() {
+		return nil, errors.New("invalid suite")
```
armfazh, line 58:
```suggestion
		return nil, errHpkeInvalidSuite
```

hpke/marshal.go
```diff
@@ -0,0 +1,154 @@
...
+		err                  error
+		kemID, kdfID, aeadID uint16
+		t                    cryptobyte.String
+	)
+
+	c := new(encdecContext)
+	s := cryptobyte.String(raw)
+	if !s.ReadUint16(&kemID) ||
+		!s.ReadUint16(&kdfID) ||
+		!s.ReadUint16(&aeadID) ||
```
armfazh, line 42:
```suggestion
	if !s.ReadUint16((*uint16)(&c.suite.KemID)) ||
		!s.ReadUint16((*uint16)(&c.suite.KdfID)) ||
		!s.ReadUint16((*uint16)(&c.suite.AeadID)) ||
```

hpke/marshal.go
```diff
@@ -0,0 +1,154 @@
...
+		b.AddBytes(c.sequenceNumber)
+	})
+	return b.Bytes()
+}
+
+// unmarshalContext parses an HPKE context.
+func unmarshalContext(raw []byte) (*encdecContext, error) {
+	var (
+		err                  error
+		kemID, kdfID, aeadID uint16
```
armfazh, line 34:
```suggestion
```

hpke/marshal.go
```diff
@@ -0,0 +1,154 @@
...
+		!s.ReadUint8LengthPrefixed(&t) ||
+		!t.ReadBytes(&c.baseNonce, len(t)) ||
+		!s.ReadUint8LengthPrefixed(&t) ||
+		!t.ReadBytes(&c.sequenceNumber, len(t)) {
+		return nil, errors.New("failed to parse context")
+	}
+
+	c.suite.KemID = KemID(kemID)
+	c.suite.KdfID = KdfID(kdfID)
+	c.suite.AeadID = AeadID(aeadID)
```
armfazh, line 56:
```suggestion
```

hpke/aead_test.go
```diff
@@ -29,8 +29,12 @@ func TestAeadSeqOverflow(t *testing.T) {
 	Nn := aead.NonceSize()
 	nonce := make([]byte, Nn)
 	_, _ = rand.Read(nonce)
-	sealer := &sealContext{&encdecContext{suite, aead, nonce, make([]byte, Nn), nil}}
-	opener := &openContext{&encdecContext{suite, aead, nonce, make([]byte, Nn), nil}}
+	sealer := &sealContext{
```
armfazh, line 32: there is no change here

hpke/aead.go
```diff
@@ -6,29 +6,42 @@ import (
...
 	sequenceNumber []byte
-	exporterSecret []byte
+
+	// operational parameters
+	cipher.AEAD
 }
 
-type sealContext struct{ *encdecContext }
-type openContext struct{ *encdecContext }
+type sealContext struct {
```
armfazh, line 20: there is no change here

hpke/marshal.go
```diff
@@ -0,0 +1,154 @@
...
+	}
+
+	c.AEAD, err = c.suite.AeadID.New(c.key)
+	if err != nil {
+		return nil, err
+	}
+
+	Nn := c.AEAD.NonceSize()
+	if len(c.baseNonce) != Nn {
+		return nil, errors.New("invalid base nonce")
```
armfazh, line 78: invalid nonce length?

armfazh (commented):
Spec says that the encryption context cannot be used for decryption. However, swapping a bit in the raw bytes of the context makes this possible.
if this is an issue, what other defense can be applied?

hpke/aead_test.go
cjpatton: Reverted.

hpke/aead.go
cjpatton: Reverted.

hpke/marshal.go
cjpatton: Changed to "invalid base nonce length". (Made a similar change elsewhere in this function.)

hpke/marshal.go
cjpatton: Done.

hpke/marshal.go
cjpatton: Done.

hpke/marshal.go
cjpatton: Done.

hpke/marshal.go
cjpatton: Done.

cjpatton:
> Spec says that the encryption context cannot be used for decryption. However, swapping a bit in the raw bytes of the context makes this possible.
> if this is an issue, what other defense can be applied?

Yes, misinterpreting a sealer as an opener (or vice versa) is potentially a security issue, but I don't see how we could solve it in general. The API is meant to guide the user towards the right behavior, but there's nothing that stops someone from bypassing the API and misinterpreting the role bit.

armfazh (approved):

hpke/marshal.go
```diff
@@ -0,0 +1,150 @@
...
+	}
+	if len(c.sequenceNumber) != Nn {
+		return nil, errors.New("invalid sequence number length")
+	}
+
+	return c, nil
+}
+
+// MarshalBinary serializes an HPKE sealer according to the format specified
+// below.  (expressed in TLS syntax). Note that this format is not defined by
```
chris-wood, line 84:
```suggestion
// below. (Expressed in TLS syntax). Note that this format is not defined by
```

chris-wood (approved):
This LGTM -- thanks!

cjpatton:
Squashed and nit-picked a bit.

PR #199, opened 2020-12-08
hpke: Add speed tests

cjpatton:
Based on #197.

Compare HPKE public-key encryption to the cost of one KEM operation amortized over many symmetric key encryption operations.


cjpatton:
This was meant as a one-off to provide some useful numbers. No need to actually include this going forward.

PR #200, opened 2020-12-10
hpke: rename algorithm identifiers

armfazh:
Now, one can name algorithms as `KEM.P256.HKDF.SHA256`.

changes:
- identifiers using structures.
- include Parsing functions for get algorithm identifiers.
- rename errors.

### DIFF
--- a/hpke/aead.go
+++ b/hpke/aead.go
@@ -56,7 +56,7 @@ func (c *encdecContext) increment() error {
 		allOnes &= c.sequenceNumber[i]
 	}
 	if allOnes == byte(0xFF) {
-		return errAeadSeqOverflows
+		return errAEADSeqOverflows
 	}
 
 	// performs an increment by 1 and verifies whether the sequence overflows.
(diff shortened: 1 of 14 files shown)


cjpatton:
Is the purpose of this to avoid using camel case for constants? Why not `KEM_P256_HKDF_SHA256` instead?

armfazh:
> Is the purpose of this to avoid using camel case for constants?
> Why not `KEM_P256_HKDF_SHA256` instead?

Yes, the purpose is to have an alternative to `KEM_P256_HKDF_SHA256`. 

https://golang.org/doc/effective_go.html#mixed-caps

cjpatton:
I don't think this is a good idea, as it's not a very common design pattern and it seems like it would be hard to maintain. Better to just have a constant for each codepoint. I prefer the current camel case names, but would be fine with `KEM_P256_HKDF_SHA256`.

armfazh:
> I don't think this is a good idea, as it's not a very common design pattern and it seems like it would be hard to maintain. Better to just have a constant for each codepoint. I prefer the current camel case names, but would be fine with `KEM_P256_HKDF_SHA256`.

Breaking the rule a bit because underscores improve clarity in this case.
see https://github.com/cloudflare/circl/pull/201

armfazh:
closed in favour of #201

PR #201, opened 2020-12-11
hpke: refactor name identifiers

armfazh:


hpke/hpke.go
```diff
@@ -67,20 +67,19 @@ const (
 
 // Suite is an HPKE cipher suite consisting of a KEM, KDF, and AEAD algorithm.
 type Suite struct {
-	KemID  KemID
-	KdfID  KdfID
-	AeadID AeadID
-	_      struct{}
+	kemID  KEM
```
wbl, line 70: By making these private we make it impossible to make suites. Are we sure programs won't ever want to look inside to see which KEM and KDF and AEAD are used?

wbl (approved):

hpke/hpke.go
armfazh: `NewSuite` is the way to construct always valid suites. 

> Are we sure programs won't ever want to look inside to see which KEM and KDF and AEAD are used?

what about a function`(Suite) Params() ` that returns a triple of algorithms.

hpke/hpke.go
```diff
@@ -67,20 +67,19 @@ const (
...
+	kdfID  KDF
+	aeadID AEAD
 }
 
-// NewSuite builds a Suite from a specified set of algorithms. Panics if an
-// algorithm identifier is not valid.
-func NewSuite(kemID KemID, kdfID KdfID, aeadID AeadID) Suite {
-	s := Suite{KemID: kemID, KdfID: kdfID, AeadID: aeadID}
+// NewSuite builds a Suite from a specified set of algorithms. Returns error
```
cjpatton, line 75:
```suggestion
// NewSuite builds a Suite from a specified set of algorithms. Returns an error
```

hpke/hpke.go
```diff
@@ -67,20 +67,19 @@ const (
...
+// if an algorithm identifier is not valid.
+func NewSuite(kemID KEM, kdfID KDF, aeadID AEAD) (*Suite, error) {
```
cjpatton, line 77: I'd rather panic then propagate an error here. The user can check if a suite is valid via `kemID.IsValid() && kdfID.IsValid() && aeadID.isValid()`.

cjpatton (changes requested):
This seems good overall, but it's missing one feature we need for ECH. Specifically, we need access to the `kdfID` field of `Suite`. Please add this, and I'll make sure that integration works before approving the PR.

cjpatton (approved):
I'm happy with this change. Let's merge this, then review a PR for https://github.com/armfazh/circl/pull/8.

chris-wood (approved):

hpke/hpke.go
```diff
@@ -279,9 +278,12 @@ func (r *Receiver) allSetup() (Opener, error) {
...
-	errKemInvalidPrivateKey   = errors.New("invalid kem private key")
-	errKemInvalidSharedSecret = errors.New("invalid shared secret")
-	errAeadSeqOverflows       = errors.New("aead sequence number overflows")
+	errInvalidHPKESuite       = errors.New("hpke: invalid HPKE suite")
+	errInvalidKDF             = errors.New("hpke: invalid KDF identifier")
+	errInvalidKEM             = errors.New("hpke: invalid KEM identifier")
+	errInvalidAEAD            = errors.New("hpke: invalid AEAD identifier")
+	errInvalidKEMPublicKey    = errors.New("hpke: invalid KEM public key")
+	errInvalidKEMPrivateKey   = errors.New("hpke: invalid KEM private key")
+	errInvalidKEMSharedSecret = errors.New("hpke: invalid shared secret")
```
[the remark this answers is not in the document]
armfazh: I'll fix this in then next PR.

PR #202, opened 2020-12-14
hpke: Exposing KDF API

armfazh:


hpke/algs.go
```diff
@@ -44,7 +49,14 @@ func (k KEM) IsValid() bool {
 	}
 }
 
-func (k KEM) Scheme() kem.AuthScheme {
+// AuthScheme returns an instance of a KEM that supports authentication. If
+// the KEM does not support authentication, it returns nil. Panics if the KEM
+// identifier is not valid.
+func (k KEM) AuthScheme() kem.AuthScheme { return k.Scheme().(kem.AuthScheme) }
```
cjpatton, line 55: This seems like option (2.) from https://github.com/armfazh/circl/pull/8#discussion_r542501005.

I still don't think this is a good idea. The complexity it adds to the API isn't needed, since we currently have no HPKE KEMs that aren't suitable for sender authentication. I vote we hold off on this change until we actually need it. At the very least, let's rever this change in the current PR, since it's not relevant to the API changes for KDF. We can discuss it later on.

hpke/algs.go
```diff
@@ -110,14 +122,56 @@ func (k KDF) IsValid() bool {
...
+	case KDF_HKDF_SHA384:
+		return crypto.SHA384.Size()
+	case KDF_HKDF_SHA512:
+		return crypto.SHA512.Size()
+	default:
+		panic(errInvalidKDF)
+	}
+}
+
+// Extract generates a pseudo-random key from a secret input and a salt value.
```
cjpatton, line 140:
```suggestion
// Extract derives a pseudorandom key from a high entropy, secret input and a salt.
```

hpke/algs.go
```diff
@@ -110,14 +122,56 @@ func (k KDF) IsValid() bool {
...
+// The size of the output is determined by KDF.ExtractSize.
+func (k KDF) Extract(secret, salt []byte) (pseudorandomKey []byte) {
+	return hkdf.Extract(k.hash(), secret, salt)
+}
+
+// Expand allows to derive an slice of bytes from a pseudo-random key and
+// domain separation string info. The maximum output length is 255*N bytes,
+// where N is the size (in bytes) of the KDF's output.
```
cjpatton, line 148:
```suggestion
// Expand derives a variable length, pseudorandom string from a pseudorandom
// key and an information string. The maximum output length is 255*N bytes,
// where N is the size (in bytes) of the KDF's output.
```

cjpatton, line 148: Also, document under which conditions this function panics.

hpke/algs.go
```diff
@@ -110,14 +122,56 @@ func (k KDF) IsValid() bool {
...
+func (k KDF) Expand(pseudorandomKey, info []byte, outputLen uint) []byte {
+	extractSize := k.ExtractSize()
+	if len(pseudorandomKey) < extractSize {
+		panic(fmt.Errorf("pseudo-random key must be %v bytes", extractSize))
```
cjpatton, line 152: "pseudo-random" -> "pseudorandom". (Here and elsewhere.)

hpke/hpke_test.go
```diff
@@ -13,26 +13,25 @@ func Example() {
...
 	// HPKE suite is a domain parameter.
-	s := hpke.NewSuite(
-		hpke.KEM_P384_HKDF_SHA384,
-		hpke.KDF_HKDF_SHA384,
-		hpke.AEAD_AES256GCM,
-	)
+	kemID := hpke.KEM_P384_HKDF_SHA384
+	kdfID := hpke.KDF_HKDF_SHA384
+	aeadID := hpke.AEAD_AES256GCM
+
```
cjpatton, line 19: nit: Remove space.

cjpatton (changes requested):
Looks good, but let's revert the change to `func (k KEM) Scheme()` and consider it in a later PR.

hpke/algs.go
armfazh: reverted

hpke/util.go
```diff
@@ -70,7 +66,7 @@ func (st state) verifyPSKInputs(psk, pskID []byte) error {
 	return nil
 }
 
-// Params returns the algorithm identifiers of the suite.
+// Params returns the algorithm identifiers conforming the suite.
```
cjpatton, line 69:
```suggestion
// Params returns the codepoints for the algorithms comprising the suite.
```

hpke/vectors_test.go
```diff
@@ -24,9 +24,14 @@ func TestVectors(t *testing.T) {
...
-
-	dhkem := s.KEMScheme()
+	kemID := KEM(v.KemID)
+	kdfID := KDF(v.KdfID)
+	aeadID := AEAD(v.AeadID)
+	test.CheckOk(kemID.IsValid() && kdfID.IsValid() && aeadID.IsValid(),
+		"bad private key", t)
+
+	s := NewSuite(kemID, kdfID, aeadID)
+	dhkem := s.kemID.Scheme()
```
cjpatton, line 34: How about
```
kemID := KEM(v.KemID)
test.CheckOk(kemID.IsValid(), "unknown KEM algorithm", t)
kem := kemID.Scheme()
```

cjpatton (approved):
LGTM modulo some minor things.

hpke/vectors_test.go
armfazh: good one

PR #207, opened 2020-12-19
Add HPKE Open and Seal phase mismatch tests.

chris-wood:
Addresses the comment [here](https://github.com/cloudflare/circl/pull/203#discussion_r546249281).

### DIFF
--- a/hpke/aead_test.go
+++ b/hpke/aead_test.go
@@ -79,6 +79,56 @@ func TestAeadNonceUpdate(t *testing.T) {
 	}
 }
 
+func TestOpenPhaseMismatch(t *testing.T) {
+	sealer, opener, err := setupAeadTest()
+	test.CheckNoErr(t, err, "setup failed")
+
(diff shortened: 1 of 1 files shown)


cjpatton (commented):
My point in the other PR was that these cases are both covered by `TestNonceUpdate`. In particular, failing to increment the sequence number, by either the sealer or opener, causes that test to fail. Granted, the feedback of the unit test is less specific:
```
FAIL: TestAeadNonceUpdate (0.00s)
    aead_test.go:75: decryption failed: cipher: message authentication failed
```

In any case, I'm fine with adding these explicit test cases.

chris-wood:
> My point in the other PR was that these cases are both covered by TestNonceUpdate.

Yep, I overlooked the fact that you encrypt the *same thing* twice, which does indeed cover this case. Feel free to close.

armfazh (approved):

PR #208, opened 2020-12-25
kem: add hybrids of Kyber and X{25519,448}

bwesterb:


kem/schemes/schemes.go
```diff
@@ -37,7 +37,9 @@ var allSchemes = [...]kem.Scheme{
 	sikep434.Scheme(),
 	sikep503.Scheme(),
 	sikep751.Scheme(),
-	hybrid.KyberSike(), // XXX temporary for testing
```
armfazh, line 40: can we have SikeX too?

armfazh (approved):

kem/schemes/schemes.go
bwesterb: Sure, why not.

PR #215, opened 2021-02-16
Add go 1.16 version.

armfazh:
Adds go1.16 version for testing.

### DIFF
--- a/simd/keccakf1600/f1600x.go
+++ b/simd/keccakf1600/f1600x.go
@@ -62,7 +62,7 @@ func IsEnabledX4() bool { return cpu.X86.HasAVX2 }
 // IsEnabledX2 returns true if the architecture supports a two-way SIMD
 // implementation provided in this package.
 func IsEnabledX2() bool {
-	// After Go 1.16 the flag cpu.ARM64.HasSHA3 is no more exposed.
+	// After Go 1.16 the flag cpu.ARM64.HasSHA3 is no longer exposed.
 	return false
 }
 
--- a/simd/keccakf1600/f1600x.go
+++ b/simd/keccakf1600/f1600x.go
@@ -61,7 +61,10 @@ func IsEnabledX4() bool { return cpu.X86.HasAVX2 }
 
 // IsEnabledX2 returns true if the architecture supports a two-way SIMD
 // implementation provided in this package.
-func IsEnabledX2() bool { return cpu.ARM64.HasSHA3 }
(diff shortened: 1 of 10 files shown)


armfazh:
todo; To investigate clobbered registers in dillithium assembler code.

bwesterb:
It's an issue with avo: https://github.com/mmcloughlin/avo/issues/156

claucece:
I also have it happening in kyber:

```
pke/kyber/internal/common/amd64.s:1448:1: frame pointer is clobbered before saving
```

The `HasSha3` does not exist anymore (in simd/keccakf1600/f1600x.go) in arm64 on go.16 as well:

```
// IsEnabledX2 returns true if the architecture supports a two-way SIMD
// implementation provided in this package.
func IsEnabledX2() bool { return cpu.ARM64.HasSHA3 }
```

We are using the x/sys one, but it might be bumped down there too: https://github.com/golang/go/commit/6b66b59328f3178cceae624a6100cfe8e7944746#diff-52c997d62192731381cdec487ea0eb77390c9ea5941e4d3ac4c9109c272417f6

@bwesterb bwesterb

claucece:
Sort of did something around it here: https://github.com/claucece/go-exp/commit/de5ce64e0205e76cee4372a6e22f9a8a4678a71c

following https://go-review.googlesource.com/c/go/+/248260/2/src/internal/bytealg/index_amd64.s

armfazh:
@cjpatton @claucece please take a look.

claucece:
LGTM ;) Thanks!

claucece (approved):

simd/keccakf1600/f1600x.go
```diff
@@ -61,7 +61,10 @@ func IsEnabledX4() bool { return cpu.X86.HasAVX2 }
 
 // IsEnabledX2 returns true if the architecture supports a two-way SIMD
 // implementation provided in this package.
-func IsEnabledX2() bool { return cpu.ARM64.HasSHA3 }
+func IsEnabledX2() bool {
+	// After Go 1.16 the flag cpu.ARM64.HasSHA3 is no more exposed.
```
cjpatton, line 65:
```suggestion
	// After Go 1.16 the flag cpu.ARM64.HasSHA3 is no longer exposed.
```
```suggestion
	// After Go 1.16 the flag cpu.ARM64.HasSHA3 is no more exposed.
```

cjpatton (approved):
LGTM

Issue #214, opened 2021-02-15
Group Order() function signature oddity


chris-wood:
@bwesterb points out that this prototype is somewhat strange:

```go
type Group interface {
...
	Order() Scalar
...
}
```

Since Scalars are integers between 0 and q-1, the output of this function should be zero! Perhaps this just needs to be a `big.Int`, instead?

@armfazh, what do you think?

armfazh:
About returning big.Int:  I think returning this type motivates users to perform scalar operations using big.Int, which is not desirable, instead we are already providing Scalar arithmetic operations.

I think the `q = Order()` function is merely for informational purposes, and not to perform any operation modulo q.
or maybe returning another type be the option.

chris-wood:
> I think the q = Order() function is merely for informational purposes, and not to perform any operation modulo q.
or maybe returning another type be the option.

I think you may have misunderstood the comment. In the example snippet you wrote -- `q = Order()` -- `q` would be 0. That's not very useful, or informational. If we're going to expose an interface that returns the order of the group, then I think we should actually return the order of the group.

bwesterb:
I do agree with @armfazh that `big.Int` isn't great for cryptographic purposes due to timing attacks.

chris-wood:
> I do agree with @armfazh that big.Int isn't great for cryptographic purposes due to timing attacks.

That's fair. I think a reasonable thing to do here is to just remove this function from the interface altogether. It's not needed.

PR #216, opened 2021-02-16
Adapt existing ristretto225 implementation to the CIRCL Group interfaces()

chris-wood:
This change wraps @bwesterb's ristretto225 implementation. I'm opening it up as a draft PR to start, since there are some questions I'd like to work through regarding the `Order()` function (see #214). Once we agree on the general design, I'll add tests along the lines of [group_test.go](https://github.com/cloudflare/circl/blob/master/group/group_test.go).

cc @claucece, @wbl

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,6 +5,7 @@ go 1.12
 require (
 	github.com/armfazh/h2c-go-ref v0.0.0-20210215173008-07e12a6f8e0d
 	github.com/bwesterb/go-ristretto v1.1.1
+	github.com/stretchr/testify v1.7.0
 	golang.org/x/crypto v0.0.0-20201208171446-5f87f3452ae9
 	golang.org/x/sys v0.0.0-20201211090839-8ad439b19e0f
 )
--- a/go.mod
+++ b/go.mod
@@ -3,6 +3,8 @@ module github.com/cloudflare/circl
 go 1.12
 
 require (
+	github.com/armfazh/h2c-go-ref v0.0.0-20210215173008-07e12a6f8e0d
+	github.com/bwesterb/go-ristretto v1.1.1
 	golang.org/x/crypto v0.0.0-20201208171446-5f87f3452ae9
 	golang.org/x/sys v0.0.0-20201211090839-8ad439b19e0f
 )
--- a/group/ristretto255.go
+++ b/group/ristretto255.go
@@ -16,6 +16,10 @@ var (
 type ristrettoGroup struct {
 }
 
+func (g ristrettoGroup) String() string {
+	return "ristretto225"
+}
+
 type ristrettoElement struct {
 	p r225.Point
 }
--- a/group/ristretto255.go
+++ b/group/ristretto255.go
@@ -1,0 +1,179 @@
+package group
+
+import (
+	"crypto"
+	"io"
+
+	h2c "github.com/armfazh/h2c-go-ref"
+
+	r225 "github.com/bwesterb/go-ristretto"
+)
+
+var (
+	Ristretto255 Group = ristrettoGroup{}
+)
+
+type ristrettoGroup struct {
+}
+
+type ristrettoElement struct {
+	p *r225.Point
+}
+
+type ristrettoScalar struct {
+	s *r225.Scalar
+}
+
+func (g ristrettoGroup) NewElement() Element {
+	return &ristrettoElement{
+		p: &r225.Point{},
+	}
+}
+
+func (g ristrettoGroup) NewScalar() Scalar {
+	return &ristrettoScalar{
+		s: &r225.Scalar{},
+	}
+}
+
+func (g ristrettoGroup) Identity() Element {
+	var zero r225.Point
+	zero.SetZero()
+	return &ristrettoElement{
+		p: &zero,
+	}
+}
+
+func (g ristrettoGroup) Generator() Element {
+	var base r225.Point
+	base.SetBase()
+	return &ristrettoElement{
+		p: &base,
+	}
+}
+
+func (g ristrettoGroup) Order() Scalar {
+	panic("not implemented")
+	return nil
+}
+
+func (g ristrettoGroup) RandomElement(r io.Reader) Element {
+	var x r225.Point
+	x.Rand()
+	return &ristrettoElement{
+		p: &x,
+	}
+}
+
+func (g ristrettoGroup) RandomScalar(r io.Reader) Scalar {
+	var x r225.Scalar
+	x.Rand()
+	return &ristrettoScalar{
+		s: &x,
+	}
+}
+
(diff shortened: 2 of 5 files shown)


group/ristretto255.go
```diff
@@ -0,0 +1,179 @@
...
+var (
+	Ristretto255 Group = ristrettoGroup{}
+)
+
+type ristrettoGroup struct {
+}
+
+type ristrettoElement struct {
+	p *r225.Point
+}
```
bwesterb, line 21: I'm surprised by the choice to have a pointer inside a struct.  Why not

```go
type ristrettoElement struct {
 	p r225.Point
 }
```

or the closely related

```go
type ristrettoElement r255.Point
```

group/ristretto255.go
```diff
@@ -0,0 +1,179 @@
...
+
+type ristrettoScalar struct {
+	s *r225.Scalar
+}
+
+func (g ristrettoGroup) NewElement() Element {
```
bwesterb, line 27: Is there the expectation that the new element should be zero?  Because as it's initialised now, it's in an undefined state.

group/ristretto255.go
```diff
@@ -0,0 +1,179 @@
...
+
+func (g ristrettoGroup) RandomScalar(r io.Reader) Scalar {
+	var x r225.Scalar
+	x.Rand()
+	return &ristrettoScalar{
+		s: &x,
+	}
+}
+
+// Note(caw): this does NOT implement HashToElement as specified in the hash-to-curve draft
```
bwesterb, line 76: I'm happy to make changes to go-ristretto to accommodate the spec.

group/ristretto255.go
```diff
@@ -0,0 +1,179 @@
...
+	return e.p.Equals(x.(*ristrettoElement).p)
+}
+
+func (e *ristrettoElement) Add(x Element, y Element) Element {
+	e.p.Add(x.(*ristrettoElement).p, y.(*ristrettoElement).p)
+	return e
+}
+
+func (e *ristrettoElement) Dbl(x Element) Element {
+	return e.Add(x, x)
```
bwesterb, line 114: I could expose a more efficient double in `go-ristretto`, if desired.

group/ristretto255.go
chris-wood: No particular reason. I changed this to make it the former.

group/ristretto255.go
chris-wood: I think it's a new API. Derive is basically FromUniformBytes, but this is a wrapper around that interface.

group/ristretto255.go
chris-wood: I'm not sure -- @armfazh?

group/ristretto255.go
chris-wood: Yes, please. :-)

chris-wood: https://github.com/bwesterb/go-ristretto/issues/21

chris-wood:
Promoting for full review after adding test vectors. I'll investigate the build failures in a bit.

chris-wood:
Hmm, the "verifying code" stage is failing in CI, due to a new external library:

```
# github.com/stretchr/testify/assert
Error: ../../../go/pkg/mod/github.com/stretchr/testify@v1.7.0/assert/assertions.go:1704:5: undefined: errors.Is
Error: ../../../go/pkg/mod/github.com/stretchr/testify@v1.7.0/assert/assertions.go:1727:6: undefined: errors.Is
Error: ../../../go/pkg/mod/github.com/stretchr/testify@v1.7.0/assert/assertions.go:1750:5: undefined: errors.As
Error: ../../../go/pkg/mod/github.com/stretchr/testify@v1.7.0/assert/assertions.go:1767:7: undefined: errors.Unwrap
Error: ../../../go/pkg/mod/github.com/stretchr/testify@v1.7.0/assert/assertions.go:1771:7: undefined: errors.Unwrap
```

@armfazh, what does this stage do?

group/ristretto255.go
armfazh: Yes, it is expected New calls the Identity method.
e.g.,
```func (g wG) NewElement() Element { return g.Identity() }```

File: go.mod [hunk 5]
```diff
@@ -3,6 +3,9 @@ module github.com/cloudflare/circl
 go 1.12
 
 require (
+	github.com/armfazh/h2c-go-ref v0.0.0-20210215173008-07e12a6f8e0d
```
armfazh, line 6: It is better to use the tagged version of h2c.
`go get github.com/armfazh/h2c-go-ref@v10.0.1`

armfazh: versions from v2 and above require a [special path](https://golang.org/ref/mod#minimal-module-compatibility) that matches the version number. 
I confirm that h2c-go-ref points to v10.0.1.

armfazh (commented):
also add ristretto to the batch of tests.

```go
func TestGroup(t *testing.T) {
	for _, g := range []group.Group{
		group.P256,
		group.P384,
		group.P521,
		group.Ristretto255,
```

chris-wood:
@armfazh done -- good catch! I had to fix the identity serialization test. The change makes me wonder if Marshal() should ever return an encoded element whose byte length is less than the expected size. (For example, 32 zeroes instead of 1 zero.) I can see arguments for either or, but here I went with the former.

armfazh:
> @armfazh done -- good catch! I had to fix the identity serialization test. The change makes me wonder if Marshal() should ever return an encoded element whose byte length is less than the expected size. (For example, 32 zeroes instead of 1 zero.) I can see arguments for either or, but here I went with the former.

I think it is ok to check for all-zeros, but how many of them? 

There should be some field in Group that tells you the encoding size. This works well with groups that have fixed-size encodings, but for NIST groups we have variable length encoding, say 1 byte, n+1 bytes, or 2n+1 bytes.
-- btw, a similar discussion in the hpke list.

chris-wood:
> I think it is ok to check for all-zeros, but how many of them?

My take: the number of zeroes should match the size of an encoded element. (That means the existing group implementations should change, since the identity encoding is `[]byte{0x00}`.) 

Can we address this in a future issue?

armfazh:
> My take: the number of zeroes should match the size of an encoded element. (That means the existing group implementations should change, since the identity encoding is `[]byte{0x00}`.)

1-byte encoding for identity follows the SEC specification.

> Can we address this in a future issue?
sure, this is not a blocker for this pr.

group/ristretto255.go
bwesterb: Done.

group/ristretto255.go
bwesterb: Could you be a bit more specific what functionality you need?

claucece:
LGTM.

chris-wood:
@armfazh what additional changes would you like to see?

armfazh (approved):

PR #219, opened 2021-03-02
Updating test vectors to draft v06.

armfazh:
Updating test vectors to draft v06.


lukevalenta (approved):
Confirmed with @armfazh that the reason these test vectors do not match https://www.ietf.org/archive/id/draft-irtf-cfrg-voprf-06.html#name-test-vectors exactly is that the seed was added after the publication of the draft, and these test vectors will match for future drafts. The `Input`, `Blind`, and `BlindedElements` fields match the current draft, but just not elements dependent on the seed.

chris-wood (approved):

PR #222, opened 2021-03-15
Update HPKE to v08.

chris-wood:
This only changes the version tag and test vectors.

### DIFF
--- a/hpke/vectors_test.go
+++ b/hpke/vectors_test.go
@@ -15,8 +15,8 @@ import (
 
 func TestVectors(t *testing.T) {
 	// Test vectors from
-	// https://github.com/cfrg/draft-irtf-cfrg-hpke/blob/draft-irtf-cfrg-hpke-07/test-vectors.json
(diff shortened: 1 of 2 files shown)


armfazh (approved):

hpke/vectors_test.go
```diff
@@ -15,8 +15,8 @@ import (
 
 func TestVectors(t *testing.T) {
 	// Test vectors from
-	// https://github.com/cfrg/draft-irtf-cfrg-hpke/blob/draft-irtf-cfrg-hpke-07/test-vectors.json
-	vectors := readFile(t, "testdata/vectors_v07.json")
+	// https://github.com/cfrg/draft-irtf-cfrg-hpke/blob/master/test-vectors.json
```
Lekensteyn, line 18: As this is a volatile resource, I would suggest using a tag as reference:
https://github.com/cfrg/draft-irtf-cfrg-hpke/blob/draft-irtf-cfrg-hpke-08/test-vectors.json

PR #226, opened 2021-05-04
oprf: Update deterministic key generation.

### LINKED ISSUE #225: oprf: Update deterministic key derivation
https://github.com/cfrg/draft-irtf-cfrg-voprf/pull/247

armfazh:
Fixes #225


oprf/keys.go
```diff
@@ -44,19 +47,24 @@ func (k *PrivateKey) Public() *PublicKey {
...
 
 // GenerateKey generates a pair of keys in accordance with the suite.
-func GenerateKey(id SuiteID) (*PrivateKey, error) {
+func GenerateKey(id SuiteID, rnd io.Reader) (*PrivateKey, error) {
 	suite, err := suiteFromID(id, BaseMode)
 	if err != nil {
 		return nil, err
 	}
-	return suite.generateKey(), nil
+	if rnd == nil {
```
wbl, line 55: Why can't the caller pass in rand.Reader themselves?

wbl (approved):

oprf/keys.go
armfazh: now, it panics. please review changes.

wbl: I think you might not even need the if and let the read function panic.

armfazh: Here we are following the api pattern in go stdlib.
e.g. https://golang.org/pkg/crypto/ecdsa/#GenerateKey

PR #227, opened 2021-05-05
Adding XOF package.

armfazh:


xof/xof.go
```diff
@@ -0,0 +1,86 @@
...
+	io.Writer
+
+	// Read reads more output from the hash. It returns io.EOF if the limit
+	// has been reached.
+	io.Reader
+
+	// Clone returns a copy of the XOF in its current state.
+	Clone() XOF
+
+	// Reset resets the XOF to its initial state.
```
wbl, line 26: While this is copied from the blake2b definition, I had some trouble understanding how this works for keyed ones. Maybe say "discards all data appended by Write" to clarify.

xof/xof.go
```diff
@@ -0,0 +1,86 @@
...
+var xofRegistry = make([]func() XOF, maxID)
+
+func init() {
+	RegisterXOF(SHAKE128, newShake128)
+	RegisterXOF(SHAKE256, newShake256)
+	RegisterXOF(BLAKE2XB, newBlake2xb)
+	RegisterXOF(BLAKE2XS, newBlake2xs)
+}
+
+const maxID = 4
```
wbl, line 60: I would add this to the end of the constant definitions.

xof/xof.go
```diff
@@ -0,0 +1,86 @@
...
+func RegisterXOF(x ID, f func() XOF) {
+	if x >= maxID {
+		panic("crypto: RegisterXOF of unknown XOF function")
+	}
+	xofRegistry[x] = f
+}
+
+var xofRegistry = make([]func() XOF, maxID)
+
+func init() {
```
wbl, line 53: This makes this package which just defines an interface import these other packages, which makes the whole registration business a bit less useful. Not that the registration was very useful in the first place imho.

wbl (changes requested):

xof/xof.go
armfazh: agree, registry not needed at all, changed now.

xof/xof.go
armfazh: updated now

claucece (approved):

wbl (approved):

PR #228, opened 2021-05-05
Removes h2c-go-ref dependency

armfazh:
Removes h2c-go-ref dependency


group/short.go
```diff
@@ -226,38 +233,34 @@ func (s *wScl) UnmarshalBinary(b []byte) error {
...
-	scl FieldHasher
-}
-
 func (g wG) HashToElement(b, dst []byte) Element {
 	var u [2]big.Int
-	h := g.getHasher(dst)
-	h.elt.HashToField(u[:], b)
-	Q0 := h.Map(&u[0])
-	Q1 := h.Map(&u[1])
+	g.precomputeHashParams()
```
wbl, line 238: Should I be concerned about concurrent access to one of these?

claucece (approved):

group/short.go
armfazh: yes, this is not thread-safe. I fixed that now.

PR #229, opened 2021-06-07
group, oprf: Exposes bytesize of group elements and scalars

armfazh:
Exposes the bytesize of group elements and scalars.

### DIFF
--- a/group/group.go
+++ b/group/group.go
@@ -19,8 +19,15 @@ var (
 	P521 Group = wG{elliptic.P521()}
 )
 
+type GroupParams struct {
+	ElementLength           uint // Length in bytes of an element.
+	CompressedElementLength uint // Length in bytes of a compressed element.
+	ScalarLength            uint // Length in bytes of a scalar.
+}
+
 // Group represents a prime-order group based on elliptic curves.
 type Group interface {
+	Params() *GroupParams // Params returns parameters for the group
 	NewElement() Element
 	NewScalar() Scalar
 	Identity() Element
--- a/group/group.go
+++ b/group/group.go
@@ -19,7 +19,7 @@ var (
 	P521 Group = wG{elliptic.P521()}
 )
 
-type GroupParams struct {
+type Params struct {
 	ElementLength           uint // Length in bytes of an element.
 	CompressedElementLength uint // Length in bytes of a compressed element.
 	ScalarLength            uint // Length in bytes of a scalar.
@@ -27,7 +27,7 @@ type GroupParams struct {
 
 // Group represents a prime-order group based on elliptic curves.
 type Group interface {
-	Params() *GroupParams // Params returns parameters for the group
+	Params() *Params // Params returns parameters for the group
 	NewElement() Element
 	NewScalar() Scalar
 	Identity() Element
--- a/group/group_test.go
+++ b/group/group_test.go
@@ -124,14 +124,21 @@ func isZero(b []byte) bool {
 }
 
 func testMarshal(t *testing.T, testTimes int, g group.Group) {
+	params := g.Params()
 	I := g.Identity()
 	got, _ := I.MarshalBinary()
 	if !isZero(got) {
 		test.ReportError(t, got, "Non-zero identity")
 	}
+	if l := uint(len(got)); !(l == 1 || l == params.ElementLength) {
+		test.ReportError(t, l, params.ElementLength)
+	}
 	got, _ = I.MarshalBinaryCompress()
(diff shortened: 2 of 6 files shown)


group/group_test.go
```diff
@@ -139,21 +139,32 @@ func testMarshal(t *testing.T, testTimes int, g group.Group) {
 		test.ReportError(t, I, II)
 	}
 
+	sizes := g.Len()
 	got1 := g.NewElement()
 	got2 := g.NewElement()
 	for i := 0; i < testTimes; i++ {
 		x := g.RandomElement(rand.Reader)
```
lukevalenta, line 146: Should we also have a test for unmarshaling the point at infinity since the length will not match in that case?

group/group.go
```diff
@@ -30,6 +30,7 @@ type Group interface {
 	RandomScalar(io.Reader) Scalar
 	HashToElement(data, dst []byte) Element
 	HashToScalar(data, dst []byte) Scalar
+	Len() Length
```
chris-wood, line 33:
```suggestion
	Length() Length
```

Alternatively, can we just inline the Length struct here? I don't see why this layer of indirection is useful.

group/short.go
```diff
@@ -132,13 +140,19 @@ func (e *wElt) MarshalBinaryCompress() ([]byte, error) {
 func (e *wElt) UnmarshalBinary(b []byte) error {
 	byteLen := (e.Params().BitSize + 7) / 8
 	l := len(b)
-	switch {
-	case l == 1 && b[0] == 0x00: // point at infinity
+	if l < 1 {
+		return ErrUnmarshal
+	}
+	switch b[0] {
+	case 0x00: // point at infinity
```
lukevalenta, line 147: Should there be a check here for `l == 1`?

chris-wood (changes requested):

group/group.go
lukevalenta: The `Length` function is needed b/c this is an interface, right?

chris-wood: Sorry, I mean, add three functions on this interface: one for element length, one for compressed element length, and one for scalar length. I don't see why we need a function on the interface that just returns this struct. Alternatively, you could generalize this a bit and call it `GroupParams`, returning a struct called `GroupParams`, that has these fields in it. Then you could add other group-related parameters, sort of like [what Go does with curves](https://golang.org/pkg/crypto/elliptic/#CurveParams).

group/short.go
armfazh: done

group/group.go
armfazh: Renaming done.
Luke: Yes, we need this function in the interface.

group/group_test.go
armfazh: added now

group/group.go
```diff
@@ -30,6 +30,11 @@ type Group interface {
 	RandomScalar(io.Reader) Scalar
 	HashToElement(data, dst []byte) Element
 	HashToScalar(data, dst []byte) Scalar
+	Length() struct {
```
chris-wood, line 33: Can we not do what Go does for curves, and make this a more generic "GroupParams" function?

armfazh: I just see your previous comment, will do some changes accrodingly. I like the idea of `GroupParams` structure

armfazh: Yet another try, pls take a look.

group/group.go
```diff
@@ -19,8 +19,15 @@ var (
...
 
+type Params struct {
+	ElementLength           uint // Length in bytes of an element.
+	CompressedElementLength uint // Length in bytes of a compressed element.
+	ScalarLength            uint // Length in bytes of a scalar.
+}
+
 // Group represents a prime-order group based on elliptic curves.
 type Group interface {
+	GetParams() Params
```
chris-wood, line 30: Does it need to be called `GetParams`, instead of simply `Params`? Go's elliptic curve package has a nice pattern:

```
type Curve interface {
    // Params returns the parameters for the curve.
    Params() *CurveParams
   ...
```

armfazh: Updated now, and called `Params` to satisfy linter:
```
type name will be used as group.GroupParams by other packages, and that stutters; consider calling this Params
```

chris-wood (approved):
LGTM!

group/group.go
chris-wood: Much better -- thanks!

lukevalenta (approved):
Looks good to me!

PR #234, opened 2021-06-24
sike: update test vectors

armfazh:
updates test vectors to SIKE library v3.4


claucece (approved):

lukevalenta (commented):
Any idea what has changes in these test vector files? Visually inspecting the files I can't find any differences but tools like `diff` and `git diff` are saying they've changed.

lukevalenta (approved):

armfazh:
> Any idea what has changes in these test vector files? Visually inspecting the files I can't find any differences but tools like `diff` and `git diff` are saying they've changed.

Nice catch!

Files differ in line endings only.
PQCkemKAT_374.rsp: DOS/Windows (CR-LF)
PQCkemKAT_374.rsp.old: UNIX (LF)

PR #236, opened 2021-06-25
sidh: no return parameter on isogeny EvaluatePoint

armfazh:
avoids to copy data when returning from isogeny evaluation.

Updates the template and three instances generated with it.


lukevalenta (approved):
Looks good to me!

PR #237, opened 2021-06-29
Updating linter version to v1.41

armfazh:
Bumping up golangci-lint version to v1.41
- replacing invocations of math/rand by crypto/rand

### DIFF
--- a/pke/kyber/internal/common/field_test.go
+++ b/pke/kyber/internal/common/field_test.go
@@ -36,29 +36,29 @@ func TestBarrettReduceFull(t *testing.T) {
 	}
 }
 
-func randSliceUint32(N uint) []uint32 {
-	bytes := make([]uint8, 4*N)
+func randSliceUint32WithMax(length uint, max uint32) []uint32 {
+	bytes := make([]uint8, 4*length)
 	n, err := rand.Read(bytes)
 	if err != nil {
 		panic(err)
 	} else if n < len(bytes) {
 		panic("short read from RNG")
 	}
-	x := make([]uint32, N)
+	x := make([]uint32, length)
 	for i := range x {
-		x[i] = binary.LittleEndian.Uint32(bytes[4*i:])
+		x[i] = binary.LittleEndian.Uint32(bytes[4*i:]) % max
 	}
 	return x
 }
 
 func TestMontReduce(t *testing.T) {
 	N := 1000
-	r := randSliceUint32(uint(N))
 	max := uint32(Q) * (1 << 16)
 	mid := int32(Q) * (1 << 15)
+	r := randSliceUint32WithMax(uint(N), max)
 
 	for i := 0; i < N; i++ {
-		x := int32(r[i]%max) - mid
+		x := int32(r[i]) - mid
 		y := montReduce(x)
 		if modQ32(x) != modQ32(int32(y)*(1<<16)) {
 			t.Fatalf("%d", x)
--- a/pke/kyber/internal/common/field_test.go
+++ b/pke/kyber/internal/common/field_test.go
@@ -1,8 +1,9 @@
 package common
 
 import (
+	"crypto/rand"
+	"encoding/binary"
 	"flag"
-	mathRand "math/rand"
 	"testing"
 )
 
@@ -35,9 +36,29 @@ func TestBarrettReduceFull(t *testing.T) {
 	}
 }
 
+func randSliceUint32(N uint) []uint32 {
+	bytes := make([]uint8, 4*N)
+	n, err := rand.Read(bytes)
+	if err != nil {
+		panic(err)
+	} else if n < len(bytes) {
+		panic("short read from RNG")
+	}
+	x := make([]uint32, N)
+	for i := range x {
+		x[i] = binary.LittleEndian.Uint32(bytes[4*i:])
+	}
+	return x
+}
+
 func TestMontReduce(t *testing.T) {
-	for i := 0; i < 1000; i++ {
-		x := mathRand.Int31n(int32(Q)*(1<<16)) - int32(Q)*(1<<15)
+	N := 1000
+	r := randSliceUint32(uint(N))
+	max := uint32(Q) * (1 << 16)
+	mid := int32(Q) * (1 << 15)
+
+	for i := 0; i < N; i++ {
+		x := int32(r[i]%max) - mid
 		y := montReduce(x)
 		if modQ32(x) != modQ32(int32(y)*(1<<16)) {
 			t.Fatalf("%d", x)
--- a/pke/kyber/internal/common/ntt_test.go
+++ b/pke/kyber/internal/common/ntt_test.go
@@ -31,18 +31,18 @@ func BenchmarkInvNTTGeneric(b *testing.B) {
 }
 
 func (p *Poly) Rand() {
-	r := randSliceUint32(uint(N))
 	max := uint32(Q)
+	r := randSliceUint32WithMax(uint(N), max)
 	for i := 0; i < N; i++ {
-		p[i] = int16(r[i] % max)
+		p[i] = int16(r[i])
 	}
 }
 
 func (p *Poly) RandAbsLeQ() {
-	r := randSliceUint32(uint(N))
 	max := 2 * uint32(Q)
+	r := randSliceUint32WithMax(uint(N), max)
 	for i := 0; i < N; i++ {
-		p[i] = int16(int32(r[i]%max) - int32(Q))
+		p[i] = int16(int32(r[i]) - int32(Q))
 	}
 }
 
--- a/pke/kyber/internal/common/ntt_test.go
+++ b/pke/kyber/internal/common/ntt_test.go
@@ -1,9 +1,6 @@
 package common
 
-import (
-	mathRand "math/rand"
-	"testing"
-)
+import "testing"
 
 func BenchmarkNTT(b *testing.B) {
 	var a Poly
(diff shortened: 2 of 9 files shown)


cjpatton:
>     * replacing invocations of math/rand by crypto/rand

Was this change made by the linter or did you make it manually?

armfazh:
> > ```
> > * replacing invocations of math/rand by crypto/rand
> > ```
> 
> Was this change made by the linter or did you make it manually?

I manually changed the instances reported by linter as insecure.

pke/kyber/internal/common/field_test.go
```diff
@@ -36,8 +37,14 @@ func TestBarrettReduceFull(t *testing.T) {
 }
 
 func TestMontReduce(t *testing.T) {
-	for i := 0; i < 1000; i++ {
-		x := mathRand.Int31n(int32(Q)*(1<<16)) - int32(Q)*(1<<15)
+	N := 1000
+	pp := make([]uint8, 4*N)
+	_, _ = rand.Read(pp)
```
cjpatton, line 42: As a general rule, I think errors should always be handled.

pke/kyber/internal/common/ntt_test.go
```diff
@@ -34,14 +35,20 @@ func BenchmarkInvNTTGeneric(b *testing.B) {
...
 	}
 }
 
 func (p *Poly) RandAbsLeQ() {
+	pp := make([]uint8, 4*N)
+	_, _ = rand.Read(pp)
+	max := 2 * uint32(Q)
 	for i := 0; i < N; i++ {
-		p[i] = int16(mathRand.Intn(int(2*Q))) - Q
+		p[i] = int16(int32(binary.LittleEndian.Uint32(pp[4*i:])%max) - int32(Q))
```
cjpatton, line 51: This loop gets plopped down a lot. Is it possible to ractor the code so that the common functionality is implemented by a single function? 
```go
func setRandPoly(p []int16) err {
        pp := make([]uint8, 4*N)
	if n, err := rand.Read(pp); err != nil {
            panic!(err)
        else if n < len(pp) {
            panic!("short read from RNG")
	}
	max := 2 * uint32(Q)
	for i := 0; i < N; i++ {
		p[i] = int16(int32(binary.LittleEndian.Uint32(pp[4*i:])%max)
        }
}
```

cjpatton (changes requested):

pke/kyber/internal/common/field_test.go
```diff
@@ -36,8 +37,14 @@ func TestBarrettReduceFull(t *testing.T) {
...
+	max := uint32(Q) * (1 << 16)
+	mid := int32(Q) * (1 << 15)
+	for i := 0; i < N; i++ {
+		r := binary.LittleEndian.Uint32(pp[4*i:])
+		x := int32(r%max) - mid
```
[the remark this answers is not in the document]
armfazh: Just reproducing the original formula
```patch
-		x := mathRand.Int31n(int32(Q)*(1<<16)) - int32(Q)*(1<<15)
```
the original code samples from [-2^15, 2^15).

pke/kyber/internal/common/ntt_test.go
armfazh: nice suggestion, refactored now

pke/kyber/internal/common/poly_test.go
```diff
@@ -1,15 +1,16 @@
...
-	mathRand "math/rand"
 	"testing"
 )
 
 func (p *Poly) RandAbsLe9Q() {
+	r := randSliceUint32(uint(N))
+	max := 9 * uint32(Q)
 	for i := 0; i < N; i++ {
-		p[i] = int16(mathRand.Intn(18*int(Q) - 9*int(Q)))
+		p[i] = int16(int32(r[i] % max))
```
cjpatton, line 13: This change looks equivalent to me. However, the way the original code was written is weird. (Why `18 * int(Q) - 9 * int(Q)` instead of just `9 * int(Q)`?) This makes me think it might be worth double checking that this is the correct range.

pke/kyber/internal/common/poly_test.go
```diff
@@ -26,7 +27,11 @@ func TestDecompressMessage(t *testing.T) {
 	var m, m2 [PlaintextSize]byte
 	var p Poly
 	for i := 0; i < 1000; i++ {
-		_, _ = cryptoRand.Read(m[:])
+		_, err := rand.Read(m[:])
```
cjpatton, line 30: nit: I would handle the first output handle paraemter.

sign/dilithium/internal/common/field_test.go
```diff
@@ -1,16 +1,34 @@
...
+	for i := range x {
+		x[i] = binary.LittleEndian.Uint32(bytes[4*i:])
+	}
+	return x
+}
+
 func TestModQ(t *testing.T) {
-	for i := 0; i < 1000; i++ {
-		x := rand.Uint32()
+	const testTimes = 1000
```
cjpatton, line 28: `testTimes` is written as `N` above. It might be a good idea to be consistent.

pke/kyber/internal/common/field_test.go
```diff
@@ -35,9 +36,29 @@ func TestBarrettReduceFull(t *testing.T) {
 	}
 }
 
+func randSliceUint32(N uint) []uint32 {
```
cjpatton, line 39: It seems like most everywhere this function is used there is an effective max. Why not add the max as a parameter here?

sign/dilithium/internal/common/pack_test.go
```diff
@@ -1,16 +1,19 @@
...
 	"testing"
 )
 
 func TestPackLe16AgainstGeneric(t *testing.T) {
 	var p Poly
 	var buf1, buf2 [PolyLe16Size]byte
+	pp := make([]uint8, 256)
+
 	for j := 0; j < 1000; j++ {
+		_, _ = rand.Read(pp)
```
cjpatton, line 14: How about re-reusing randSliceUint32 here?
```go
p = randSliceUint32(256)
for i := range p {
   p[i] &= 0xF
}
```

If you end up passing a max parameter, this is just `p = randUint32Slice(256, 0xF)`.

cjpatton (changes requested):
Just some minor readability things.

pke/kyber/internal/common/poly_test.go
armfazh: @bwesterb could you please give us some light here?

armfazh: let's move this change forward, and do changes in other PR if bwesterb thinks it's needed.

pke/kyber/internal/common/poly_test.go
armfazh: resolved

sign/dilithium/internal/common/field_test.go
armfazh: N is a package parameter, not the number of tests

pke/kyber/internal/common/field_test.go
armfazh: good one!

sign/dilithium/internal/common/pack_test.go
armfazh: excelent suggestion

pke/kyber/internal/common/poly_test.go
```diff
@@ -27,8 +27,8 @@ func TestDecompressMessage(t *testing.T) {
 	var m, m2 [PlaintextSize]byte
 	var p Poly
 	for i := 0; i < 1000; i++ {
-		_, err := rand.Read(m[:])
-		if err != nil {
+		n, err := rand.Read(m[:])
+		if err != nil || n != len(m) {
 			t.Error(err)
```
cjpatton, line 32: if `err == nil` then the output here will read `<nil>`. What about
```go
if n, err := rand.Read(m[:]); err != nil {
   t.Fatal(err)
} else if n != len(m) {
   t.Fatal("short read from RNG")
}
```

sign/dilithium/internal/common/field_test.go
```diff
@@ -4,22 +4,25 @@ import (
...
 	"testing"
 )
 
 var runVeryLongTest = flag.Bool("very-long", false, "runs very long tests")
 
-func randSliceUint32(N uint) []uint32 {
-	bytes := make([]uint8, 4*N)
+func randSliceUint32(length uint) []uint32 { return randSliceUint32Max(length, math.MaxUint32) }
+
+func randSliceUint32Max(length uint, max uint32) []uint32 {
```
cjpatton, line 15: I would suggest something like `randSliceUint32WithMax` to make it clear that this function has an *additional* parameter called `max`.

pke/kyber/internal/common/field_test.go
```diff
@@ -35,9 +36,29 @@ func TestBarrettReduceFull(t *testing.T) {
 	}
 }
 
+func randSliceUint32Max(N uint, max uint32) []uint32 {
```
cjpatton, line 39: nit: s/N/length/ in order to match the reimplementation of this function below.

sign/dilithium/internal/common/field_test.go
cjpatton: Oh, gotcha.

cjpatton (approved):
Approved, with a few suggestions.

pke/kyber/internal/common/poly_test.go
armfazh: that's right

sign/dilithium/internal/common/field_test.go
armfazh: done

pke/kyber/internal/common/field_test.go
armfazh: resolved

cjpatton (approved):

pke/kyber/internal/common/poly_test.go
bwesterb: The intended code was `int16(mathRand.Intn(18*int(Q)) - mathRand.Intn(9*int(Q)))`.

armfazh: Thanks Bas, I will fix this on #240

PR #239, opened 2021-07-03
Includes test vectors for hashing to curve

armfazh:
Includes test to verify hash to curve functionality matches with latest hash_to_curve IETF draft.
https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-hash-to-curve-11


claucece (approved):
LGTM ;)

PR #241, opened 2021-07-08
oprf: updating VOPRF test vectors to v07

armfazh:
Updating VOPRF test vectors to v07
https://datatracker.ietf.org/doc/draft-irtf-cfrg-voprf/07
JSON: https://github.com/cfrg/draft-irtf-cfrg-voprf/blob/draft-irtf-cfrg-voprf-07/poc/vectors/allVectors.json


claucece (approved):
This is great! Thank you!

PR #245, opened 2021-07-15
Remove CI integration for ancient go versions

wbl:
We should follow go's versioning policy to take advantage of new
library and language features.


cjpatton (approved):
Works for me!

armfazh (approved):

PR #249, opened 2021-07-17
bls12: exposing some methods of Gt and sizes for elements

armfazh:


ecc/bls12381/pair.go
```diff
@@ -1,3 +1,8 @@
+// Package bls12381 provides bilinear pairings using the BLS12-381 curve.
+//
+// A pairing system consists of three groups G1 and G2 (adiitive notation) and
+// Gt (multiplicative notation) of the same order.
+// Scalars can be used interchangebly between groups.
```
armfazh, line 5:
```suggestion
// Scalars can be used interchangeably between groups.
```

wbl (approved):

PR #255, opened 2021-07-29
group: colon is not a valid character in filenames

armfazh:
Siimilar to https://github.com/armfazh/h2c-go-ref/pull/4


wbl (changes requested):
Can we add a check to validate filenames in committed files before merge?

armfazh:
> Can we add a check to validate filenames in committed files before merge?

#256 is currently failing on windows, so this PR solves that issue, and we will be able to detect this issue through CI

wbl (approved):

PR #258, opened 2021-08-02
bls12: hash to G1 complaint with h2c draft

armfazh:
Implements encode and hash to curve as recommended in IETF draft. See https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-hash-to-curve-11

Removes insecure try-and-increment method for hashing.

Changes:
 - Implements encode and hash to curve as recommended in IETF draft.
 - G1 ciphersuites for BLS12-381.
 - Removes certain memory allocations.
 - Adding test vectors for hash to G1.
 - Return int rather than bool in ff predicates.

### DIFF
--- a/ecc/bls12381/constants.go
+++ b/ecc/bls12381/constants.go
@@ -9,34 +9,110 @@ type Scalar = ff.Scalar
 const ScalarSize = ff.ScalarSize
 
 var (
-	g1ParamB  = ff.Fp{ /*4*/ }  // 4
-	g1Param3B = ff.Fp{ /*12*/ } // 3*G1ParamB
-	g1GenX    = ff.Fp{
-		// 0xfb3af00adb22c6bb, 0x6c55e83ff97a1aef, 0xa14e3a3f171bac58,
-		// 0xc3688c4f9774b905, 0x2695638c4fa9ac0f, 0x17f1d3a73197d794,
+	g1Params struct{ b, _3b, genX, genY ff.Fp }
+	g2Params struct{ b, _3b, genX, genY ff.Fp2 }
+	g1Isog11 struct {
+		a, b, c2 ff.Fp
+		c1       [ff.FpSize]byte
+		cofactor Scalar
+		xNum     [12]ff.Fp
+		xDen     [11]ff.Fp
+		yNum     [16]ff.Fp
+		yDen     [16]ff.Fp
 	}
-	g1GenY = ff.Fp{
-		// 0xcaa232946c5e7e1, 0xd03cc744a2888ae4, 0xdb18cb2c04b3ed,
-		// 0xfcf5e095d5d00af6, 0xa09e30ed741d8ae4, 0x8b3f481e3aaa0f1,
-	}
-	g2ParamB  = ff.Fp2{}
-	g2Param3B = ff.Fp2{}
-	g2GenX    = ff.Fp2{}
-	g2GenY    = ff.Fp2{}
 )
 
-func init() {
-	g1ParamB.SetUint64(4)
-	g1Param3B.SetUint64(12)
-	_ = g1GenX.SetString("0x17f1d3a73197d7942695638c4fa9ac0fc3688c4f9774b905a14e3a3f171bac586c55e83ff97a1aeffb3af00adb22c6bb")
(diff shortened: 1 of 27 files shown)


ecc/bls12381/constants.go
```diff
@@ -9,34 +9,110 @@ type Scalar = ff.Scalar
 const ScalarSize = ff.ScalarSize
 
 var (
-	g1ParamB  = ff.Fp{ /*4*/ }  // 4
-	g1Param3B = ff.Fp{ /*12*/ } // 3*G1ParamB
-	g1GenX    = ff.Fp{
-		// 0xfb3af00adb22c6bb, 0x6c55e83ff97a1aef, 0xa14e3a3f171bac58,
-		// 0xc3688c4f9774b905, 0x2695638c4fa9ac0f, 0x17f1d3a73197d794,
+	g1Params struct{ B, _3B, GenX, GenY ff.Fp }
```
wbl, line 12: having a private struct with public members is a bit odd

ecc/bls12381/ff/cyclo6.go
```diff
@@ -8,8 +8,8 @@ type Cyclo6 Fp12
 func (z Cyclo6) String() string { return (Fp12)(z).String() }
 
 func (z *Cyclo6) Set(x *Cyclo6)                  { (*Fp12)(z).Set((*Fp12)(x)) }
-func (z Cyclo6) IsEqual(x *Cyclo6) bool          { return (Fp12)(z).IsEqual((*Fp12)(x)) }
-func (z Cyclo6) IsIdentity() bool                { i := &Fp12{}; i.SetOne(); return z.IsEqual((*Cyclo6)(i)) }
+func (z Cyclo6) IsEqual(x *Cyclo6) int           { return (Fp12)(z).IsEqual((*Fp12)(x)) }
```
wbl, line 11: Ideally we'd document what these return

ecc/bls12381/ff/fp.go
```diff
@@ -25,23 +25,30 @@ func (z *Fp) SetUint64(n uint64)       { z.toMont(&fpRaw{n}) }
...
 func (z *Fp) Sub(x, y *Fp)             { fiatFpMontSub(&z.i, &x.i, &y.i) }
 func (z *Fp) Mul(x, y *Fp)             { fiatFpMontMul(&z.i, &x.i, &y.i) }
 func (z *Fp) Sqr(x *Fp)                { fiatFpMontSquare(&z.i, &x.i) }
-func (z *Fp) Inv(x *Fp)                { z.expVarTime(x, fpOrderMinus2[:]) }
+func (z *Fp) Inv(x *Fp)                { z.ExpVarTime(x, fpOrderMinus2[:]) }
 func (z *Fp) toMont(in *fpRaw)         { fiatFpMontMul(&z.i, in, &fpRSquare) }
 func (z Fp) fromMont() (out fpRaw)     { fiatFpMontMul(&out, &z.i, &fpMont{1}); return }
+func (z Fp) Sgn0() int                 { zz := z.fromMont(); return int(zz[0] & 1) }
+func (z *Fp) CMov(x, y *Fp, b int) {
+	mask := 0 - uint64(b&0x1)
```
wbl, line 40: We should document similar to https://pkg.go.dev/crypto/subtle#ConstantTimeSelect

ecc/bls12381/g1.go
```diff
@@ -52,14 +54,17 @@ func (g *G1) Neg() { g.y.Neg() }
 // SetIdentity assigns g to the identity element.
 func (g *G1) SetIdentity() { g.x = ff.Fp{}; g.y.SetOne(); g.z = ff.Fp{} }
 
+// isZeroProj returns true if the point is not a projective point.
```
wbl, line 57: Not the greatest of names imho: sounds like it checks if the point is the identity of the curve in projective representation. How about "isNotP1Point"?

wbl (approved):

ecc/bls12381/g1.go
armfazh: changed to `isValidProjective`.
Interestingly, it would be good to avoid this case by-design.

ecc/bls12381/ff/fp.go
armfazh: I added a string,  I will devote an entire PR only for documentation.

ecc/bls12381/constants.go
armfazh: done

PR #265, opened 2021-08-11
Optimize multiplication and make it constant time

wbl:


ecc/bls12381/g1.go
```diff
@@ -174,10 +182,40 @@ func (g *G1) Add(P, Q *G1) {
 // ScalarMult calculates g = kP.
 func (g *G1) ScalarMult(k *Scalar, P *G1) { g.scalarMult(k.Bytes(), P) }
 
+// ScalarMultVarTime calculates g=kP, leaking k, for short k
+func (g *G1) ScalarMultVarTimeShort(k *Scalar, P *G1) { g.scalarMultVarTimeShort(k.Bytes(), P) }
```
armfazh, line 186: `ScalarMultVarTimeShort`--> `ScalarMultVarTime`

Not sure if we want to expose this function

ecc/bls12381/g1.go
```diff
@@ -174,10 +182,40 @@ func (g *G1) Add(P, Q *G1) {
...
+		idx := 0xf & (k[i/8] >> uint(i%8))
+		for j := 0; j < 16; j++ {
+			T.cmov(&mults[j], subtle.ConstantTimeByteEq(idx, uint8(j)))
+		}
+		Q.Add(&Q, T)
+	}
+	g.Set(&Q)
+}
+
+// Because the arguments are so short, nothing helps
```
armfazh, line 214: is this part of the documentation of this function? if so, better to be more explicit

ecc/bls12381/g1.go
wbl: let's keep it internal. I added short to emphasize that the tradeoffs are different.

ecc/bls12381/g1.go
```diff
@@ -177,7 +185,35 @@ func (g *G1) ScalarMult(k *Scalar, P *G1) { g.scalarMult(k.Bytes(), P) }
...
+			T.cmov(&mults[j], subtle.ConstantTimeByteEq(idx, uint8(j)))
+		}
+		Q.Add(&Q, T)
+	}
+	g.Set(&Q)
+}
+
+// scalarMultShort multiplies by short, constant scalars.
+func (g *G1) scalarMultShort(k []byte, P *G1) {
+	// Since the scalar is short and low Hamming weight not much helps.
```
bwesterb, line 213: Addition chains might help — in this case I found one with one double less, which is not worth the trouble.

bwesterb (approved):

armfazh (approved):
Agree with Bas' comments.

Issue #257, opened 2021-08-01
bls12: make serialization compatible with other implementations


armfazh:
The zkcrypto's implementation defines a particular format for serialization of G1 and G2 elements.
See  https://raw.githubusercontent.com/zkcrypto/bls12_381/0.1.1/src/notes/serialization.rs

PR #272, opened 2021-09-05
bls12: changing serialization of Fp to big-endian format

armfazh:
These changes are needed to accomplish #257 .

### DIFF
--- a/ecc/bls12381/gt.go
+++ b/ecc/bls12381/gt.go
@@ -8,14 +8,16 @@ const GtSize = ff.URootSize
 // Gt represents an element of the output (multiplicative) group of a pairing.
 type Gt struct{ i ff.URoot }
 
-func (z Gt) String() string               { return z.i.String() }
-func (z *Gt) Set(x *Gt)                   { z.i.Set(&x.i) }
-func (z *Gt) SetBytes(b []byte) error     { return z.i.SetBytes(b) }
-func (z Gt) Bytes() []byte                { return z.i.Bytes() }
-func (z *Gt) SetIdentity()                { z.i.SetIdentity() }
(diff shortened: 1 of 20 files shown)


ecc/bls12381/gt.go
```diff
@@ -8,14 +8,16 @@ const GtSize = ff.URootSize
...
-func (z *Gt) Set(x *Gt)                   { z.i.Set(&x.i) }
-func (z *Gt) SetBytes(b []byte) error     { return z.i.SetBytes(b) }
-func (z Gt) Bytes() []byte                { return z.i.Bytes() }
-func (z *Gt) SetIdentity()                { z.i.SetIdentity() }
-func (z Gt) IsEqual(x *Gt) bool           { return z.i.IsEqual(&x.i) == 1 }
-func (z Gt) IsIdentity() bool             { i := &Gt{}; i.SetIdentity(); return z.IsEqual(i) }
-func (z *Gt) Mul(x, y *Gt)                { z.i.Mul(&x.i, &y.i) }
-func (z *Gt) Sqr(x *Gt)                   { z.i.Sqr(&x.i) }
-func (z *Gt) Inv(x *Gt)                   { z.i.Inv(&x.i) }
+func (z Gt) String() string           { return z.i.String() }
```
wbl, line 11: I'm not sure why these changed. And now that I look at then I'm not sure why we pass-by-value sometimes.

wbl (approved):

wbl:
Would be nice to see test vectors confirming interop in tests.

ecc/bls12381/gt.go
armfazh: As it is passed by value, this ensures that the method will not modify the value (by-accident).

armfazh:
> Would be nice to see test vectors confirming interop in tests.

Test vectors were added in #273 .

PR #275, opened 2021-09-05
bls12: removing Set functions

armfazh:
using native copy functions rather than using a setter function.


bwesterb:
I'm curious: what is your reasoning for removing `Set`?  For me both ways are fine.

bwesterb (approved):

armfazh:
> I'm curious: what is your reasoning for removing `Set`? For me both ways are fine.

calling functions as simple as `func (z *Type) f(x*Type) { *z =*x }` has a cost=5 for the inliner. This code has lots of nested expressions like this; which makes the cost increments for a simple operation such as assigning.

wbl (approved):

PR #277, opened 2021-09-10
bls12: implements Marshaler interfaces and SetBytes for Scalar

armfazh:
Implements the Marshaler interfaces from go stdlib for serialization and deserialization.
Also, Scalar exposes `SetBytes([]byte)` to set scalars using variable-length data.

### DIFF
--- a/ecc/bls12381/ff/common.go
+++ b/ecc/bls12381/ff/common.go
@@ -35,14 +35,21 @@ func setString(in string, order []byte) ([]uint64, error) {
 		return nil, errInputRange
 	}
 	inBytes := inBig.FillBytes(make([]byte, len(order)))
-	return setBytes(inBytes, order)
+	return setBytesBounded(inBytes, order)
 }
 
-func setBytes(in []byte, order []byte) ([]uint64, error) {
+func setBytesBounded(in []byte, order []byte) ([]uint64, error) {
 	if isLessThan(in, order) == 0 {
 		return nil, errInputRange
 	}
 	return conv.BytesBe2Uint64Le(in), nil
+}
+
+func setBytesUnbounded(in []byte, order []byte) []uint64 {
+	inBig := new(big.Int).SetBytes(in)
+	inBig.Mod(inBig, new(big.Int).SetBytes(order))
+	inBytes := inBig.FillBytes(make([]byte, len(order)))
+	return conv.BytesBe2Uint64Le(inBytes)
 }
 
 // isLessThan returns 1 if 0 <= x < y, otherwise 0. Assumes that slices have the same length.
--- a/ecc/bls12381/ff/fp12.go
+++ b/ecc/bls12381/ff/fp12.go
@@ -8,11 +8,7 @@ const Fp12Size = 2 * Fp6Size
 // Fp12 represents an element of the field Fp12 = Fp6[w]/(w^2-v)., where v in Fp6.
 type Fp12 [2]Fp6
 
-func (z Fp12) String() string { return fmt.Sprintf("0: %v\n1: %v", z[0], z[1]) }
-func (z *Fp12) SetBytes(b []byte) error {
-	return errFirst(z[0].SetBytes(b[:Fp6Size]), z[1].SetBytes(b[Fp6Size:2*Fp6Size]))
-}
-func (z Fp12) Bytes() []byte       { return append(z[0].Bytes(), z[1].Bytes()...) }
+func (z Fp12) String() string      { return fmt.Sprintf("0: %v\n1: %v", z[0], z[1]) }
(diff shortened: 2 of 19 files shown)


ecc/bls12381/ff/common.go
```diff
@@ -35,16 +35,23 @@ func setString(in string, order []byte) ([]uint64, error) {
...
+func setBytesBounded(in []byte, order []byte) ([]uint64, error) {
 	if isLessThan(in, order) == 0 {
 		return nil, errInputRange
 	}
 	return conv.BytesBe2Uint64Le(in), nil
 }
 
+func setBytesUnbounded(in []byte, order []byte) []uint64 {
+	inBig := new(big.Int).SetBytes(in)
+	inBig.Mod(inBig, new(big.Int).SetBytes(order))
```
bwesterb, line 50: Are all scalars non-secret?

ecc/bls12381/ff/fp12.go
```diff
@@ -72,6 +68,22 @@ func (z *Fp12) ExpVarTime(x *Fp12, n []byte) {
...
 
+func (z *Fp12) UnmarshalBinary(b []byte) error {
+	return errFirst(
+		z[1].UnmarshalBinary(b[:Fp6Size]),
+		z[0].UnmarshalBinary(b[Fp6Size:2*Fp6Size]),
+	)
+}
+
+func (z Fp12) MarshalBinary() (b []byte, e error) {
+	if b1, e := z[1].MarshalBinary(); e == nil {
```
bwesterb, line 79: You don't return the error because `:=` create a second instance of `e` that shows the return value.

ecc/bls12381/ff/fp2.go
```diff
@@ -56,6 +52,22 @@ func (z *Fp2) Inv(x *Fp2) {
...
 
+func (z *Fp2) UnmarshalBinary(b []byte) error {
+	return errFirst(
+		z[1].UnmarshalBinary(b[:FpSize]),
+		z[0].UnmarshalBinary(b[FpSize:2*FpSize]),
+	)
+}
+
+func (z Fp2) MarshalBinary() (b []byte, e error) {
+	if b1, e := z[1].MarshalBinary(); e == nil {
```
bwesterb, line 63: Same issue.

ecc/bls12381/ff/fp6.go
```diff
@@ -144,6 +136,25 @@ func (z *Fp6) Frob(x *Fp6) {
 	z[2].Mul(&z[2], &Fp2{frob6V2, Fp{}})
 }
 
+func (z Fp6) MarshalBinary() (b []byte, e error) {
+	if b2, e := z[2].MarshalBinary(); e == nil {
```
bwesterb, line 140: Same issue.

ecc/bls12381/ff/common.go
armfazh: this function must be marked as non-constant time. Do you have any ideas on how to perform this operation in constant time?

wbl: What length are we applying it to? I think I only need it for double length ones, in which case a reduction is all that is required.

ecc/bls12381/g1.go
```diff
@@ -98,9 +98,12 @@ func (g G1) encodeBytes(compressed bool) []byte {
 		isBigYCoord = byte(g.y.IsNegative())
 	}
 
-	bytes := g.x.Bytes()
+	bytes, e := g.x.MarshalBinary()
+	err(e)
```
wbl, line 102: Given that this is only used as part of the MarshalBinary, why can't we propagate the error? Obviously there never should be an error, but in that case bytes, _ would be clearer to me.

wbl (changes requested):

ecc/bls12381/ff/fp12.go
armfazh: fixed now

ecc/bls12381/ff/fp2.go
```diff
@@ -56,6 +52,22 @@ func (z *Fp2) Inv(x *Fp2) {
 	z[1].Neg()
 }
 
+func (z *Fp2) UnmarshalBinary(b []byte) error {
+	return errFirst(
+		z[1].UnmarshalBinary(b[:FpSize]),
```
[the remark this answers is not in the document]
armfazh: good catch

ecc/bls12381/g1.go
armfazh: updated now

bwesterb (approved):
Provided we fix the constant-time issue soon enough.

wbl (approved):

PR #290, opened 2021-09-24
kem: test that KEMs generate more than one secret

bwesterb:


kem/schemes/schemes_test.go
```diff
@@ -114,6 +114,17 @@ func TestApi(t *testing.T) {
 				t.Fatal()
 			}
 
+			ct3, ss3, err := scheme.Encapsulate(pk2)
```
armfazh, line 117: it can also test that the decapsulation of `ct3` matches `ss3`

armfazh (approved):

Issue #295, opened 2021-10-07
bls12: add addition chain for Fp inversion


armfazh:
add addition chain for Fp inversion

PR #296, opened 2021-10-08
bls12: addition chain for inversion in Fp

armfazh:
Using special addition chain for inversion in Fp.

Chain was found using [mmcloughlin/addchain](https://github.com/mmcloughlin/addchain/tree/v0.3.0) v0.3.0 
McLoughlin, Michael Ben. (2021). https://doi.org/10.5281/zenodo.4758226

```
benchmark               old ns/op     new ns/op     delta
BenchmarkFp12/Inv-8     112649        89706         -20.37%
BenchmarkFp2/Inv-8      95025         68185         -28.25%
BenchmarkFp6/Inv-8      100276        73924         -26.28%
BenchmarkFp/Inv-8       100091        70305         -29.76%
```

cc: @mmcloughlin


armfazh:
Fixes #295

PR #300, opened 2021-10-26
oprf: updates test vectors to draft voprf-v8

armfazh:
Compatible with https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-voprf-08


lukevalenta (approved):
LGTM! Confirmed that text vectors match https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-voprf-08#appendix-A.

PR #301, opened 2021-11-03
group: add SetUint64 method to Group.Scalar.

armfazh:


chris-wood (approved):

group/ristretto255.go
```diff
@@ -145,6 +146,8 @@ func (e *ristrettoElement) UnmarshalBinary(data []byte) error {
 	return e.p.UnmarshalBinary(data)
 }
 
+func (s *ristrettoScalar) SetUint64(n uint64) { s.s.SetBigInt(new(big.Int).SetUint64(n)) }
```
bwesterb, line 149: Better to use `SetBytes`. Scalars are encoded little endian, so simply write the uint64 into the `[32]byte` at the first 8 bytes using encoding/binary.LittleEndian.

bwesterb: I can also add a SetUint64.

PR #305, opened 2021-12-02
group: adds a function for sampling non-zero scalars.

### LINKED ISSUE #303: Random non-zero scalar generation
Currently, `Group.RandomScalar()` returns a scalar from the group Z_p, which includes the zero element. Some protocols require randomly sampling a scalar from the set of scalars with a multiplicative inverse (Z_p^*). One can implement this with `Group.RandomScalar`, but I wonder if adding a function to the Group interface would be useful. Thoughts?

armfazh:
Fixes #303 

check commit: 1baa8f29f3d421dc13a87a7935bdd4e8cf7a1151


chris-wood:
Is it possible to keep `f13259a` out of this PR? It _seems_ unrelated.

go.mod
```diff
@@ -4,6 +4,6 @@ go 1.15
 
 require (
 	github.com/bwesterb/go-ristretto v1.2.1
-	golang.org/x/crypto v0.0.0-20210921155107-089bfa567519
-	golang.org/x/sys v0.0.0-20211007075335-d3039528d8ac
+	golang.org/x/crypto v0.0.0-20211117183948-ae814b36b871
+	golang.org/x/sys v0.0.0-20211124211545-fe61309f8881
```
bwesterb, line 8: Is this intentional?

bwesterb (approved):

group/ristretto255.go
```diff
@@ -145,6 +155,7 @@ func (e *ristrettoElement) UnmarshalBinary(data []byte) error {
 	return e.p.UnmarshalBinary(data)
 }
 
+func (s *ristrettoScalar) String() string     { return fmt.Sprintf("0x%x", s.s.Bytes()) }
```
chris-wood, line 158: Do we need this?

chris-wood (approved):

go.mod
armfazh: Yes

group/ristretto255.go
armfazh: Yes.

PR #308, opened 2021-12-10
Add blind RSA protocol support

### LINKED ISSUE #307: Two message blind signature protocols
This issue tracks adding blind signature protocol support to CIRCL. We currently have a [digital signature package](https://github.com/cloudflare/circl/tree/master/sign) for computing non-interactive signatures, including ed2559 and ed448, as well as various post quantum algorithms. We don't have an API for interactive protocols that compute signatures, like we do for the [OPRF](https://github.com/cloudflare/circl/tree/master/oprf). 

Adding such an API is somewhat non-trivial as some protocols, like clause blind Schnorr, require more than two messages between client (verifier) and server (signer). However, it seems reasonable to start with two message protocols like blind RSA and blind BLS, rather than try and generalize to protocols with n >= 2 messages. 

To that end, here's a simple start:

~~~
type Verifier interface {
    Blind(message []byte) ([]byte, VerifierState, error) // return the message sent to the signer and verifier state, or an error
}

type VerifierState interface {
    Finalize(data []byte) ([]byte, error) // return the signature upon success, and an error otherwise
}

type Signer interface {
    BlindSign(data []byte) ([]byte, error) // return the signed, blinded message, or an error
}
~~~

One would use it like so:

~~~
verifier := NewFooVerifier(... public key, options ...)
signer := NewFooSigner(... private key, options ...)

msg := []byte("test")
req, state, err := veriifer.Blind(msg)
resp, err := signer.BlindSign(req)
sig, err := state.Finalize(resp)

// use (msg, sig) pair as needed
~~~

The abstraction across this API is a byte stream, so all of the scheme-specific details about the signature contents and protocol messages are hidden. We could choose to expose a bit more, but that seems unnecessarily complicated. 

@armfazh, @cjpatton: what do you think? I'd like to add [blind RSA](https://github.com/cfrg/draft-irtf-cfrg-blind-signatures) support, so if this seems reasonable I'll send a PR.

chris-wood:
This change adds a generic interface for two-message blind signature protocols, along with a concrete implementation of the blind RSA protocol as currently being specified by the CFRG. This protocol is a very simple wrapper around the existing RSA implementation in the Go standard library.

I'm starting this as a draft PR since I have questions around how we might create signers and verifiers, and how the ergonomics of the API look:

1. Should the `RSASigner` and `RSAVerifier` constructors accept `crypto.rsa` types as input, or more standardized formats for keying material (PKCS12 for private keys, for example)?
2. Currently, one controls signature randomness by either supplying (or not) a source of randomness to the `Blind` function. If the randomness is `nil`, the signature is deterministic, and vice versa. Internally, we still always need a source of randomness to generate the blind, which we pull from `crypto/rand`. Does this sort of knob for controlling signature randomness make sense, or should we instead have a `BlindDeterministic` (or whatever) variant?
3. Where do we stick the Go [LICENSE file](https://github.com/golang/go/blob/master/LICENSE) (which is required to reuse and modify their code)?
4. ... there are probably other things we might want to ask ourselves =) 

cc @wbl, @cjpatton, @claucece 

Closes #307.


blindsign/blindsign.go
```diff
@@ -0,0 +1,32 @@
...
+// ensures that the server does not learn anything about the client's
+// input during the BlindSign step.
+package blindsign
+
+// A Verifier represents a specific instance of a blind signature verifier.
+type Verifier interface {
+	// Blind a message according to the Signer's public key, producing an
+	// encoded protocol message and VerifierState for completing the protocol.
+	// Returns an error if any of the inputs are invalid or if the blinding
+	// step fails.
```
cjpatton, line 15: nit (here and below): I don't think this comment is necessary.
```suggestion
```

blindsign/blindrsa/blindrsa.go
```diff
@@ -0,0 +1,213 @@
...
+
+	return blindedMsg, RSAVerifierState{
+		encodedMsg: encodedMsg,
+		verifier:   v,
+		rInv:       rInv,
+	}, nil
+}
+
+// Blind initializes the blind RSA protocol using an input message and source of randomness. If
+// randomness is not provided, the output signature is deterministic. Otherwise, the signature
```
cjpatton, line 90: > the output signature is deterministic.

This appears not to be true, due to https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-rsa-blind-signatures-02#section-5.1.1.

cjpatton (commented):
Just a couple comments to start. API looks sensible enough. One question I had is whether blind signatures that use techniques other than RSA would be able to implement this API?

blindsign/blindrsa/blindrsa.go
chris-wood: The output of `Blind` is always random, but the output of the _protocol_, i.e., the signature itself, can be either random or deterministic.

chris-wood:
> One question I had is whether blind signatures that use techniques other than RSA would be able to implement this API?

Yeah, I think so. I could prototype blind BLS using this API as a sanity check. We don't have a BLS implementation yet, as far as I know, so that would have to be in a separate PR. Sounds like a TODO project for the next couple of weeks!

blindsign/blindsign.go
```diff
@@ -0,0 +1,32 @@
...
+// A blind signature protocol is a two-party protocol for computing
+// a digital signature. One party (the server) holds the signing key,
+// and the other (the client) holds the message input. Blindness
+// ensures that the server does not learn anything about the client's
+// input during the BlindSign step.
+package blindsign
+
+// A Verifier represents a specific instance of a blind signature verifier.
+type Verifier interface {
+	// Blind a message according to the Signer's public key, producing an
```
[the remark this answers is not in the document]
chris-wood: ACK, thanks. I'll update accordingly.

blindsign/blindrsa/rsa.go
```diff
@@ -0,0 +1,149 @@
+package blindrsa
```
armfazh, line 1: does this file came from go stdlib? if so, then add the license header.

blindsign/blindrsa/blindrsa.go
```diff
@@ -0,0 +1,213 @@
...
+	randReader := random
+	if randReader == nil {
+		randReader = rand.Reader
+	}
+	r, err := rand.Int(randReader, key.N)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	if r.Cmp(bigZero) == 0 {
```
armfazh, line 55: A cheaper way to test for zero `r.Sgn() == 0`

blindsign/blindrsa/blindrsa.go
```diff
@@ -0,0 +1,213 @@
...
+		r = bigOne
```
armfazh, line 56: `r.SetUint64(1)` this allows to remove the global `bigOne`, `BigZero` variables

armfazh (approved):

blindsign/blindrsa/rsa.go
chris-wood: Yeah, this came from the standard library.

blindsign/blindrsa/blindrsa.go
chris-wood: This is resolved by just calling `Sign` instead -- thanks!

blindsign/blindrsa/blindrsa_test.go
```diff
@@ -0,0 +1,279 @@
...
+
+	if sig1 == nil || sig2 == nil {
+		t.Fatal("nil signature output")
+	}
+	if !bytes.Equal(sig1, sig2) {
+		t.Fatal("deterministic signatures mismatched when they should be equal")
+	}
+}
+
+func TestRandomSignVerify(t *testing.T) {
```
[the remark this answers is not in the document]
chris-wood: Added!

blindsign/blindrsa/rsa.go
chris-wood: I added the license to each of the files.

chris-wood:
@cjpatton, @armfazh: I applied your suggestions. I think this PR is now ready for approval and merge. I'll squash before we merge.

chris-wood:
@armfazh is there a way to silence the linter that complains about use of `math/rand`? I'm using it to mock `crypto/rand`. Do we have a way to do this already?

blindsign/blindrsa/blindrsa.go
cjpatton: Oh gotcha.

cjpatton:
> @armfazh is there a way to silence the linter that complains about use of `math/rand`? I'm using it to mock `crypto/rand`. Do we have a way to do this already?

It looks like your goal is to just have a deterministic stream of bits for testing purposes. You can do this by implementing your own io.Reader, which is all that crypto/rand.Reader is: https://pkg.go.dev/crypto/rand#pkg-variables

chris-wood:
> It looks like your goal is to just have a deterministic stream of bits for testing purposes. You can do this by implementing your own io.Reader, which is all that crypto/rand.Reader is: https://pkg.go.dev/crypto/rand#pkg-variables

Oh, yes, duh. That is simpler. Pushed that change!

armfazh (approved):

blindsign/blindrsa/blindrsa_test.go
```diff
@@ -0,0 +1,320 @@
...
+	"encoding/json"
+	"io"
+	"io/ioutil"
+	"math/big"
+	"testing"
+)
+
+func TestSignVerify(t *testing.T) {
+	message := []byte("hello world")
+	key, err := rsa.GenerateKey(rand.Reader, 4096)
```
cjpatton, line 19: `go test` takes about 10 seconds on my machine. This test alone takes 1.8 seconds. If you knock this down from 4096 to 2048 bits, it takes just 0.1 seconds. I applaud your conservative choice of RSA params, but maybe we should go for the minimum recommended for tests :)

Another way to speed up the tests would be to just generate a single key pair once and hard-code it in this file.

Here and below.

blindsign/blindrsa/blindrsa.go
```diff
@@ -0,0 +1,213 @@
...
+func encodeMessageEMSAPSS(message []byte, key *rsa.PublicKey, hash hash.Hash, salt []byte) ([]byte, error) {
+	hash.Write(message)
+	digest := hash.Sum(nil)
+	hash.Reset()
+	emBits := key.N.BitLen() - 1
+	encodedMsg, err := emsaPSSEncode(digest[:], emBits, salt, hash)
+	return encodedMsg, err
+}
+
+func blind(random io.Reader, key *rsa.PublicKey) (*big.Int, *big.Int, error) {
```
cjpatton, line 45: I was confused by the purpose of this method because it uses "blind" as a noun, whereas the API call "Blind()" uses it as a verb. How about "generateBlindingFactor"?

blindsign/blindrsa/blindrsa.go
```diff
@@ -0,0 +1,213 @@
...
+	return RSAVerifier{
+		pk:   pk,
+		hash: hash,
+	}
+}
+
+func encodeMessageEMSAPSS(message []byte, key *rsa.PublicKey, hash hash.Hash, salt []byte) ([]byte, error) {
+	hash.Write(message)
+	digest := hash.Sum(nil)
+	hash.Reset()
```
cjpatton, line 39: nit: This assumes that the hash state was initialized before calling this method. While this assumption is valid with this code currently, in my experience with the hash.Hash API, this is a footgun. I would move this to the top of this function.

blindsign/blindrsa/blindrsa.go
```diff
@@ -0,0 +1,213 @@
...
+	}, nil
+}
+
+// Blind initializes the blind RSA protocol using an input message and source of randomness. If
+// randomness is not provided, the output signature is deterministic. Otherwise, the signature
+// includes a randomly generated PSS salt whose length equals the size of the underlying hash
+// function.
+//
+// See the specification for more details:
+// https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-rsa-blind-signatures-02#section-5.1.1
```
cjpatton, line 95:
```suggestion
// https://cfrg.github.io/draft-irtf-cfrg-blind-signatures/draft-irtf-cfrg-rsa-blind-signatures.html#section-5.2
```

blindsign/blindrsa/blindrsa.go
```diff
@@ -0,0 +1,213 @@
...
+		salt = nil
+	} else {
+		salt = make([]byte, v.hash.Size())
+		_, err := random.Read(salt)
+		if err != nil {
+			return nil, nil, err
+		}
+	}
+
+	r, rInv, err := blind(random, v.pk)
```
cjpatton, line 108: If `random == nil` then the blinding factor will be chosen deterministically (and thus predictable to the signer). This needs to be called out in the method's comment, as it would seem to violate the security property the API is supposed to provide.

It's one thing for the message encoding to be deterministic, it's another for the blinding factor to be deterministic.

blindsign/blindrsa/blindrsa_test.go
chris-wood: Chatted offline. We'll replace this with a hard-coded key pair!

blindsign/blindrsa/blindrsa.go
chris-wood: This is pointing to the published version, not the editor's copy.

blindsign/blindrsa/blindrsa_test.go
chris-wood: Done.

chris-wood:
@cjpatton suggestions applied, and the deterministic variant was dropped. Please let me know if more changes are needed!

cjpatton (approved):

PR #309, opened 2021-12-21
Verify the signatures for sanity's sake

chris-wood:
Because why not? :)


cjpatton (approved):
Nice, this validates one of the assertions made in the spec.

armfazh (approved):

PR #310, opened 2022-01-03
Pad blind RSA protocol messages to fixed modulus width

chris-wood:
Before, some messages would contain `|N|-1` bytes, and the spec requires the endpoints to abort in that case. This change ensures that all protocol messages are of the appropriate fixed width.

### DIFF
--- a/blindsign/blindrsa/blindrsa.go
+++ b/blindsign/blindrsa/blindrsa.go
@@ -78,7 +78,9 @@ func (v RSAVerifier) fixedBlind(message, salt []byte, r, rInv *big.Int) ([]byte,
 	z.Mul(z, x)
 	z.Mod(z, v.pk.N)
 
-	blindedMsg := z.Bytes()
+	kLen := (v.pk.N.BitLen() + 7) / 8
+	blindedMsg := make([]byte, kLen)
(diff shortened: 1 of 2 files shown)


blindsign/blindrsa/blindrsa.go
```diff
@@ -144,6 +147,7 @@ func verifyBlindSignature(pub *rsa.PublicKey, hashed, sig []byte) error {
 func (state RSAVerifierState) Finalize(data []byte) ([]byte, error) {
 	kLen := (state.verifier.pk.N.BitLen() + 7) / 8
 	if len(data) != kLen {
+		fmt.Println(len(data))
```
cjpatton, line 150: Presumably this is a debug line that you meant to delete?

cjpatton (commented):
This is pretty subtle, so it's worth adding a test that enforces it.

blindsign/blindrsa/blindrsa.go
chris-wood: Uh, yes, oops.

chris-wood:
@cjpatton added tests for the lengths!

cjpatton (approved):

armfazh (approved):

PR #311, opened 2022-01-20
Added implementation of FrodoKEM-640-SHAKE-CCA.

xvzcf:
Also tagging @dstebila

### DIFF
--- a/kem/frodo/frodo640shake/frodo.go
+++ b/kem/frodo/frodo640shake/frodo.go
@@ -15,6 +15,7 @@ const (
 	paramN             = 640
 	paramNbar          = 8
 	logQ               = 15
+	logQMask           = ((1 << logQ) - 1)
 	seedASize          = 16
 	pkHashSize         = 16
 	extractedBits      = 2
@@ -110,7 +111,7 @@ func newKeyFromSeed(seed []byte) (*PublicKey, *PrivateKey, error) {
 	if err != nil {
 		return nil, nil, err
 	}
-	mulAddASPlusE(&pk.matrixB, S[:], E[:], &A)
+	mulAddASPlusE(&pk.matrixB, &A, S[:], E[:])
 
 	// Populate the private key
 	copy(sk.hashInputIfDecapsFail[:], seed[0:SharedKeySize])
@@ -246,9 +247,9 @@ func (pk *PublicKey) EncapsulateTo(ct []byte, ss []byte, seed []byte) error {
 	if err != nil {
 		return err
 	}
-	mulAddSAPlusE(&Bp, Sp, Ep, &A)
+	mulAddSAPlusE(&Bp, Sp, &A, Ep)
 
-	mulAddSBPlusE(&V, &pk.matrixB, Sp, Epp)
+	mulAddSBPlusE(&V, Sp, &pk.matrixB, Epp)
 
 	// Encode mu, and compute C = V + enc(mu) (mod q)
 	encodeMessage(C[:], mu[:])
@@ -356,15 +357,15 @@ func (sk *PrivateKey) DecapsulateTo(ss, ct []byte) error {
 	if err != nil {
 		return err
 	}
-	mulAddSAPlusE(&BBp, Sp[:], Ep[:], &A)
+	mulAddSAPlusE(&BBp, Sp[:], &A, Ep[:])
 
 	// Reduce BBp modulo q
 	for i := range BBp {
-		BBp[i] = BBp[i] & ((1 << logQ) - 1)
+		BBp[i] = BBp[i] & logQMask
 	}
 
 	// compute W = Sp*B + Epp
-	mulAddSBPlusE(&W, &sk.pk.matrixB, Sp, Epp)
+	mulAddSBPlusE(&W, Sp, &sk.pk.matrixB, Epp)
 
 	// Encode mu, and compute CC = W + enc(mu') (mod q)
 	encodeMessage(CC[:], muprime[:])
@@ -506,7 +507,7 @@ func (sk *PrivateKey) Equal(other kem.PrivateKey) bool {
 		return false
 	}
 	return ctCompareU16(sk.matrixS[:], oth.matrixS[:]) == 0 &&
-		bytes.Equal(sk.hashInputIfDecapsFail[:], oth.hashInputIfDecapsFail[:]) &&
+		subtle.ConstantTimeCompare(sk.hashInputIfDecapsFail[:], oth.hashInputIfDecapsFail[:]) == 1 &&
 		sk.pk.Equal(oth.pk) &&
 		bytes.Equal(sk.hpk[:], oth.hpk[:])
 }
@@ -523,9 +524,8 @@ func (pk *PublicKey) Equal(other kem.PublicKey) bool {
 		return false
 	}
 
-	maskQ := uint16((1 << logQ) - 1)
 	for i := range pk.matrixB {
-		if (pk.matrixB[i] & maskQ) != (oth.matrixB[i] & maskQ) {
+		if (pk.matrixB[i] & logQMask) != (oth.matrixB[i] & logQMask) {
 			return false
 		}
 	}
--- a/kem/frodo/frodo640shake/frodo.go
+++ b/kem/frodo/frodo640shake/frodo.go
@@ -110,11 +110,11 @@ func newKeyFromSeed(seed []byte) (*PublicKey, *PrivateKey, error) {
 	}
 	sample(SE[:])
 
-	err = expandSeedIntoA(A[:], pk.seedA[:], shake128)
+	err = expandSeedIntoA(&A, &pk.seedA, &shake128)
 	if err != nil {
 		return nil, nil, err
 	}
-	mulAddASPlusE(pk.matrixB[:], S[:], E[:], A[:])
+	mulAddASPlusE(&pk.matrixB, S[:], E[:], &A)
 
 	// Populate the private key
 	copy(sk.hashInputIfDecapsFail[:], seed[0:SharedKeySize])
@@ -246,17 +246,17 @@ func (pk *PublicKey) EncapsulateTo(ct []byte, ss []byte, seed []byte) error {
 	}
 	sample(SpEpEpp[:])
 
-	err = expandSeedIntoA(A[:], pk.seedA[:], shake128)
+	err = expandSeedIntoA(&A, &pk.seedA, &shake128)
 	if err != nil {
 		return err
 	}
-	mulAddSAPlusE(Bp[:], Sp, Ep, A[:])
+	mulAddSAPlusE(&Bp, Sp, Ep, &A)
 
-	mulAddSBPlusE(V[:], pk.matrixB[:], Sp, Epp)
+	mulAddSBPlusE(&V, &pk.matrixB, Sp, Epp)
 
 	// Encode mu, and compute C = V + enc(mu) (mod q)
 	encodeMessage(C[:], mu[:])
-	add(C[:], V[:], C[:])
+	add(&C, &V, &C)
 
 	// Prepare the ciphertext
 	pack(ct[:matrixBpPackedSize], Bp[:])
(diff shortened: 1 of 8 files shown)


kem/frodo/frodo640shake/frodo.go
```diff
@@ -0,0 +1,628 @@
...
+		return false
+	}
+	if sk.pk == nil && oth.pk == nil {
+		return true
+	}
+	if sk.pk == nil || oth.pk == nil {
+		return false
+	}
+	for i := range sk.matrixS[:] {
+		if sk.matrixS[i] != oth.matrixS[i] {
```
bwesterb, line 514: Not constant time.

kem/frodo/frodo640shake/frodo.go
```diff
@@ -0,0 +1,628 @@
...
+
+	var SE [2 * paramN * paramNbar]uint16
+	var byteSE [2 * len(SE)]byte
+	S := SE[0 : paramN*paramNbar]
+	E := SE[paramN*paramNbar : 2*paramN*paramNbar]
+
+	var A [paramN * paramN]uint16
+
+	// Generate the secret value s, and the seed for S, E, and A. Add seedA to the public key
+	shake128 := xof.SHAKE128.New()
```
bwesterb, line 84: Instead of using the interface, it might be faster to use the shake128 package directly (as is done in Kyber for instance) as that can help prevent allocations due to making it easier for Go's escape analysis.

kem/frodo/frodo640shake/matrix_shake.go
```diff
@@ -0,0 +1,64 @@
...
+				sum += A[i*paramN+j] * s[k*paramN+j]
+			}
+			// Adding e. No need to reduce modulo 2^15, extra bits are taken
+			// care of during packing later on.
+			out[i*paramNbar+k] += sum
+		}
+	}
+}
+
+func mulAddSAPlusE(out []uint16, s []uint16, e []uint16, A []uint16) {
```
bwesterb, line 50: It is probably worthwhile to use `*[whatever]uint16` instead of `[]uint16` where possible.

kem/frodo/kat_test.go
```diff
@@ -0,0 +1,77 @@
...
+	"github.com/cloudflare/circl/internal/nist"
+	"github.com/cloudflare/circl/kem/schemes"
+)
+
+func TestPQCgenKATKem(t *testing.T) {
+	kats := []struct {
+		name string
+		want string
+	}{
+		// Computed from testdata/kat.rsp, which in turn is taken from:
```
bwesterb, line 21: Then why do we include `kat.rsp` itself still? Maybe leave `kat.rsp` out?

bwesterb (commented):
Thanks Goutam! Looks good on first pass. Haven't checked the nitty-gritty details (yet).

kem/frodo/kat_test.go
```diff
@@ -0,0 +1,77 @@
...
+	"github.com/cloudflare/circl/kem/schemes"
+)
+
+func TestPQCgenKATKem(t *testing.T) {
+	kats := []struct {
+		name string
+		want string
+	}{
+		// Computed from:
+		// https://github.com/microsoft/PQCrypto-LWEKE/blob/master/KAT/PQCkemKAT_19888_shake.rsp
```
armfazh, line 22: replace master by the commit number

kem/frodo/frodo640shake/matrix_shake.go
```diff
@@ -0,0 +1,64 @@
...
```
claucece, line 50: While this is following more the 'C-like' syntax, you can always return the parameter instead of passing it as argument...

claucece (commented):
Looks good! As this doesn't have the proofs yet, I'm leaving some details out that will be caught by the proofs.

kem/frodo/frodo640shake/matrix_shake.go
bwesterb: Returning slices might lead to allocations that can be prevented like this.

bwesterb: (By passing an "out" slice, you make the caller responsible for the allocation.)

kem/frodo/frodo640shake/matrix_shake.go
xvzcf: Made the changes wherever I could, let me know if I missed a spot.

bwesterb: Thanks!

kem/frodo/frodo640shake/frodo.go
```diff
@@ -0,0 +1,623 @@
...
+		return false
+	}
+	if sk.pk == nil && oth.pk == nil {
+		return true
+	}
+	if sk.pk == nil || oth.pk == nil {
+		return false
+	}
+	return ctCompareU16(sk.matrixS[:], oth.matrixS[:]) == 0 &&
+		bytes.Equal(sk.hashInputIfDecapsFail[:], oth.hashInputIfDecapsFail[:]) &&
```
armfazh, line 513: better to use ConstantTimeCompare
 https://pkg.go.dev/crypto/subtle#ConstantTimeCompare

kem/frodo/frodo640shake/frodo.go
```diff
@@ -0,0 +1,623 @@
...
+func (pk *PublicKey) Pack(buf []byte) {
+	if len(buf) != PublicKeySize {
+		panic("buf must be of length PublicKeySize")
+	}
+
+	copy(buf[:seedASize], pk.seedA[:])
+	pack(buf[seedASize:], pk.matrixB[:])
+}
+
+// TODO: Unpacks pk from buf.
```
armfazh, line 463: is this TODO need to be resolved?

kem/frodo/frodo640shake/matrix_shake.go
```diff
@@ -0,0 +1,64 @@
...
+		}
+
+		for j := 0; j < paramN; j++ {
+			A[(i*paramN)+j] = uint16(ARow[j*2]) | (uint16(ARow[(j*2)+1]) << 8)
+		}
+	}
+	return nil
+}
+
+func mulAddASPlusE(out *[paramN * paramNbar]uint16, s []uint16, e []uint16, A *[paramN * paramN]uint16) {
```
armfazh, line 34: we can sort the parameters in order to differentiate the left-side vs the right-side multiplication
 mulAddASPlusE( out, A, S, E )
 mulAddSAPlusE( out, S, A, E )

kem/frodo/frodo640shake/matrix_shake.go
```diff
@@ -0,0 +1,64 @@
...
+		if err != nil {
+			return err
+		}
+		_, err = xof.Read(ARow[:])
+		if err != nil {
+			return err
+		}
+
+		for j := 0; j < paramN; j++ {
+			A[(i*paramN)+j] = uint16(ARow[j*2]) | (uint16(ARow[(j*2)+1]) << 8)
```
armfazh, line 28: here does it need reduction? if not, please add a comment why is not the case.

kem/frodo/frodo640shake/util.go
```diff
@@ -0,0 +1,151 @@
...
+}
+
+func mulBS(out *[paramNbar * paramNbar]uint16, b *[paramNbar * paramN]uint16, s *[paramN * paramNbar]uint16) {
+	for i := 0; i < paramNbar; i++ {
+		for j := 0; j < paramNbar; j++ {
+			out[i*paramNbar+j] = 0
+			for k := 0; k < paramN; k++ {
+				out[i*paramNbar+j] += b[i*paramN+k] * s[j*paramN+k]
+			}
+			out[i*paramNbar+j] = out[i*paramNbar+j] & ((1 << logQ) - 1)
```
armfazh, line 136: can we promote this value as a constant 
((1 << logQ) - 1)

kem/frodo/frodo640shake/util.go
```diff
@@ -0,0 +1,151 @@
...
+			temp := (msg[msgPos] & maskQ) + (1 << (logQ - extractedBits - 1))
+			temp >>= (logQ - extractedBits)
+			temp &= maskExtractedBits
+			out[i] |= byte(temp) << (j * extractedBits)
+			msgPos++
+		}
+	}
+}
+
+func mulAddSBPlusE(out *[paramNbar * paramNbar]uint16, b *[paramN * paramNbar]uint16, s []uint16, e []uint16) {
```
armfazh, line 113: suggestion: we can make types to help on passing parameters correctly.
type nb_nb [paramNbar * paramNbar]uint16

must distinguish when cardinality coincides, e.g. 
type n_nb  [paramN * paramNbar]uint16
type nb_n  [paramNbar * paramN]uint16

kem/frodo/frodo640shake/noise.go
```diff
@@ -0,0 +1,18 @@
...
+func sample(sampled []uint16) {
+	for i := 0; i < len(sampled); i++ {
+		var gaussianSample uint16 = 0
+		sign := sampled[i] & 1
+		unifSample := sampled[i] >> 1
+
+		for j := 0; j < CDFTableLen-1; j++ {
+			gaussianSample += (CDFTable[j] - unifSample) >> 15
+		}
+		sampled[i] = ((-sign) ^ gaussianSample) + sign
```
armfazh, line 16: when sign=1, sampled gets `sampled = (0xFF..FF ^ gaussianSample) + 1`
is that final `plus one` part of the result?

armfazh (changes requested):
some minor changes

bwesterb (approved):
Thanks for this PR and your previous changes.

I’ve gone through the code in detail on a plane. I would like some more documentation on what is going on and there is some low hanging fruit for optimization. I don’t think they’re blockers. I didn’t have internet on the plane, so I’m sorry that I’ll have to leave my comments as one big list.

- note in comment that mbar is nbar
- Note in comment that extractedbits is B
- Reading/writing to shake never errors so we don’t need to check. Use _, _ = h.Read to prevent compiler warning.
- Keygen
    - Note in comment that it’s S transpose, not S
    - Why is the seed as it is? Doesn’t seem to be defined in the FrodoKEM spec. Also, I’d say 128 bits should be enough.
    - The seed doesn’t contain seed_A (instead it contains z, with seed_A=shake(z)) so it doesn’t make sense for len_seedA to be part of the definition of KeySeedSize
    - You can remove the copy for shakeInputForSE by first writing a single byte and then the remainder.
    - It would be nice to add a test that the CDFTable indeed corresponds to Table 3 in the spec.
    - Although it’s all pretty standard, it wouldn’t hurt to add a comment or two what is going on in sample().
    - Please add a comment to note how you make a 2-dimensional array 1-dimensional. (Ie. Rows after each other.)
    - You’re not reducing in expandSeedIntoA. Probably that doesn’t matter but it deserves a comment that you’re leaving that out on purpose.
    - If you want you can get rid of some extra copy()s. For instance, you can sample directly into sk.matrixS
- Encapsulation
    - This is probably premature optimization (— no need to look into it), but it might be faster to read one block of shake at a time and then call sample() on it before doing the next shake block. That reduces pressure on caches. Same goes for keygen.
- Decapsulation
    - There is some shared code between encaps and decaps (generation of S’ and E’). Perhaps move that into a function?
    - Check whether Go turns ((1 << logQ)-1) into a constant or recomputes it.
- Pack: This could be made much more efficient by packing eight at a time. Not a blocker.
- Encode/DecodeMessage: we know the size of msg in advance. The go compiler is too stupid to realize so better make it see by using constant-size array.
-

kem/frodo/frodo640shake/noise.go
xvzcf: Yes, since `flipBits(a) + 1 ≡ -a (mod 2^16)` if a is 16 bit number,  this just flips the sign of `gaussianSample` if `sign = 1`. Should I add a comment here?

armfazh (approved):

bwesterb:
Thanks @xvzcf !

PR #313, opened 2022-01-26
[oprf] Publicize oprf.Server.PrivateKey

afjoseph:
Hey. Thanks for the fantastic repo!

This change is minimal and there might be an API I missed that does exactly this, so please let me know if I missed it.

This change allows protocols like [OPAQUE](https://datatracker.ietf.org/doc/html/draft-krawczyk-cfrg-opaque-06) to serialize and save the private key of a server, so an `oprf.Server` instance can be recreated easily at a later step.

More concretely in OPAQUE, the password registration step requires a new `kU` parameter - which is the `oprf.Server`'s private key, created [here in oprf.NewServer()](https://github.com/afjoseph/circl/blob/c9f04ddc2908c0995f6122f2b4c3ff7a0298bc33/oprf/server.go#L36) - to be saved in [the last step](https://datatracker.ietf.org/doc/html/draft-krawczyk-cfrg-opaque-06#section-3.1) of the password registration procedure, and later retrieved during password authentication. This can be done after this PR by using the same saved private key inside `oprf.NewServer`.

Cheers and many thanks. Lemme know if a more graceful solution is required (i.e., exposing an `oprf.Server.[De]SerializeKey` function)

### DIFF
--- a/oprf/server.go
+++ b/oprf/server.go
@@ -12,7 +12,7 @@ import (
 // Server is a representation of a OPRF server during protocol execution.
 type Server struct {
 	suite
-	privateKey PrivateKey
+	PrivateKey PrivateKey
 }
 
 // NewServer creates a Server in base mode, and generates a key if no skS is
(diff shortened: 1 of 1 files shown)


armfazh:
There's another way to proceed without exposing the field.
1. Create a private key calling either `GenerateKey` or `DeriveKey`. This will return a PrivateKey object.
2. The private key can stored using the serialization methods provided.
3. The private key can be used for creating a new Server calling either `NewServer` or `NewVerifiableServer`.

Let me know if this fulfills your use case.

afjoseph:
Hey @armfazh. Thank you for your reply. Your solution suffices and works perfect. Closing this ticket :)

PR #315, opened 2022-02-01
Performs modular reduction for all scalars.

### LINKED ISSUE #312: Inadequate scalar reduction in p384 leads to panic
https://github.com/cloudflare/circl/blob/15f39042039a5d953aa3d11650576df0fcd0ecfd/ecc/p384/p384opt.go#L55

`reduceScalar` is supposed to "shorten a scalar modulo the order of the curve" but it doesn't do that for all inputs and this can lead to a panic in wnaf.go.

PoC:

```go
package main

import (
    "github.com/cloudflare/circl/ecc/p384"
    "math/big"
)

func main() {
    a_x, _ := new(big.Int).SetString("2197194086093689212280862101867524553402030273258457080984175289262818302155461584526794911762464589043284585654550", 10)
    a_y, _ := new(big.Int).SetString("27340319491623798325691320370638865713982009585794110374349453846146179674941474386159930807215381526124834607054138", 10)
    b, _ := new(big.Int).SetString("39402006196394479212279040100143613805079739270465446667946905279627659399113263569398956308152294913554433653942684", 10)

    curve := p384.P384()
    curve.ScalarMult(a_x, a_y, b.Bytes())
}
```

armfazh:
Scalars are reduced only when its size in bytes is bigger that the order=N.
There are, however, scalars of such size but larger than N that escaped to be reduced.
With this change, now scalars are fully reduced.

Fixes #312

### DIFF
--- a/ecc/p384/opt_test.go
+++ b/ecc/p384/opt_test.go
@@ -4,7 +4,9 @@
 package p384
 
 import (
+	"bytes"
 	"crypto/elliptic"
+	"crypto/rand"
 	"math/big"
 	"testing"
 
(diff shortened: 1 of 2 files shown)


ecc/p384/opt_test.go
```diff
@@ -4,14 +4,43 @@
...
+		for i := 0; i < sizeFp; i++ {
+			buffer[i] = 0xFF
+		}
+		_, err := rand.Read(buffer[sizeFp:])
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		for i := sizeFp; i < 3*sizeFp; i++ {
+			k := buffer[:i]
```
chris-wood, line 31: Do we also want to test the case where `len(k) < sizeFp`? It looks like this just tests the cases where `len(k) >= sizeFp`.

chris-wood (approved):

chris-wood:
The change looks good! I wonder if we can expand on the tests, though.

ecc/p384/opt_test.go
armfazh: typo, fixing now.

ecc/p384/opt_test.go
```diff
@@ -4,14 +4,43 @@
...
+
+		for i := 0; i < sizeFp; i++ {
+			buffer[i] = 0xFF
+		}
+		_, err := rand.Read(buffer[sizeFp:])
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		for i := sizeFp; i < 3*sizeFp; i++ {
```
armfazh, line 30:
```suggestion
		for i := 0; i < 3*sizeFp; i++ {
```

bwesterb (approved):

PR #318, opened 2022-02-27
[oprf] Publicize ClientRequest fields

afjoseph:
Hey. Thanks for the fantastic project!

I'm working on a client-server project where the client code is javascript. I've compiled CIRCL from Go to javascript (and wasm) using GopherJS and Go's internal tooling successfully.

The "bridge" code that handles serializations to-and-fro Go and JS need to (de)serialize `oprf.ClientRequest`. Serializers (JSON, YAML, etc.) can't do that if the fields in oprf.ClientRequest are private.

This PR publicizes those fields.

An alternative method would be to add a `MarshalJSON()` inside `oprf` package. I like this solution more, but I don't know if you will allow it: I didn't see similar serialization methods.

A third approach is to serialize the whole struct into one big byte slice, concatenated one after the other. This works too. Lemme know which works best for your project and I'll take care of it.

Many thanks :)

### DIFF
--- a/oprf/client.go
+++ b/oprf/client.go
@@ -16,17 +16,17 @@ type Client struct {
 
 // ClientRequest is a structure to encapsulate the output of a Request call.
 type ClientRequest struct {
-	inputs   [][]byte
-	blinds   []Blind
-	elements []group.Element
+	Inputs   [][]byte
+	Blinds   []Blind
+	Elements []group.Element
 }
 
 // BlindedElements returns the serialized blinded elements produced for the client request.
 func (r ClientRequest) BlindedElements() [][]byte {
 	var err error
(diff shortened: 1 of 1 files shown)


armfazh:
@afjoseph you raised a very good point, definitely the package needs serialization methods for structures that travel through network.

But there are some news:
The OPRF package will be upgraded to be compliant with draft v09, see #319 . So I recommend you wait for the new version to be merged.

This new version includes serialization methods for `EvaluationRequest` and `Evaluation` structures, so everything dumps to a slice of bytes.
Feel free to test the new changes, and let me know if something else is needed.

afjoseph:
Hey @armfazh. Thanks for the fantastic work. I'll close my PR.

PR #319, opened 2022-03-01
oprf: update API and test vectors to draft v09

### LINKED ISSUE #317: Update OPRF package to latest version of the draft
It's currently at -08. Let's move to -09.

armfazh:
oprf: update API and test vectors to [draft v09](https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-voprf-09)

The `oprf` package now supports three modes OPRF, VOPRF and POPRF.

Also adds a package for DLEQ proofs, which can be of particular interest.

Closes #317


group/dleq/dleq.go
```diff
@@ -0,0 +1,260 @@
...
+)
+
+const (
+	labelSeed         = "Seed-"
+	labelChallenge    = "Challenge"
+	labelComposite    = "Composite"
+	labelHashToScalar = "HashToScalar-"
+)
+
+type Params struct {
```
chris-wood, line 20: Maybe call this type `Prover`, and the corresponding function `Prove`? That way, the usage would look like:

```
prover := dleq.Prover{...}
proof, err := prover.Prove(k, A, kA, b, kB, rnd)
```

chris-wood (changes requested):
I have two blocking comments: 

1. The server's `Evaluate` interface, which currently accepts a blinded input and info string, which feels not quite right.
2. The `AddInfo` function is currently exported when it probably doesn't need to be.

group/dleq/dleq.go
armfazh: done

oprf/client.go
```diff
@@ -58,150 +41,161 @@ func NewVerifiableClient(id SuiteID, pkS *PublicKey) (*Client, error) {
...
+	finData := &FinalizeData{inputs, blinds, evalReq, nil}
+	return finData, evalReq, nil
 }
 
-func (c *Client) blindMultiplicative(blindedElt []group.Element, inputs [][]byte, blinds []Blind) {
-	for i := range inputs {
-		p := c.suite.Group.HashToElement(inputs[i], c.suite.getDST(hashToGroupDST))
-		blindedElt[i] = c.suite.Group.NewElement()
-		blindedElt[i].Mul(p, blinds[i])
+func (c *Client) AddInfo(r *FinalizeData, info []byte) error {
```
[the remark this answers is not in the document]
armfazh: changed API, now only PartialOblivious can add information

oprf/server.go
```diff
@@ -47,76 +37,93 @@ func newServer(id SuiteID, m Mode, skS *PrivateKey) (*Server, error) {
...
 			return nil, err
 		}
 	}
 
-	return &Evaluation{out, proof}, nil
+	return &Evaluation{evaluations, proof}, nil
+}
+
+// Evaluate evaluates a set of blinded inputs from the client.
+func (s *Server) Evaluate(req *EvaluationRequest, info []byte) (*Evaluation, error) {
```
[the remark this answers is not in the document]
armfazh: That's true, I changed the API. Now, only PartialOblivious can add information

oprf/oprf.go
```diff
@@ -160,94 +236,125 @@ func mustWrite(h io.Writer, bytes []byte) {
...
+}
 
-	dst := s.getDST(seedDST)
-	binary.BigEndian.PutUint16(lenBuf, uint16(len(dst)))
-	mustWrite(H, lenBuf)
-	mustWrite(H, dst)
+// EvaluationRequest contains the blinded elements to be evaluated by the Server.
+type EvaluationRequest struct {
+	g        group.Group
+	elements []blinded
```
chris-wood, line 267: I think this should be exported so that applications can encode these as necessary. Right now, users of this API _have_ to use the `MarshalBinary` interface below, which might not be what every application wants or needs.

armfazh: internals of protocol messages are public.
Similar to what is requested in #318

chris-wood: Excellent!

chris-wood (approved):
Thanks for the updates! From an API perspective, I think this is workable for most (all?) use cases that I can think of. I didn't carefully review the implementation of each mode, but trust that the test vectors have us covered.

PR #320, opened 2022-03-01
Expose randomly generated values for Blind RSA (salt and blind)

chris-wood:
This is useful when generating test vectors for applications using this package.

### DIFF
--- a/blindsign/blindrsa/blindrsa.go
+++ b/blindsign/blindrsa/blindrsa.go
@@ -192,7 +192,9 @@ func (state RSAVerifierState) CopyBlind() []byte {
 
 // CopySalt returns an encoding of the per-message salt used in the protocol.
 func (state RSAVerifierState) CopySalt() []byte {
-	return state.salt
+	salt := make([]byte, len(state.salt))
+	copy(salt, state.salt)
+	return salt
 }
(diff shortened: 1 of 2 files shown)


blindsign/blindrsa/blindrsa.go
```diff
@@ -165,6 +184,17 @@ func (state RSAVerifierState) Finalize(data []byte) ([]byte, error) {
...
 
+// CopyBlind returns an encoding of the blind value used in the protocol.
+func (state RSAVerifierState) CopyBlind() []byte {
+	r := new(big.Int).ModInverse(state.rInv, state.verifier.pk.N)
+	return r.Bytes()
+}
+
+// CopySalt returns an encoding of the per-message salt used in the protocol.
+func (state RSAVerifierState) CopySalt() []byte {
+	return state.salt
```
armfazh, line 195: it needs to do an actual copy, otherwise `state.salt` can be modified externally.

armfazh (approved):

blindsign/blindrsa/blindrsa.go
chris-wood: Good catch -- fixed!

PR #322, opened 2022-03-04
Enable Ristretto255 for OPRFs.

armfazh:
oprf.SuiteR255 represents the OPRF with Ristretto255 and SHA-512.

### DIFF
--- a/group/ristretto255.go
+++ b/group/ristretto255.go
@@ -98,16 +98,31 @@ func (g ristrettoGroup) HashToElementNonUniform(b, dst []byte) Element {
 }
 
 func (g ristrettoGroup) HashToElement(msg, dst []byte) Element {
+	// Compliaint with draft-irtf-cfrg-hash-to-curve.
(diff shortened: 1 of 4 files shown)


group/ristretto255.go
```diff
@@ -97,16 +97,24 @@ func (g ristrettoGroup) HashToElementNonUniform(b, dst []byte) Element {
 	return g.HashToElement(b, dst)
 }
 func (g ristrettoGroup) HashToElement(msg, dst []byte) Element {
```
bwesterb, line 99: To double-check: there are no existing users of Ristretto255 that we're breaking now?

bwesterb (approved):

group/ristretto255.go
armfazh: With this PR, oprf is the only package in CIRCL that uses ristretto
hard to tell with other external packages.

chris-wood (approved):

PR #323, opened 2022-03-04
Export Group and Hash for each OPRF suite

chris-wood:
This is needed so clients can go from `suite` to `group` for serialization functions. Otherwise, clients have to know that, e.g., `oprf.SuiteP384` corresponds to `group.P384`. Now, clients can access the group by `oprf.SuiteP384.Group`.

### DIFF
--- a/oprf/oprf.go
+++ b/oprf/oprf.go
@@ -86,11 +86,11 @@ type Suite interface{ cannotBeImplementedExternally() }
 
 var (
 	// SuiteP256 represents the OPRF with P-256 and SHA-256.
-	SuiteP256 Suite = params{ID: 3, g: group.P256, h: crypto.SHA256}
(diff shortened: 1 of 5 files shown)


oprf/oprf.go
```diff
@@ -164,15 +164,15 @@ func NewPartialObliviousServer(s Suite, key *PrivateKey) PartialObliviousServer
...
 
 type params struct {
-	ID uint16
-	m  Mode
-	g  group.Group
-	h  crypto.Hash
+	ID    uint16
+	m     Mode
+	Group group.Group
+	Hash  crypto.Hash
```
armfazh, line 170:
```suggestion
	group.Group
	crypto.Hash
```

armfazh (approved):

PR #324, opened 2022-03-04
Add deterministic Blind function 

chris-wood:
... along with function to copy an evaluation requests's blind values.

### DIFF
--- a/oprf/client.go
+++ b/oprf/client.go
@@ -28,7 +28,7 @@ func (c client) Blind(inputs [][]byte) (*FinalizeData, *EvaluationRequest, error
 		return nil, nil, ErrInvalidInput
 	}
 
-	blinds := make([]blind, len(inputs))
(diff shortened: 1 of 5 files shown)


oprf/client.go
```diff
@@ -36,6 +36,26 @@ func (c client) Blind(inputs [][]byte) (*FinalizeData, *EvaluationRequest, error
 	return c.blind(inputs, blinds)
 }
 
+func (c client) DeterministicBlind(inputs, blinds [][]byte) (*FinalizeData, *EvaluationRequest, error) {
```
armfazh, line 39: If we are going to expose blinds, I would rather prefer
Let do these changes:
```patch
- blind
- DeterministicBlind(inputs, blinds [][]byte)
+ Blind
+ DeterministicBlind(inputs, blinds []Blind)
```

armfazh (approved):

PR #326, opened 2022-03-04
group: adds Set/Copy methods for group elements and scalars

armfazh:


group/ristretto255.go
```diff
@@ -121,6 +121,15 @@ func (e *ristrettoElement) IsEqual(x Element) bool {
 	return e.p.Equals(&x.(*ristrettoElement).p)
 }
 
+func (e *ristrettoElement) Set(x Element) Element {
+	e.p.Set(&x.(*ristrettoElement).p)
+	return e
+}
+
+func (e *ristrettoElement) Copy() Element {
+	return &ristrettoElement{*new(r255.Point).Set(&e.p)}
```
lukevalenta, line 130: Should there be a separate `Point.Copy()` function? Otherwise, couldn't the copied `ristrettoElement` share the same underlying memory depending on the `Point` implementation?

chris-wood (approved):
LGTM pending resolution to @lukevalenta's question.

group/ristretto255.go
bwesterb: The definition of `ristrettoElement` is

```
type ristrettoElement struct {                                                   
    p r255.Point                                                                 
}
```

and `r255.Point` is also a struct.

Hence
```
func (e *ristrettoElement) Copy() Element {
    ret := *e
    return &ret
}
```

will suffice.

bwesterb: (I think we'll end up with similar assembler, so it's more a preference of style.)

bwesterb (approved):

lukevalenta (approved):

PR #327, opened 2022-03-15
Bumping up to Go1.18

armfazh:
Enable CI tests with Go1.18 and moves compatibility pointer to Go1.16


armfazh:
Well, the arm64 golang image is not ready yet. Let's wait if it is released soon.
https://hub.docker.com/r/arm64v8/golang

bwesterb (approved):

PR #328, opened 2022-03-16
Refactoring FrodoKEM implementation.

xvzcf:
Addressed more review comments.


kem/frodo/frodo640shake/frodo.go
```diff
@@ -1,4 +1,5 @@
 // Package frodo640shake implements the variant FrodoKEM-640 with SHAKE.
+// Multi-dimensional arrays are stored in 1-dimensional arrays in row-major order.
```
bwesterb, line 2: Better to move that down to the `type` block where those multi-dimensional types are defined.

bwesterb (approved):
This is great, thanks.

armfazh (approved):

Issue #329, opened 2022-03-16
TestScalarMult/wrong_P is failing with Go 1.18


GlenPickle:
Per https://go.dev/doc/go1.18: "Operating on invalid curve points (those for which the IsOnCurve method returns false, and which are never returned by [Unmarshal](https://go.dev/pkg/crypto/elliptic#Unmarshal) or a Curve method operating on a valid point) has always been undefined behavior, can lead to key recovery attacks, and is now unsupported by the new backend."

This change means that TestScalarMult/wrong_P will always fail, and should be removed.

armfazh:
Have you take a look at #327 
I rewrote that test to one that is equivalent.

PR #330, opened 2022-03-16
Remove test that is no longer supported with Go 1.18

GlenPickle:
Closes https://github.com/cloudflare/circl/issues/329

Removing this test will resolve the failing test and allow this package to build successfully again in Ubuntu.

If you would prefer to use build tags to have this test still run with Go 1.17 but not with Go 1.18 I can make that change, just let me know.


armfazh:
Solved in #327

PR #331, opened 2022-05-03
Update with latest RSA from go stdlib

armfazh:
Updates with latest RSA from go stdlib.
Also removes some code not used by blindRSA.


blindsign/blindrsa/rsa.go
```diff
@@ -87,12 +83,17 @@ func encrypt(c *big.Int, pub *rsa.PublicKey, m *big.Int) *big.Int {
...
 	if c.Cmp(priv.N) > 0 {
 		err = rsa.ErrDecryption
 		return
 	}
+	if priv.N.Sign() == 0 {
+		return nil, rsa.ErrDecryption
+	}
 
 	var ir *big.Int
 	if random != nil {
```
lukevalenta, line 99: Are we missing this line? https://cs.opensource.google/go/go/+/master:src/crypto/rsa/rsa.go;l=558

lukevalenta (approved):
One comment about the omitted `randutil.MaybeReadByte`, but from what I can tell that's non-essential and only meant to stop callers from relying on non-guaranteed behavior. Otherwise, lgtm.

blindsign/blindrsa/rsa.go
bwesterb: Doesn't seem too important.

bwesterb (approved):

blindsign/blindrsa/rsa.go
armfazh: it shuffles the random source in case a user-provided source is provided.
in this package, the random source is always `crypto/rand`, which it's assumed to be secure.

PR #333, opened 2022-05-03
Bump golangci-lint to v1.45 and format files with gofumpt

armfazh:
gofumpt is enabled by default and some files required some easy fixes.


lukevalenta (approved):
lgtm!

PR #343, opened 2022-06-28
kem: Allows registering and querying by CurveID

bwesterb:


kem/hybrid/hybrid.go
```diff
@@ -56,18 +56,21 @@ var kyber512X kem.Scheme = &scheme{
 	"Kyber512-X25519",
 	kyber512.Scheme(),
 	hpke.KEM_X25519_HKDF_SHA256.Scheme(),
+	0x01fd,
```
bwesterb, line 59: These are temporary and we might not want to include them, yet.

armfazh: from where these numbers came from, please add a link to a reference.

kem/tls/tls.go
```diff
@@ -0,0 +1,30 @@
+// Package tls adds helpers to integrate the KEMs supported in Circl into TLS.
```
armfazh, line 1:
```suggestion
// Package tls adds helpers to integrate the KEMs supported in CIRCL into TLS.
```

armfazh:
Looks good, but CI is not happy.

kem/tls/tls.go
```diff
@@ -0,0 +1,30 @@
...
+	allSchemesByCurveID = make(map[uint]kem.Scheme)
+	for _, scheme := range schemes.All() {
+		if tlsScheme, ok := scheme.(TLSScheme); ok {
+			allSchemesByCurveID[tlsScheme.TLSCurveID()] = scheme
+		}
+	}
+}
+
+// Returns scheme with the given TLS CurveID, if any.
+func SchemeByCurveID(id uint) kem.Scheme { return allSchemesByCurveID[id] }
```
armfazh, line 23: this function may be moved to `/circl/kem/schemes` which is the place where the register lives, doing so we don't need the `/circl/kem/tls` package, which may be confused with the stdlib package.

bwesterb: I used a separate package to be consistent with the pki package, that provides similar functionality but then querying signature schemes by OID. If you’re ok being inconsistent with that, I’ll move it.

armfazh: the issue I'm pointing is that when importing `/circl/kem/tls`, it looks like is a replacement of the `tls` package. what about renaming to `tls_helpers`?
or putting these functions in the `pki` package.

armfazh:
@bwesterb  Let's rebase this PR if it's still needed.
CurveIDs may have changed since then.
cc: @Lekensteyn

bwesterb:
Yeah, let's just start fresh with this if we need it.

bwesterb:
(Closing for now.)

PR #348, opened 2022-07-12
group: Implements Shamir and Feldman secret sharing.

armfazh:
New:
 - Introduces Lagrange interpolation [1].
 - Shamir's secret sharing [2] on top of the group interface.
 - Feldman's verifiable secret sharing [3] on top of the group interface.

References:
 [1] https://mathworld.wolfram.com/LagrangeInterpolatingPolynomial.html
 [2] https://dl.acm.org/doi/10.1145/359168.359176
 [3] https://ieeexplore.ieee.org/document/4568297


group/secretsharing/ss.go
```diff
@@ -0,0 +1,114 @@
...
+
+	"github.com/cloudflare/circl/group"
+)
+
+type SecretShare struct {
+	ID    uint
+	Share group.Scalar
+}
+
+type ShamirSS struct {
```
chris-wood, line 16:
```suggestion
type ShamirSecretSplitter struct {
```

I would prefer if exported names were not abbreviated. I don't know if `Splitter` is the best name for this thing, but something like that?

group/secretsharing/ss.go
```diff
@@ -0,0 +1,114 @@
...
+		id := uint(i + 1)
+		x.SetUint64(uint64(id))
+		shares[i].ID = id
+		shares[i].Share = poly.evaluate(x)
+	}
+
+	return shares
+}
+
+func (s ShamirSS) ShardSecret(rnd io.Reader, secret group.Scalar) []SecretShare {
```
chris-wood, line 48: Maybe just `Shard`?

group/secretsharing/ss.go
```diff
@@ -0,0 +1,114 @@
...
+	return s.generateShares(s.polyFromSecret(rnd, secret))
+}
+
+func (s ShamirSS) RecoverSecret(shares []SecretShare) (group.Scalar, error) {
```
chris-wood, line 52: Maybe this can be `Unshard`?

group/secretsharing/ss.go
```diff
@@ -0,0 +1,114 @@
...
+		x[i].SetUint64(uint64(shares[i].ID))
+		px[i] = shares[i].Share
+	}
+
+	return LagrangeInterpolate(s.G, x, px)
+}
+
+type Commitment = group.Element
+
+type FeldmanSS struct {
```
chris-wood, line 72:
```suggestion
type VerifiableSecretSplitter struct {
```

I don't think we need to expose the implementation detail in the name.

group/secretsharing/ss_test.go
```diff
@@ -0,0 +1,119 @@
...
+	vs, err := secretsharing.NewVerifiable(g, t, n)
+	test.CheckNoErr(tt, err, "failed to create ShamirSS")
+
+	want := g.RandomScalar(rand.Reader)
+	shares, com := vs.ShardSecret(rand.Reader, want)
+	test.CheckOk(len(shares) == int(n), "bad num shares", tt)
+	test.CheckOk(len(com) == int(t+1), "bad num commitments", tt)
+
+	for i := range shares {
+		test.CheckOk(shares[i].Verify(g, com), "failed one share", tt)
```
chris-wood, line 53: We should also add a test where we modify one of the shares to force this check to fail.

chris-wood (changes requested):
The ergonomics of the API seem good, but I think we can do better with the naming. I'm curious to hear what @cjpatton has to say about the names.

group/secretsharing/ss.go
armfazh: I opted by an unexported struct called `secretSharing`.

group/secretsharing/ss.go
armfazh: done

group/secretsharing/ss.go
armfazh: In several sources the common term is `Recover`

group/secretsharing/ss.go
armfazh: I opted by `VerifiableSecretSharing` to be compatible with the non-verifiable version

group/secretsharing/ss_test.go
armfazh: done

group/secretsharing/poly.go
```diff
@@ -0,0 +1,73 @@
...
+import (
+	"errors"
+	"io"
+
+	"github.com/cloudflare/circl/group"
+)
+
+type polynomial struct {
+	deg   uint64
+	coeff []group.Scalar
```
bwesterb, line 12: Document that this represents the polynomial  `\sum sum_i  coeff[i] x^i`. (You might've flipped the order.)

group/secretsharing/poly.go
```diff
@@ -0,0 +1,73 @@
...
+	px := p.coeff[p.deg].Copy()
+	for i := int(p.deg) - 1; i >= 0; i-- {
+		px.Mul(px, x)
+		px.Add(px, p.coeff[i])
+	}
+	return px
+}
+
+func LagrangeCoefficient(g group.Group, x []group.Scalar, index uint64) group.Scalar {
+	if index > uint64(len(x)) {
```
bwesterb, line 34: `>=`?

group/secretsharing/poly.go
```diff
@@ -0,0 +1,73 @@
...
+
+func LagrangeInterpolate(g group.Group, x, px []group.Scalar) (group.Scalar, error) {
+	if len(x) != len(px) {
+		return nil, errors.New("lagrange: bad input length")
+	}
+
+	zero := g.NewScalar()
+	for i := range x {
+		if x[i].IsEqual(zero) {
+			return nil, errors.New("lagrange: tried to evaluate on zero")
```
bwesterb, line 62: Wouldn't you then also want to check that each `x` is distinct?

group/secretsharing/poly.go
```diff
@@ -0,0 +1,73 @@
...
+		if j != int(index) {
+			num.Mul(num, x[j])
+			den.Mul(den, tmp.Sub(x[j], x[index]))
+		}
+	}
+
+	return num.Mul(num, tmp.Inv(den))
+}
+
+func LagrangeInterpolate(g group.Group, x, px []group.Scalar) (group.Scalar, error) {
```
bwesterb, line 54: Document what this function does.

group/secretsharing/ss.go
```diff
@@ -0,0 +1,138 @@
...
+import (
+	"errors"
+	"fmt"
+	"io"
+
+	"github.com/cloudflare/circl/group"
+)
+
+type SecretShare struct {
+	ID    uint64
```
bwesterb, line 17: `uint64` is a bit big.

group/secretsharing/ss.go
```diff
@@ -0,0 +1,138 @@
...
+	Shard(rnd io.Reader, secret group.Scalar) []SecretShare
+	Recover(shares []SecretShare) (secret group.Scalar, err error)
+}
+
+type secretSharing struct {
+	G    group.Group
+	T, N uint64
+}
+
+// New returns a struct implementing SecretSharing interface.
```
bwesterb, line 32: Document meaning of `t` and `n`.

group/secretsharing/ss.go
```diff
@@ -0,0 +1,138 @@
...
+
+func (s secretSharing) Shard(rnd io.Reader, secret group.Scalar) []SecretShare {
+	return s.generateShares(s.polyFromSecret(rnd, secret))
+}
+
+func (s secretSharing) Recover(shares []SecretShare) (group.Scalar, error) {
+	if l := len(shares); l <= int(s.T) {
+		return nil, fmt.Errorf("secretsharing: do not met threshold %v with %v shares", s.T, l)
+	} else if l > int(s.N) {
+		return nil, fmt.Errorf("secretsharing: %v shares above max number of shares %v", l, s.N)
```
bwesterb, line 69: Do we really want to error here?

group/secretsharing/ss.go
```diff
@@ -0,0 +1,138 @@
...
+}
+
+func (s secretSharing) Shard(rnd io.Reader, secret group.Scalar) []SecretShare {
+	return s.generateShares(s.polyFromSecret(rnd, secret))
+}
+
+func (s secretSharing) Recover(shares []SecretShare) (group.Scalar, error) {
+	if l := len(shares); l <= int(s.T) {
+		return nil, fmt.Errorf("secretsharing: do not met threshold %v with %v shares", s.T, l)
+	} else if l > int(s.N) {
```
bwesterb, line 68: Nit: no need for the else.

group/secretsharing/ss.go
```diff
@@ -0,0 +1,138 @@
...
+		return nil, fmt.Errorf("secretsharing: %v shares above max number of shares %v", l, s.N)
+	}
+
+	x := make([]group.Scalar, len(shares))
+	px := make([]group.Scalar, len(shares))
+	for i := range shares {
+		x[i] = s.G.NewScalar()
+		x[i].SetUint64(shares[i].ID)
```
bwesterb, line 76: Passing the same share multiple times leads to silent failure (well, depending on whether the scalar errors on inverting zero.)

group/secretsharing/ss.go
```diff
@@ -0,0 +1,138 @@
...
+
+type verifiableSecretSharing struct {
+	s secretSharing
+}
+
+// SecretSharing implements Feldman's secret sharing.
+type VerifiableSecretSharing interface {
+	Shard(rnd io.Reader, secret group.Scalar) ([]SecretShare, SharesCommitment)
+	Recover(shares []SecretShare) (secret group.Scalar, err error)
+	Verify(share SecretShare, coms SharesCommitment) bool
```
bwesterb, line 93: Document what this verifies.

group/secretsharing/ss.go
```diff
@@ -0,0 +1,138 @@
...
+}
+
+func (v verifiableSecretSharing) Shard(rnd io.Reader, secret group.Scalar) ([]SecretShare, SharesCommitment) {
+	poly := v.s.polyFromSecret(rnd, secret)
+	shares := v.s.generateShares(poly)
+
+	vecComm := make(SharesCommitment, v.s.T+1)
+	for i, ki := range poly.coeff {
+		vecComm[i] = v.s.G.NewElement()
+		vecComm[i].MulGen(ki)
```
bwesterb, line 111: We need to add a big fat warning. If our secret is a small scalar, it can easily be recovered from the commitment. E.g., it's easy to see if our secret was 2 from vecComm[0].

bwesterb (changes requested):

group/secretsharing/ss.go
armfazh: suggestions? int?

group/secretsharing/ss.go
armfazh: do you mean that returning the error is not necessary, and do panic instead. ?

group/secretsharing/ss.go
armfazh: what is the issue here? (didn't get your comment)

group/secretsharing/ss.go
armfazh: now, the Lagrange module checks whether the shares are all different.

group/secretsharing/poly.go
armfazh: done

group/secretsharing/poly.go
armfazh: good one

group/secretsharing/poly.go
```diff
@@ -0,0 +1,73 @@
...
+
+import (
+	"errors"
+	"io"
+
+	"github.com/cloudflare/circl/group"
+)
+
+type polynomial struct {
+	deg   uint64
```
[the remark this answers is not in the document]
armfazh: of course, it's simpler.

group/secretsharing/ss.go
bwesterb: `uint16` or `uint32`

group/secretsharing/ss.go
bwesterb: Sorry, I misread. This is fijne.

secretsharing/ss.go
```diff
@@ -0,0 +1,171 @@
...
+}
+
+func (s ss) Shard(rnd io.Reader, secret group.Scalar) []Share {
+	return s.generateShares(s.polyFromSecret(rnd, secret))
+}
+
+func (s ss) Recover(shares []Share) (group.Scalar, error) {
+	if l := len(shares); l <= int(s.t) {
+		return nil, fmt.Errorf("secretsharing: do not met threshold %v with %v shares", s.t, l)
+	} else if l > int(s.n) {
```
bwesterb, line 92: No need for the `else` because of the `return`.

bwesterb (approved):

secretsharing/ss.go
bwesterb: Didn’t see a warning.

group/secretsharing/ss.go
armfazh: resolved offline, no issue.

math/polynomial/polynomial.go
```diff
@@ -53,6 +53,14 @@ func (p Polynomial) Evaluate(x group.Scalar) group.Scalar {
 	return px
 }
```
cjpatton, line 55: Needs a docucommnet. Plus, I think it's worth documenting the fact that this is a deep-copy.

secretsharing/ss.go
```diff
@@ -0,0 +1,172 @@
+// Package secretsharing provides methods to split secrets in shares.
+//
+// A (t,n) secret sharing allows to split a secret into n shares, such that the
+// secret can be only recovered given more than t shares.
+//
+// The New function creates a Shamir secret sharing [1], which relies on
+// Lagrange polynomial interpolation.
```
cjpatton, line 7: Other schemes are possible, I would suggest renaming New to NewShamir or something.

secretsharing/ss.go
```diff
@@ -0,0 +1,172 @@
...
+//
+// The NewVerifiable function creates a Feldman secret sharing [2], which
```
cjpatton, line 9: Similar comment here. Maybe NewVerifiableFeldman

secretsharing/ss.go
```diff
@@ -0,0 +1,172 @@
...
+	"fmt"
+	"io"
+
+	"github.com/cloudflare/circl/group"
+	"github.com/cloudflare/circl/math/polynomial"
+)
+
+// Share represents a share of a secret.
+type Share struct {
+	ID    uint
```
cjpatton, line 30: Needs documentation. What is the ID? Is it unique, and if so, in what context? How is this used in cryptographic computations, if at all?

secretsharing/ss.go
```diff
@@ -0,0 +1,172 @@
...
+	Share group.Scalar
+}
+
+// SecretSharing implements a (t,n) Shamir's secret sharing.
```
cjpatton, line 34: Would this interface be implemented by schemes other than Shamir?

secretsharing/ss.go
```diff
@@ -0,0 +1,172 @@
...
+type SecretSharing interface {
+	// Params returns the t and n parameters of the secret sharing.
+	Params() (t, n uint)
+	// Shard splits the secret into n shares.
+	Shard(rnd io.Reader, secret group.Scalar) []Share
+	// Recover returns the secret provided more than t shares are given.
+	Recover(shares []Share) (secret group.Scalar, err error)
```
cjpatton, line 41: 1. Under what conditions might an error be returned? (Document them.)
2. We certainly require at least `t` shares. I wonder if it would be useful to also require *no more than* `t` shares? At least for Shamir, the result should be the same for any `len(shares) >= t`. I think it would clearer to enforce this.

secretsharing/ss.go
```diff
@@ -0,0 +1,172 @@
+// Package secretsharing provides methods to split secrets in shares.
+//
+// A (t,n) secret sharing allows to split a secret into n shares, such that the
+// secret can be only recovered given more than t shares.
```
cjpatton, line 4:
```suggestion
// secret can be recovered from any subset of t shares.
```

secretsharing/ss.go
```diff
@@ -0,0 +1,172 @@
...
+//
+// The New function creates a Shamir secret sharing [1], which relies on
+// Lagrange polynomial interpolation.
+//
+// The NewVerifiable function creates a Feldman secret sharing [2], which
+// extends Shamir's by allowing to verify that a share corresponds to the
+// secret.
```
cjpatton, line 11: I think we should try to define "verifiability" a bit more precisely. What does it mean for a share to "correspond" to the secret?

secretsharing/ss.go
```diff
@@ -0,0 +1,172 @@
...
+}
+
+type ss struct {
+	g    group.Group
+	t, n uint
+}
+
+// New returns a struct implementing SecretSharing interface. A (t,n) secret
+// sharing allows to split a secret into n shares, such that the secret can be
+// only recovered given more than t shares. It panics if 0 < t <= n does not
```
cjpatton, line 51:
```suggestion
// only recovered given more than t shares. Returns an error if 0 < t <= n does not
```

secretsharing/ss.go
```diff
@@ -0,0 +1,172 @@
...
+	zero := s.g.NewScalar()
+
+	return l.Evaluate(zero), nil
+}
+
+type SharesCommitment = []group.Element
+
+type vss struct{ s ss }
+
+// SecretSharing implements a (t,n) Feldman's secret sharing.
```
cjpatton, line 115:
```suggestion
// VerifiableSecretSharing implements a (t,n) Feldman's secret sharing.
```

cjpatton: Here again: Is it your intention that we will only implement Feldman's?

secretsharing/ss.go
```diff
@@ -0,0 +1,172 @@
...
+type VerifiableSecretSharing interface {
+	// Params returns the t and n parameters of the secret sharing.
+	Params() (t, n uint)
+	// Shard splits the secret into n shares, and a commitment of the secret
```
cjpatton, line 119: Commitment to what?

secretsharing/ss.go
```diff
@@ -0,0 +1,172 @@
...
+	// and the shares.
+	Shard(rnd io.Reader, secret group.Scalar) ([]Share, SharesCommitment)
+	// Recover returns the secret provided more than t shares are given.
+	Recover(shares []Share) (secret group.Scalar, err error)
+	// Verify returns true if the share corresponds to a committed secret using
+	// the commitment produced by Shard.
+	Verify(share Share, coms SharesCommitment) bool
+}
+
+// New returns a struct implementing VerifiableSecretSharing interface. A (t,n)
```
cjpatton, line 129:
```suggestion
// New returns a struct implementing the VerifiableSecretSharing interface. A (t,n)
```

secretsharing/ss.go
```diff
@@ -0,0 +1,172 @@
...
+// secret sharing allows to split a secret into n shares, such that the secret
+// can be only recovered given more than t shares. It is possible to verify
```
cjpatton, line 131:
```suggestion
// can be recovered with any t shares. It is possible to verify
```

secretsharing/ss.go
```diff
@@ -0,0 +1,172 @@
...
+// whether a share corresponds to a secret. It panics if 0 < t <= n does not
```
cjpatton, line 132:
```suggestion
// whether a share corresponds to a secret. Returns an error if 0 < t <= n does not
```

cjpatton (changes requested):
First pass focuses on API, I will look at the math next :)

High level questions about math/polynomial:
1. This is defined around the "scalars" of a "group".  Do the operations in this package require the scalars to comprise a finite field, or more specifically, the group to have prime order?
2. If "yes" to (1.), why not define this package around a generic finite field?

It's a bit odd that constructing a new impl of `SecretSharing` requires a `Group`. Shamir's (and Feldman's I imagine, but I haven't checked) is well-defined for any finite field, hence this API is more restrictive than necessary.

If this API is only intended for use with cryptographic protocols built from an implementation of `Group`, then I would make this very clear in the documentation.

secretsharing/ss.go
armfazh: updated.

secretsharing/ss.go
armfazh: agree, removed the interface

secretsharing/ss.go
armfazh: done

secretsharing/ss.go
armfazh: What about allowing passing `n >= len(shares) > t`, but only evaluating t+1 internally?

secretsharing/ss.go
armfazh: maybe not, I just remove the interface as is not really needed

secretsharing/ss.go
armfazh: added a simple description

secretsharing/ss.go
armfazh: updated text without entering in too much details.

secretsharing/ss.go
armfazh: done

secretsharing/ss.go
armfazh: done

math/polynomial/polynomial.go
armfazh: done

armfazh:
> 1. This is defined around the "scalars" of a "group".  Do the operations in this package require the scalars to comprise a finite field, or more specifically, the group to have prime order?
> 2. If "yes" to (1.), why not define this package around a generic finite field?

It's true that secret sharing package can be defined over a finite field. In CIRCL, currently we have no such abstraction. This may require to gather all the field implementations we have in CIRCL. 
For the purposes of threshold cryptography, we use ecc groups and its set of scalars. At this point we are covering this use case, but we can accommodate more use cases when needed.

secretsharing/ss.go
```diff
@@ -0,0 +1,159 @@
...
+// Shard splits the secret into n shares, and also returns a commitment to both
+// the secret and the shares.
+func (v vss) Shard(rnd io.Reader, secret group.Scalar) ([]Share, SharesCommitment) {
+	poly := v.s.polyFromSecret(rnd, secret)
+	shares := v.s.generateShares(poly)
+	coeffs := poly.Coefficients()
+	shareComs := make(SharesCommitment, len(coeffs))
+	for i := range coeffs {
+		shareComs[i] = v.s.g.NewElement().MulGen(coeffs[i])
+	}
```
cjpatton, line 131: Suppose a user secret shares `s` twice. Then the commitment to `s`, `g^s`, would be the same in each each secret sharing, thus leaking the fact that the secret is the same. There might be related attacks that are interesting, e.g.: if the second secret is `s+1`, then I can tell this by `g^(s+1)` is just `g * g^s`, where `g^s` is the commitment to the first secret.

It's notable that plain Shamir secret sharing does not have this deficiency. I think we need to warn users about this, at the very least. In fact, it probably makes sense to guide users towards "safe" applications. for instance, committing to a random scalar `s` is perfectly safe.

Stepping back: What's the motivation to implement Feldman's in particular? There might be other VSS schemes that are safe for other applications.

secretsharing/ss.go
```diff
@@ -0,0 +1,159 @@
...
+
+// Params returns the t and n parameters of the secret sharing.
+func (v vss) Params() (t, n uint) { return v.s.Params() }
+
+// Shard splits the secret into n shares, and also returns a commitment to both
+// the secret and the shares.
+func (v vss) Shard(rnd io.Reader, secret group.Scalar) ([]Share, SharesCommitment) {
+	poly := v.s.polyFromSecret(rnd, secret)
+	shares := v.s.generateShares(poly)
+	coeffs := poly.Coefficients()
```
cjpatton, line 127: I wonder if you could save a malloc here by accessing the coefficients directly below?

armfazh:
Hi folks, to move forward with this PR, please leave your comments to the latest changes. And approve if you are ok with the current status.

cc: @bwesterb @chris-wood @cjpatton

secretsharing/ss.go
```diff
@@ -0,0 +1,159 @@
+// Package secretsharing provides methods to split secrets in shares.
```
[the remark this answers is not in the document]
armfazh: done

secretsharing/ss.go
```diff
@@ -0,0 +1,159 @@
...
+// Let n be the number of parties, and t the number of corrupted parties such
+// that 0 <= t < n. A (t,n) secret sharing allows to split a secret into n
+// shares, such that the secret can be recovered from any subset of t+1 shares.
+//
+// The NewShamirSecretSharing function creates a Shamir secret sharing [1],
+// which relies on Lagrange polynomial interpolation.
+//
+// The NewFeldmanSecretSharing function creates a Feldman secret sharing [2],
+// which extends Shamir's by allowing to verify that a share was honestly
+// generated.
```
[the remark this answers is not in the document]
armfazh: I would say: "she holds a share of a ~recoverable~ committed secret."
Now it says: 
```
... extends Shamir's by allowing to verify that a share is part of a committed secret.
```

secretsharing/ss.go
```diff
@@ -0,0 +1,159 @@
...
+// if 0 <= t < n does not hold.
+func NewShamirSecretSharing(g group.Group, t, n uint) (ss, error) {
+	if t >= n {
+		return ss{}, errors.New("secretsharing: bad parameters")
+	}
+	return ss{g: g, t: t, n: n}, nil
+}
+
+// Params returns the t and n parameters of the secret sharing.
+func (s ss) Params() (t, n uint) { return s.t, s.n }
```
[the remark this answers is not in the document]
armfazh: done. Refactored now.

secretsharing/ss.go
```diff
@@ -0,0 +1,159 @@
...
+
+func (s ss) polyFromSecret(rnd io.Reader, secret group.Scalar) (p polynomial.Polynomial) {
+	c := make([]group.Scalar, s.t+1)
+	for i := range c {
+		c[i] = s.g.RandomScalar(rnd)
```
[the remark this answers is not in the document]
armfazh: done

secretsharing/ss.go
```diff
@@ -0,0 +1,159 @@
...
+	return polynomial.New(c)
+}
+
+func (s ss) generateShares(poly polynomial.Polynomial) []Share {
+	shares := make([]Share, s.n)
+	x := s.g.NewScalar()
+	for i := range shares {
+		id := i + 1
+		x.SetUint64(uint64(id))
+		shares[i].ID = uint(id)
```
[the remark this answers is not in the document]
armfazh: done

secretsharing/ss.go
```diff
@@ -0,0 +1,159 @@
...
+func (s ss) Recover(shares []Share) (group.Scalar, error) {
+	if l := len(shares); l <= int(s.t) {
+		return nil, fmt.Errorf("secretsharing: does not reach the threshold %v with %v shares", s.t, l)
+	} else if l > int(s.n) {
+		return nil, fmt.Errorf("secretsharing: %v shares above max number of shares %v", l, s.n)
+	}
+
+	x := make([]group.Scalar, s.t+1)
+	px := make([]group.Scalar, s.t+1)
+	for i := range shares[:s.t+1] {
```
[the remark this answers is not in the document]
armfazh: > As a user who knows nothing about secret sharing, I would wonder if I would get different secrets when recovering from different numbers of shares.

This behaviour is captured in the unit tests. So it's safe to pass a number of shares between t+1 and n.

secretsharing/ss.go
```diff
@@ -0,0 +1,159 @@
...
+// that 0 <= t < n. A (t,n) secret sharing allows to split a secret into n
+// shares, such that the secret can be recovered from any subset of t+1 shares.
+//
+// The NewShamirSecretSharing function creates a Shamir secret sharing [1],
+// which relies on Lagrange polynomial interpolation.
+//
+// The NewFeldmanSecretSharing function creates a Feldman secret sharing [2],
+// which extends Shamir's by allowing to verify that a share was honestly
+// generated.
+//
```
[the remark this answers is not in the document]
armfazh: Done

> Is an implementation of group.Group guaranteed to have prime order?

Yes, for the instances of this library.

secretsharing/ss.go
```diff
@@ -0,0 +1,159 @@
...
+func NewFeldmanSecretSharing(g group.Group, t, n uint) (vss, error) {
+	s, err := NewShamirSecretSharing(g, t, n)
+	return vss{s}, err
+}
+
+// Params returns the t and n parameters of the secret sharing.
+func (v vss) Params() (t, n uint) { return v.s.Params() }
+
+// Shard splits the secret into n shares, and also returns a commitment to both
+// the secret and the shares.
```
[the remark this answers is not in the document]
armfazh: done

secretsharing/ss.go
armfazh: > thus leaking the fact that the secret is the same.

This is similar to PKI, if a user publishes two certificates holding the same public key, then it can be inferred that the corresponding private key is the same. 

Also not sure whether or not the Sharding method should protect against this case.

Added this warning: `Sharding a secret more than once produces ShareCommitments with the same first entry`


> What's the motivation to implement Feldman's in particular?

This scheme is going to be used in the frost package and for ECDSA threshold signatures.

secretsharing/ss.go
armfazh: good one. Refactored now, it saves allocs.

chris-wood (changes requested):
Functionally, this seems correct, but I think we can still improve the API to make it easier to build things like STAR on top of this.

secretsharing/ss.go
cjpatton: SGTM.

math/polynomial/polynomial.go
```diff
@@ -53,6 +56,15 @@ func (p Polynomial) Evaluate(x group.Scalar) group.Scalar {
...
 
+// Coefficient returns a deep-copy of the n-th polynomial's coefficient.
+// Note coefficients are sorted in ascending order with respect to the degree.
+func (p Polynomial) Coefficient(n uint) group.Scalar {
```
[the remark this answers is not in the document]
armfazh: The odd case is `Degree()` returning -1 for the special case of the zero polynomial. We might remove this special case. 
@bwesterb what do you think?

secretsharing/ss.go
```diff
@@ -0,0 +1,165 @@
...
+	"io"
+
+	"github.com/cloudflare/circl/group"
+	"github.com/cloudflare/circl/math/polynomial"
+)
+
+// Share represents a share of a secret.
+type Share struct {
+	ID    uint64       // ID uniquely identifies a share in a secret sharing instance.
+	Value group.Scalar // Value stores the share generated from a secret sharing instance.
```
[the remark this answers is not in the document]
armfazh: I initially used uint64 as the identifiers (the values of x) are always in the range 1 to n.

Changing to group Scalar makes sense, even though the internal values are the same: scalars in the range from 1 to n. 
--
The reason for ID is it will help to identify the party's share (the x coord) and its Value (i.e., p(x))

I didn't use the geometric nomenclature as it is offloaded to the `math/polynomial` package.

secretsharing/ss.go
```diff
@@ -0,0 +1,165 @@
...
+type SecretSharing struct {
+	g    group.Group
+	t, n uint
+}
+
+// NewShamirSecretSharing implements a (t,n) Shamir's secret sharing.
+// A (t,n) secret sharing allows to split a secret into n shares, such that the
+// secret can be only recovered from any subset of t+1 shares. Returns an error
+// if 0 <= t < n does not hold.
+func NewShamirSecretSharing(g group.Group, t, n uint) (SecretSharing, error) {
```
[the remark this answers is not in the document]
armfazh: done

secretsharing/ss.go
```diff
@@ -0,0 +1,165 @@
...
+
+type VerifiableSecretSharing struct{ s SecretSharing }
+
+// NewFeldmanSecretSharing implements a (t,n) Feldman's verifiable secret
+// sharing. A (t,n) secret sharing allows to split a secret into n shares, such
+// that the secret can be only recovered from any subset of t+1 shares. This
+// method is verifiable because once the shares and the secret are committed
+// during sharding, one can later verify whether the share was generated
+// honestly. Returns an error if 0 < t <= n does not hold.
+func NewFeldmanSecretSharing(g group.Group, t, n uint) (VerifiableSecretSharing, error) {
```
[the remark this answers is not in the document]
armfazh: done

secretsharing/ss.go
```diff
@@ -0,0 +1,159 @@
...
+// that the secret can be only recovered from any subset of t+1 shares. It's
+// verifiable as it allows checking whether a share is part of a secret committed
+// during sharding.
+func NewFeldmanSecretSharing(t uint) (v VerifiableSecretSharing) { v.s.t = t; return }
+
+// Shard splits the secret into n shares, and also returns a commitment to both
+// the secret and the shares. The ShareCommitment must be sent to each party
+// so each party can verify its share is correct. Sharding a secret more
+// than once produces ShareCommitments with the same first entry.
+func (v VerifiableSecretSharing) Shard(rnd io.Reader, secret group.Scalar, n uint) ([]Share, SharesCommitment, error) {
```
cjpatton, line 111: Consider deduplicating this code and https://github.com/cloudflare/circl/pull/348/files#diff-72aa5d969ef83bfdefc8848bc63ba19bc4a5e483f4882a6da009778470690dfbR60-R70.

cjpatton (approved):

secretsharing/ss.go
armfazh: done

secretsharing/ss.go
```diff
@@ -0,0 +1,183 @@
...
+}
+
+type Splitter struct {
+	g    group.Group
+	t    uint
+	poly polynomial.Polynomial
+}
+
+// NewSplitter returns a Splitter that can shard a secret with threshold t.
+func NewSplitter(rnd io.Reader, t uint, secret group.Scalar) (sp Splitter) {
```
chris-wood, line 139: This interface isn't used in the tests -- what is the use case?

secretsharing/ss.go
```diff
@@ -0,0 +1,183 @@
...
+	"io"
+
+	"github.com/cloudflare/circl/group"
+	"github.com/cloudflare/circl/math/polynomial"
+)
+
+// Share represents a share of a secret.
+type Share struct {
+	// ID uniquely identifies a share in a secret sharing instance.
+	ID group.Scalar
```
chris-wood, line 34: Can we rename this to something else? Maybe `Input`, and maybe rename `Value` to `Output`?

chris-wood (commented):
I think the interface still isn't quite right, so let me step back and identify the use cases I think are important to solve.

1. A user wants to share a secret with `n` other users with a recovery threshold of `t`. This is the classic secret sharing setting. The caller could provide "IDs," or they could be provided at random. I think it's probably easier if they're generated randomly and then the caller assigns these IDs to each user. 
2. `t` users want to independently produce random shares of the same secret such that someone else (a collector or combiner) can recover the secret. Each user has to agree on the secret sharing polynomial independently and non-interactively, i.e., without communication, otherwise this doesn't work. This is the STAR use case.

Based on my understanding of this PR, it looks like (1) is partly addressed (though I would make the API somewhat different so as to not require a monotonically increasing selection of share input [x coordinates]), but (2) does not seem to be addressed. 

It's fine if we want to address these in separate PRs, but I'd like to understand what the plan is for both. @armfazh, can you please clarify? In particular, if the plan is to do this in two separate PRs, can you please clarify how the use case in (1) is satisfied?

secretsharing/ss.go
armfazh: removed

armfazh:
> I think the interface still isn't quite right, so let me step back and identify the use cases I think are important to solve.
> 
> 1. A user wants to share a secret with `n` other users with a recovery threshold of `t`. This is the classic secret sharing setting. The caller could provide "IDs," or they could be provided at random. I think it's probably easier if they're generated randomly and then the caller assigns these IDs to each user.
> 2. `t` users want to independently produce random shares of the same secret such that someone else (a collector or combiner) can recover the secret. Each user has to agree on the secret sharing polynomial independently and non-interactively, i.e., without communication, otherwise this doesn't work. This is the STAR use case.
> 
> Based on my understanding of this PR, it looks like (1) is partly addressed (though I would make the API somewhat different so as to not require a monotonically increasing selection of share input [x coordinates]), but (2) does not seem to be addressed.
> 
> It's fine if we want to address these in separate PRs, but I'd like to understand what the plan is for both. @armfazh, can you please clarify? In particular, if the plan is to do this in two separate PRs, can you please clarify how the use case in (1) is satisfied?


The code I just pushed satisfy these cases: 
a) both _textbook_ Shamir and Feldman secret; and 
b) the case when the share is generated with a prescribed ID (case 1).

For case 2: one solution is that `rand` (the first param of `New(rand, ...)`) can be passed as a PRNG with a seed, this makes the internals of the secret sharing deterministic (and reproducible). 
Other solutions can be adopted later (but not in this PR).

PR #356, opened 2022-07-29
group: removes order method from group interface

### LINKED ISSUE #214: Group Order() function signature oddity

armfazh:
There is no point in returning a `Scalar` for Order method.
Also, returning a big.Int is not desirable.
Since the group order is public and well-know for the groups supported, then it's better to remove the method.

Based on top of: #355  (merge that first)
Fixes #214

### DIFF
--- a/group/group.go
+++ b/group/group.go
@@ -26,8 +26,6 @@ type Group interface {
 	Identity() Element
 	// Creates an element of the group set to the generator of the group.
 	Generator() Element
-	// Returns a scalar set to the group order.
-	Order() Scalar
 	// RandomElement creates an element chosen at random (using randomness
 	// from rnd) from the set of group elements. Use crypto/rand.Reader as
 	// a cryptographically secure random number generator
--- a/group/group_test.go
+++ b/group/group_test.go
@@ -164,15 +164,17 @@ func testCSelect(t *testing.T, testTimes int, g group.Group) {
 }
 
 func testOrder(t *testing.T, testTimes int, g group.Group) {
+	I := g.Identity()
 	Q := g.NewElement()
-	order := g.Order()
+	minusOne := g.NewScalar().SetUint64(1)
+	minusOne.Neg(minusOne)
 	for i := 0; i < testTimes; i++ {
 		P := g.RandomElement(rand.Reader)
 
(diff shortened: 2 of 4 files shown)


bwesterb (approved):

chris-wood (changes requested):
> Since the group order is public and well-know for the groups supported, then it's better to remove the method.

This is _precisely_ why we want to keep the function on the Group interface. Applications should not need to bake this into their code, possibly getting it wrong. (I could see them confusing the underlying field order with the group's order, for example.) They should just call the `Order` function on the desired group's interface.

bwesterb:
> They should just call the `Order` function on the desired group's interface.

The current `Order` function is broken because it uses `Scalar`.

I think it's bad to return a `big.Int`.

Let's step back: for what reasons would the user need the order and can we provide functions directly for those use-cases instead?

chris-wood:
> I think it's bad to return a big.Int.

Say more?

> Let's step back: for what reasons would the user need the order and can we provide functions directly for those use-cases instead?

The existing use cases we have are for determining scalar validity during deserialization, but that should already be covered by the deserialize functions on the Scalar interface. I don't know of any other uses, but as above, I don't think it's a good idea to punt this to applications should they need to use it for something we've not considered.

bwesterb:
> Say more?

Operations on `big.Int` are not constant time. We can add it, but then we need to add some very explicit warnings.

bytemare:
One use case: some app doesn't implement itself any group, and uses this package, but reimplements hash-to-scalar. The app doesn't have all the parameters of the group it handles, only abstractions, so it's useful for that app to get the order from the interface.

bwesterb:
What is your preference on how to proceed @armfazh ?

armfazh:
> What is your preference on how to proceed @armfazh ?

I have updated this PR to remove the `Order` method from the `Group` interface.

PR #357, opened 2022-07-30
math: adds polynomials and Lagrange polynomials.

armfazh:
adds polynomials and Lagrange polynomials to be used in secret sharing package

### DIFF
--- a/math/polynomial/polynomial.go
+++ b/math/polynomial/polynomial.go
@@ -1,0 +1,147 @@
+// Package polynomial provides representations of polynomials over the scalars
+// of a group.
+package polynomial
+
+import "github.com/cloudflare/circl/group"
+
+// Polynomial stores a polynomial over the set of scalars of a group.
+type Polynomial struct {
+	// Internal representation is in polynomial basis:
+	// Thus,
+	//     p(x) = \sum_i^k c[i] x^i,
+	// where k = len(c)-1 is the degree of the polynomial.
+	c []group.Scalar
+}
+
+// New creates a new polynomial given its coefficients in ascending order.
+// Thus,
+//     p(x) = \sum_i^k c[i] x^i,
+// where k = len(c)-1 is the degree of the polynomial.
+//
+// The zero polynomial has degree equal to -1 and can be instantiated passing
+// nil to New.
+func New(coeffs []group.Scalar) (p Polynomial) {
+	if l := len(coeffs); l != 0 {
+		p.c = make([]group.Scalar, l)
+		for i := range coeffs {
+			p.c[i] = coeffs[i].Copy()
+		}
+	}
+
+	return
+}
+
+func (p Polynomial) Degree() int {
+	i := len(p.c) - 1
+	for i > 0 && p.c[i].IsZero() {
+		i--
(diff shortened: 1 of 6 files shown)


math/polynomial/polynomial.go
```diff
@@ -0,0 +1,123 @@
...
+	c []group.Scalar
+}
+
+// New creates a new polynomial given its coefficients in ascending order.
+// Thus,
+//     p(x) = \sum_i^k c[i] x^i,
+// where k = len(c)-1 is the degree of the polynomial.
+func New(coeffs []group.Scalar) Polynomial {
+	if len(coeffs) == 0 {
+		panic("polynomial: invalid length")
```
bwesterb, line 26: If this is a generic Polynomial API: how do represent the zero polynomial? If you do it as `[0]`, then degree doesn’t match length-1 anymore.

math/polynomial/polynomial.go
```diff
@@ -0,0 +1,123 @@
...
+	}
+	return Polynomial{coeffs}
+}
+
+func (p Polynomial) Degree() uint { return uint(len(p.c) - 1) }
```
bwesterb, line 31: What if `New` is called with zeroes on the higher coefficients?

math/polynomial/polynomial.go
```diff
@@ -0,0 +1,123 @@
...
+//     p(x) = \sum_i^k y[i] L_j(x), where k is the degree of the polynomial,
+//     L_j(x) = \prod_i^k (x-x[i])/(x[j]-x[i]),
+//     y[i] = p(x[i]), and
+//     all x[i] are different.
+// It panics if one of these conditions does not hold.
+func NewLagrangePolynomial(g group.Group, x, y []group.Scalar) LagrangePolynomial {
+	if len(x) != len(y) {
+		panic("lagrange: invalid length")
+	}
+	if !isAllDifferent(x) {
```
bwesterb, line 66: Nit: areAllDifferent.

math/polynomial/polynomial.go
```diff
@@ -0,0 +1,123 @@
...
+		}
+	}
+
+	return num.Mul(num, den.Inv(den))
+}
+
+func isAllDifferent(x []group.Scalar) bool {
+	m := make(map[string]struct{})
+	for i := range x {
+		k := x[i].(fmt.Stringer).String()
```
bwesterb, line 116: You could also use `MarshalBinary` and cast the `[]byte` to `string`.

bwesterb (approved):
Left some minor comments.

math/polynomial/polynomial.go
armfazh: this implementation does not consider the representation of the zero polynomial.
It can be done by adding some checks here and there, but it's not needed for the use case. We might added if it's really needed.

armfazh: on a second thought, I added support to it, it wasn't so difficult after all.

math/polynomial/polynomial.go
armfazh: updated to consider this case

PR #364, opened 2022-08-12
Adds RSA Threshold signatures

jbis9051:
# RSA Threshold Signatures

This is an implementation of ["Practical Threshold Signatures" by Victor Shoup](https://www.iacr.org/archive/eurocrypt2000/1807/18070209-new.pdf).
Protocol 1 is implemented.

## Threshold Primer

Let *l* be the total number of players, *t* be the number of corrupted players, and *k* be the threshold.
The idea of threshold signatures is that at least *k* players need to participate to form a valid signature.

Setup consists of a dealer generating *l* key shares from a key pair and "dealing" them to the players. In this implementation the dealer is trusted.

During the signing phase, at least *k* players use their key share and the message to generate a signature share.
Finally, the *k* signature shares are combined to form a valid signature for the message.

## Modifications

1. Our implementation is not robust. That is, the corrupted players can prevent a valid signature from being formed by the non-corrupted players. As such, we remove all verification.
2. The paper requires p and q to be safe primes. We do not.


tss/rsa/rsa_threshold.go
```diff
@@ -0,0 +1,251 @@
...
+		// x_i^{|2λ|}
+		tmp.Exp(share.xi, &abslam, pub.N)
+		if abslam.Cmp(&exp) == 1 {
+			tmp.ModInverse(&tmp, pub.N)
+		}
+		// TODO  first compute all the powers for the negative exponents (but don't invert yet); multiply these together and then invert all at once. This is ok since (ab)^-1 = a^-1 b^-1
+
+		w.Mul(w, &tmp)
+	}
+	w.Mod(w, pub.N)
```
armfazh, line 171: need to add modular reduction, otherwise numbers grow fast.
```suggestion
		w.Mul(w, &tmp).Mod(w, pub.N)
	}

```

tss/rsa/rsa_threshold.go
```diff
@@ -0,0 +1,251 @@
...
+	// Sign(Pad(Hash(M)))
+
+	hasher := hash.New()
+	hasher.Write(msg)
+	digest := hasher.Sum(nil)
+
+	return padder.Pad(pub, hash, digest)
+}
+
+type Signature []byte
```
armfazh, line 141:
```suggestion
type Signature = []byte
```

tss/rsa/rsa_threshold.go
```diff
@@ -0,0 +1,251 @@
...
+
+// CombineSignShares combines t SignShare's to produce a valid signature
+func CombineSignShares(players int64, pub *rsa.PublicKey, shares []SignShare, msg []byte) (Signature, error) {
```
armfazh, line 144: is there any check that validates the right number of shares wrt the threshold?

tss/rsa/util.go
```diff
@@ -0,0 +1,26 @@
...
+var ONE = big.NewInt(1)
+
+func calculateDelta(l int64) *big.Int {
+	// ∆ = l!
+	delta := big.Int{}
+	delta.MulRange(1, l)
+	return &delta
+}
+
+func createPrivateKey(p, q *big.Int, e int) *rsa.PrivateKey {
```
armfazh, line 17: move this function to a *_test.go file, if is only used for testing.

tss/rsa/rsa_threshold_test.go
```diff
@@ -0,0 +1,419 @@
...
+
+func BenchmarkSignCombine_SHA256_4096_3_2_Scheme_BlindParallel(b *testing.B) {
+	const players = 3
+	const threshold = 2
+	const bits = 4096
+	const algo = crypto.SHA256
+	benchmarkSignCombineHelper(rand.Reader, true, b, players, threshold, bits, algo, PKS1v15)
+}
+
+func BenchmarkSignCombine_MD5_4096_3_2_Scheme(b *testing.B) {
```
armfazh, line 364: MD5 is no more in use.

tss/rsa/rsa_threshold_test.go
```diff
@@ -0,0 +1,419 @@
...
+		err = rsa.VerifyPSS(pub, algo, hashed, sig, padder.(*PSSPadder).Opts)
+	} else {
+		panic("logical error")
+	}
+
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	t.Log("verified signature")
```
armfazh, line 207:
```suggestion
```

tss/rsa/util.go
```diff
@@ -0,0 +1,26 @@
+package rsa
+
+import (
+	"crypto/rsa"
+	"math/big"
+)
+
+var ONE = big.NewInt(1)
+
+func calculateDelta(l int64) *big.Int {
```
armfazh, line 10:
```suggestion
func calculateDelta(l uint64) *big.Int {
```

tss/rsa/pss/pss.go
```diff
@@ -0,0 +1,169 @@
+// https://cs.opensource.google/go/go/+/refs/tags/go1.18.3:src/crypto/rsa/pss.go
```
armfazh, line 1: create an internal folder (tss/rsa/internal) to put these files.

armfazh (changes requested):
Looks good in general. Most of the comments are easy to fix.

tss/rsa/rsa_threshold.go
```diff
@@ -0,0 +1,251 @@
...
+	"errors"
+	"fmt"
+	"io"
+	"math"
+	"math/big"
+)
+
+// l or `players` - total number of players
+// t - the number of corrupted players
+// k or `threshold` - the number of signature shares needed to obtain a signature (aka the threshold)
```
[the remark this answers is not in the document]
jbis9051: k is `threshold`, Done.

tss/rsa/rsa_threshold.go
```diff
@@ -0,0 +1,251 @@
...
+
+func validateParams(players, threshold int64) error {
+	if players <= 1 {
+		return errors.New("rsa_threshold: players (l) invalid: should be > 1")
+	}
+	if threshold < 1 || threshold > players {
+		return fmt.Errorf("rsa_threshold: threshold (k) invalid: %d < 1 || %d > %d", threshold, threshold, players)
+	}
+	return nil
```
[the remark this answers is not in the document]
jbis9051: t is the number of corrupted players. This is unknown. This requirement is only for robustness.

tss/rsa/rsa_threshold.go
```diff
@@ -0,0 +1,251 @@
...
+}
+
+// Deal takes in an existing RSA private key generated elsewhere. If cache is true, cached values are stored in KeyShare taking up more memory by reducing Sign time.
+// See KeyShare documentation.
+func Deal(randSource io.Reader, players, threshold int64, key *rsa.PrivateKey, cache bool) ([]KeyShare, error) {
```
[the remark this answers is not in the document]
jbis9051: threshold is k here

tss/rsa/rsa_threshold_test.go
jbis9051: I think it's nice to see how much of the bench is hashing but we can remove

tss/rsa/util.go
jbis9051: calculateDelta calls MulRange which requires an int64. This is an internal function so there's no reason to force the conversion if it's not needed.

tss/rsa/rsa_threshold_test.go
armfazh: yes let's remove it, usually hashing is not issue for performance.

tss/rsa/rsa_threshold.go
```diff
@@ -0,0 +1,251 @@
...
+}
+
+// f(X) = Σ^{k-1}_{i=0} | a_i * X^i (mod m)
+func computePolynomial(k int64, a []*big.Int, x int64, m *big.Int) *big.Int {
+	sum := big.NewInt(0)
+
+	//  Σ^{k-1}_{i=0}
+	for i := int64(0); i <= k-1; i++ {
+		// X^i
+		// TODO optimize: we can compute x^{n+1} from the previous x^n
```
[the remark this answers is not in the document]
jbis9051: Added TODO

tss/rsa/keyshare.go
```diff
@@ -0,0 +1,190 @@
...
+	}
+
+	si := big.Int{}
+	bytes := make([]byte, siLen)
+	copy(bytes, data[3:3+siLen])
+	si.SetBytes(bytes)
+
+	var twoDeltaSi *big.Int
+	if len(data[3+siLen:]) > 0 {
+		tmp := make([]byte, len(data[3+siLen:]))
```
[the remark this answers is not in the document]
jbis9051: We require that the slice passed to this function is only bytes pertaining to keyshare. That is, you may not pass a slice that contains a serialized keyshare and other bytes after it. This is a requirement by the interface itself as we have no way to indicate to the caller how many bytes were used.

armfazh (approved):
Changes look good. Thanks @jbis9051

PR #368, opened 2022-08-16
 Bumping up to go 1.19 version.

armfazh:


go.mod
```diff
@@ -1,9 +1,9 @@
 module github.com/cloudflare/circl
 
-go 1.16
+go 1.17
```
cjpatton, line 3: Should this be 19?

armfazh: no, this is to announce "kind-of" support to previous versions.

PR #377, opened 2022-09-28
Add Scalar/SetBigInt function

chris-wood:
This change enables one to set Scalars to integer values encapsulated in big.Int types. Previously, only uint64 inputs were admitted.


bwesterb (approved):

bwesterb:
I've added a warning to the interface about `big.Int`s not being constant-time.

PR #378, opened 2022-10-04
Implement Classical McEliece

### LINKED ISSUE #360: Implement Classic McEliece
Classic McEliece is well-studied kem and suitable for offline file encrypt without decryption error. Will there be plan to implement it?

octeep:
This is a draft of a Classical McEliece that implements all the parameters specified in the round 3 submission. It passes the KAT tests provided in the official website. 

The current major problem right now is that public key generation takes so much time it causes tests on CI to timeout and fail. This is because the Gaussian Elimination loop inside `pkGen` in `pk_gen.go` can take millions of iterations for larger parameters. Vectorization should be able to make it faster by like 30x but I haven't implemented it yet. Currently there are only generic Go implementations.

Some test data outside of the KAT test are lifted from [a Rust implementation](https://github.com/Colfenor/classic-mceliece-rust). These test data are to ensure that some specific functions work correctly.

Closes #360


math/gf4096/gf4096.go
```diff
@@ -0,0 +1,83 @@
...
+
+	out = sqr(sqr(out))
+	out = Mul(out, tmp3) // a^1023 = a^(255*2*2 + 3)
+
+	out = Mul(sqr(out), a) // a^2047 = a^(1023*2 + 1)
+	return sqr(out)        // a^4094 = a^(2047 * 2)
+}
+
+// Div calculates a / b
+func Div(b, a Gf) Gf {
```
armfazh, line 81:
```suggestion
func Div(a, b Gf) Gf {
```

math/gf8192/gf8192.go
```diff
@@ -0,0 +1,130 @@
...
+	return uint16(x & GfMask)
+}
+
+// Inv calculates the multiplicative inverse of Gf element a
+func Inv(a Gf) Gf {
+	return Div(a, 1)
+}
+
+// Div calculates a / b
+func Div(b, a Gf) Gf {
```
armfazh, line 121:
```suggestion
	return Div(1, a)
}

// Div calculates a / b
func Div(a, b Gf) Gf {
```

armfazh:
@octeep thanks for pushing this. Also, could you please squash your commits.

We will review this code closely.

---

Definitely key generation takes lots of time (ARM64 tests timed out, while x64's not).

What about adding a build constraint to compile this code only on x64?

What do you think @bwesterb ?

bwesterb:
> What do you think @bwesterb ?

I'll have a look tomorrow.

octeep:
> @octeep thanks for pushing this. Also, could you please squash your commits.
> 
> We will review this code closely.
> 
> Definitely key generation takes lots of time (ARM64 tests timed out, while x64's not).
> 
> What about adding a build constraint to compile this code only on x64?
> 
> What do you think @bwesterb ?

I've squashed my commit. This probably shouldn't be merged yet since the Classic McEliece team might soon release their round 4 specs and software soon (See [this](https://groups.google.com/a/list.nist.gov/g/pqc-forum/c/I5LBixSRy-Q/m/dt1eR19eBgAJ)). It might be better to wait for them to publish the round 4 so this PR can be updated accordingly.

bwesterb:
Thanks for this @octeep 

>  This probably shouldn't be merged yet since the Classic McEliece team might soon release their round 4 specs and software soon

Ok.

Can I ask you how did this implementation came to be: is it a fresh implementation from the spec or did you translate an existing implementation (`ref` or rust)?

I had an initial look today. To set expectations: it will of course take quite some time to carefully review this.

Some initial comments:

1. Key generation is very slow and I confirmed it's mostly in the Gaussian elimination. It's much slower than the C reference implementation. How much experience do you have with optimising for Go? If you're new to it: the Go compiler is very dumb. They favour compilation speed against optimisation potential. Dumb things like unrolling loops can make a big difference. I think an optimised Gaussian elimination might be worthwhile.
2. Have you given thought to organising the types/code in such a way that it's easy to add in SIMD optimisations?
3. Having megabytes of test data is unworkable. Please reduce that to a few or even a single hash.  
4. The comments on the code aren't terrible, definitely compared to most crypto code, but we'd like to have a high standard.

octeep:
> Thanks for this @octeep
> 
> > This probably shouldn't be merged yet since the Classic McEliece team might soon release their round 4 specs and software soon
> 
> Ok.
> 
> Can I ask you how did this implementation came to be: is it a fresh implementation from the spec or did you translate an existing implementation (`ref` or rust)?

Most of the code is based on the 'Optimized Implementation' in the round 3 NIST submission package by the Classic McEliece team. However, when some methods cannot directly be ported from C (for example due to pointer arithmetic), the code is lifted from the Rust implementation instead. The Rust implementation is also based on the Optimized Implementation as far as I can tell.
> 
> I had an initial look today. To set expectations: it will of course take quite some time to carefully review this.
> 
> Some initial comments:
> 
> 1. Key generation is very slow and I confirmed it's mostly in the Gaussian elimination. It's much slower than the C reference implementation. How much experience do you have with optimising for Go? If you're new to it: the Go compiler is very dumb. They favour compilation speed against optimisation potential. Dumb things like unrolling loops can make a big difference. I think an optimised Gaussian elimination might be worthwhile.

I unfortunately do not have much experience with Go optimization. I also want to note that when I compile the C optimized implementation with vectorization explicitly disabled, the performance is only slightly faster than Go. When I compile the C optimized implementation and tell gcc to show the optimization tree, I see that gcc has performed vectorization automatically, which is why it is much faster than Go.

The `vec` implementations in Addition Implementations from the NIST submission package seems to perform much faster even when I compiled it with vectorization disabled. Maybe this can be ported to generic Go for an optimized Gaussian Elimination?

> 2. Have you given thought to organising the types/code in such a way that it's easy to add in SIMD optimisations?

If we are talking about optimizing `pk_gen` specifically then I don't think it would be too hard to add SIMD optimizations. Just add some build flags in the `pk_gen` template and some slight refactoring should be enough.

If we are talking about optimizing all the components with SIMD then no, I haven't given much thoughts into it. 

> 3. Having megabytes of test data is unworkable. Please reduce that to a few or even a single hash.

I will look into that. I don't think reducing it to a few hashes  can work since the test cases contain sample inputs. Maybe we can compress it or represent these inputs in raw bytes rather than as hexadecimal ASCII.

> 4. The comments on the code aren't terrible, definitely compared to most crypto code, but we'd like to have a high standard.

I will try to document the code to the best of my understanding.

octeep:
I have managed to trim down `testdata.txt` to 90kb. Would this be acceptable?

bwesterb:
> Most of the code is translated from the 'Optimized Implementation' in the round 3 NIST submission package by the Classic McEliece team. However, when some methods cannot directly be ported from C (for example due to pointer arithmetic), the code is lifted from the Rust implementation instead. The Rust implementation is also based on the Optimized Implementation as far as I can tell.

Ok! Please add appropriate attributions.

> The `vec` implementations in Addition Implementations from the NIST submission package seems to perform much faster even when I compiled it with vectorization disabled. Maybe this can be ported to generic Go for an optimized Gaussian Elimination?

Yes, that's a great plan!

(For those reading along: `vec` packs a few field elements into `uint64`s.)

> If we are talking about optimizing all the components with SIMD then no, I haven't given much thoughts into it.

I'll get back to this when I found time to properly study the code.

> I will try to document the code to the best of my understanding.

Great! If you're unsure about certain parts, please leave a comment as such for us to fill in.

bwesterb:
> I have managed to trim down `testdata.txt` to 90kb. Would this be acceptable?

@armfazh ?

armfazh:
> I have managed to trim down `testdata.txt` to 90kb. Would this be acceptable?

After bzip2-ing that file, size gets down to 50 KB, which I think is acceptable.

armfazh:
As a general comment, please introduce `// TODO` comments in places where code optimization is needed, such as better number representation or SIMD.
They might be not resolved in this PR, but flagging them is good for further development.

octeep:
> Thanks again for your efforts!
> 
> > Most of the code is translated from the 'Optimized Implementation' in the round 3 NIST submission package by the Classic McEliece team. However, when some methods cannot directly be ported from C (for example due to pointer arithmetic), the code is lifted from the Rust implementation instead. The Rust implementation is also based on the Optimized Implementation as far as I can tell.
> 
> Ok! Please add appropriate attributions.

I'm not sure how I should add attributions here. I have added a notice on each `mceliece.go` saying that some code are translated from the Rust implementation, and included the author's name and github link. Do I need to do more than that? (e.g. include their MIT license) 

> 
> > The `vec` implementations in Addition Implementations from the NIST submission package seems to perform much faster even when I compiled it with vectorization disabled. Maybe this can be ported to generic Go for an optimized Gaussian Elimination?
> 
> Yes, that's a great plan!

I am trying to do that in the `vec` branch in my fork repo. I can't promise if I will ever finish that but you can track the progress there if interested.

> (For those reading along: `vec` packs a few field elements into `uint64`s.)
> 
> > If we are talking about optimizing all the components with SIMD then no, I haven't given much thoughts into it.
> 
> I'll get back to this when I found time to properly study the code.
> 
> > I will try to document the code to the best of my understanding.
> 
> Great! If you're unsure about certain parts, please leave a comment as such for us to fill in.

octeep:
Ok good news: I managed to port the `vec` implementation of `pk_gen` to go and tested it on mceliece6960119. The KAT test now went from 120s to 17s. Still not as fast as the C implementation but it is much better than the current Go one.

bwesterb:
> Ok good news: I managed to port the `vec` implementation of `pk_gen` to go and tested it on mceliece6960119. The KAT test now went from 120s to 17s. Still not as fast as the C implementation but it is much better than the current Go one.

Good job!

octeep:
I've ported the `vec` implementation for all the systematic parameters (parameters that don't end with a `f`). Right now amd64 and macos CI finishes within 3 minutes, but I still can't get it to finish within 10 minutes on arm64.

After porting the `vec` implementation to semi-systematic parameters it should be able to finish within 10 minutes.

octeep:
> I've ported the `vec` implementation for all the systematic parameters (parameters that don't end with a `f`). Right now amd64 and MacOS implementations finish on CI finishes within 3 minutes, but I still can't get it to finish within 10 minutes on arm64.
> 
> After porting the `vec` implementation to semi-systematic parameters it should be able to finish within 10 minutes.

I've now also ported `vec` for semi-systematic parameters. On arm64 CI it now completes under 10 minutes so it passes CI now.

octeep:
> The round 4 version has been updated, see https://classic.mceliece.org/nist.html

I've updated the implementation to the round 4 specification. It passes KAT tests and finishes in 10 minutes on CI.

armfazh (changes requested):
Hi @octeep, so far I have checked the field arithmetic, it looks ok and there are some minor changes requested.
I will continue the review of other parts of the code.

d-z-m:
Hello :wave: , just wanted to say this change is desired, and that I hope code review proceeds in the near future! :rocket:

d-z-m:
Hello, just wanted to inquire if there's been any movement on this! Would love to see this get reviewed soon.

armfazh:
Hi, I will continue with the review of this PR in the next couple days. Thanks for your patience.

mariobm:
Any updates on this one?

stv0g:
I am also really interested in this PR as I am planning to use this for the [Go implementation](https://github.com/stv0g/go-rosenpass) of the [WireGuard Rosenpass Key Exchange](https://rosenpass.eu/).

stv0g:
I noticed that the current implementation differs from liboqs with respect to the ciphertext length.
See the following note:

https://classic.mceliece.org/mceliece-pc-20221023.pdf

stv0g:
> I've updated the implementation to the round 4 specification. It passes KAT tests and finishes in 10 minutes on CI.

The file headers are still mentioning round 3:

```
// Package mceliece348864 implements the IND-CCA2 secure key encapsulation mechanism
// mceliece348864 as submitted to round 3 of the NIST PQC competition and
// described in
```

stv0g:
@armfazh Can we remove the on-hold label from the issue? Round 4 specs have been released. There is nothing blocking this from getting finished :)

https://classic.mceliece.org/nist.html

stv0g:
Hi @bwesterb, @pufferffish, @armfazh,

is there something I can do to support you in getting this implementation merged?
Like code-review, testing or something similar?

Thanks :)

armfazh:
@stv0g we appreciate a code review, our team will be providing feedback as we get more bandwidth.

kem/schemes/schemes_test.go
```diff
@@ -155,6 +155,16 @@ func Example_schemes() {
 	// Kyber512
 	// Kyber768
 	// Kyber1024
+	// mceliece348864
```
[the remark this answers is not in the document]
stv0g: See liboqs for reference: https://github.com/open-quantum-safe/liboqs/blob/6e1f49aa48971b0fc8658e94ca41d1ad4fe480ee/src/kem/kem.h#L42

kem/mceliece/gen.go
```diff
@@ -0,0 +1,162 @@
...
+	McElieceParam8192128 = Param{
+		Gf:             "gf2e13",
+		PublicKeySize:  1357824,
+		PrivateKeySize: 14120,
+		CiphertextSize: 208,
+		SysN:           8192,
+		SysT:           128,
+	}
+	Instances = []Instance{
+		{Name: "mceliece348864", Param: McElieceParam348864},
```
stv0g, line 100: I propose to use names in line with liboqs here:
https://github.com/open-quantum-safe/liboqs/blob/6e1f49aa48971b0fc8658e94ca41d1ad4fe480ee/src/kem/kem.h#L42

This also matches the capitalization of e.g. `Kyber512`

kem/mceliece/templates/mceliece.templ.go
```diff
@@ -0,0 +1,896 @@
...
+	pi := [1 << gfBits]int16{}
+	pivots := uint64(0xFFFFFFFF)
+
+	copy(seed[1:], entropy[:])
+
+	for {
+		// expanding and updating the seed
+		err := shake256(r[:], seed[0:33])
+		if err != nil {
+			panic(err)
```
stv0g, line 103: Can we return the error in `DeriveKeypair()`.

I would say `panic()` these code paths should be discouraged.

kem/mceliece/templates/mceliece.templ.go
```diff
@@ -0,0 +1,896 @@
...
+	if err != nil {
+		return nil, nil, err
+	}
+	pk, sk := deriveKeyPair(seed[:])
+	return pk, sk, nil
+}
+
+func (*scheme) DeriveKeyPair(seed []byte) (kem.PublicKey, kem.PrivateKey) {
+	if len(seed) != seedSize {
+		panic("seed must be of length EncapsulationSeedSize")
```
stv0g, line 818: Same as below. Can we return an error here?

d-z-m:
Hi there,

Interested in seeing this move forward, any updates?

david415:
Hi. I get that everyone is busy doing other things and I also understand that I'm essentially the most impatient person in the world which is why I've decided to simply lift all this code into the katzenpost monorepo... with proper attribution of course.

However I would much rather this pull request get merged and the code maintained by you guys. I hate maintaining forks. But I hate waiting for things even more.

david415:
It looks like code contributions from your code contributor has stalled. i suggested you fork his dev branch and open a new pull request... and fix all the code corrections you were telling them to fix. If you don't do this, then this PR will likely continue to fester here meanwhile Rome is burning... society cracking at the foundations, abortion clinics need PQ KEMs like this one.

PR #379, opened 2022-10-17
Add deterministic blind RSA verifier

chris-wood:
This is necessary for some prototyping work in PPM.

I also updated the blind signer `Verifier` interface to implement the signature verification routine, as before it was assumed that the caller would know the internal details of how to verify signatures produced using a verifier.


chris-wood:
@armfazh can you please provide instructions for how to run `gofumpt`?

armfazh:
> @armfazh can you please provide instructions for how to run `gofumpt`?

Sure, here it is:

```sh
cd circl
go install mvdan.cc/gofumpt@latest
gofumpt -l -w ./blindsign
```

armfazh (approved):

PR #398, opened 2023-02-01
Use untyped consts for Kyber params

tmthrgd:
These two constants are unnecessarily typed. This flows through all the way to the kem/kyber/kyberN packages where they cause `PrivateKeySize` and `PublicKeySize` to also be unnecessarily typed.


bwesterb:
@armfazh This change should not break compatibility, but it will break compatibility if we need to reintroduce types on these constants for some reason later on. I don't have strong feelings either way. What do you think?

armfazh (approved):
Since these are constants (rather than variables) then I think is safe to keep them untyped.

PR #400, opened 2023-02-09
Adding Ascon, an AEAD lightweight cipher.

armfazh:
Reference: https://ascon.iaik.tugraz.at/index.html

Benchmarks:
old= naive substitution layer
new= 64-bit operations (constant-time)

|benchmark  | Operation    | old MB/s | new MB/s | speedup |
|---|---|---|---|---|
|Ascon128  | Open-64-8    | 1.71 | 125.59   | 73.44x |
|Ascon128  | Seal-64-8    | 1.58 | 123.48   | 78.15x |
|Ascon128  | Open-1350-8  | 2.86 | 211.97   | 74.12x |
|Ascon128  | Seal-1350-8  | 2.72 | 210.11   | 77.25x |
|Ascon128  | Open-8192-8  | 3.06 | 203.36   | 66.46x |
|Ascon128  | Seal-8192-8  | 2.91 | 209.82   | 72.10x |
|Ascon128a | Open-64-8    | 2.19 | 157.34   | 71.84x |
|Ascon128a | Seal-64-8    | 2.09 | 155.94   | 74.61x |
|Ascon128a | Open-1350-8  | 3.94 | 309.84   | 78.64x |
|Ascon128a | Seal-1350-8  | 4.15 | 290.40   | 69.98x |
|Ascon128a | Open-8192-8  | 4.14 | 308.30   | 74.47x |
|Ascon128a | Seal-8192-8  | 4.17 | 301.98   | 72.42x |


cipher/ascon/ascon.go
```diff
@@ -0,0 +1,260 @@
...
+	if len(key) != KeySize {
+		return nil, ErrKeySize
+	}
+	if !(m == Ascon128 || m == Ascon128a) {
+		return nil, ErrMode
+	}
+	c := new(Cipher)
+	c.mode = m
+	copy(c.key[:], key)
+	var _ cipher.AEAD = c
```
bwesterb, line 64: ^ why?

cipher/ascon/ascon.go
```diff
@@ -0,0 +1,260 @@
...
+			for i := 0; i < bcs; i++ {
+				a.state[i] ^= add[i]
+			}
+			a.perm(pB)
+		}
+		if len(add) >= 0 {
+			for i := 0; i < len(add); i++ {
+				a.state[i] ^= add[i]
+			}
+			a.state[len(add)] ^= 0x80
```
bwesterb, line 167: If `add` is a multiple of `bcs` then we don't XOR in `0x80`. Is that correct?

cipher/ascon/ascon.go
```diff
@@ -0,0 +1,260 @@
...
+		x[2] ^= roundConst[ri]
+
+		// pS -- substitution layer
+		for j := 0; j < 64; j++ {
+			sx := subs[0|
+				(((x[0]>>j)&0x1)<<4)|
+				(((x[1]>>j)&0x1)<<3)|
+				(((x[2]>>j)&0x1)<<2)|
+				(((x[3]>>j)&0x1)<<1)|
+				(((x[4]>>j)&0x1)<<0)]
```
bwesterb, line 231: Not constant time.

bwesterb (changes requested):

cipher/ascon/ascon.go
armfazh: it's a compile-time check that `c` implements the AEAD interface.

bwesterb: It's a bit out of place here — better to make it part of a test.

armfazh: done

cipher/ascon/ascon.go
armfazh: not sure,

bwesterb:
Nice moving to int64. How does the performance improve? The substitution table still isn't constant time.

cipher/ascon/ascon.go
armfazh: done

cipher/ascon/ascon.go
bwesterb: If I read §2.4.2 of the Ascon spec, then the padding can only be omitted in case the associated data is empty.

armfazh: there is an if at the top that jumps when add is empty

cipher/ascon/ascon.go
```diff
@@ -0,0 +1,261 @@
...
+	bcs := blockSize * int(a.mode)
+	pB := permB + 2*(int(a.mode)-1)
+	if len(add) > 0 {
+		for ; len(add) >= bcs; add = add[bcs:] {
+			for i := 0; i < bcs; i += 8 {
+				a.s[i/8] ^= binary.BigEndian.Uint64(add[i : i+8])
+			}
+			a.perm(pB)
+		}
+		if len(add) >= 0 {
```
bwesterb, line 161: This expression is always true, right?

armfazh: unless it's skipped by the if in line 154

bwesterb: `len(add) ≥ 0` is always true as a length is always... ≥0!

armfazh: correct, just changed.

bwesterb (approved):

PR #402, opened 2023-02-13
kem: add P-256 + Kyber768Draft00 hybrid

bwesterb:


bwesterb:
@armfazh

kem/hybrid/ckem.go
```diff
@@ -0,0 +1,211 @@
...
+		return false
+	}
+	if oth.scheme != sk.scheme {
+		return false
+	}
+	return subtle.ConstantTimeCompare(oth.key, sk.key) == 1
+}
+
+func (sk *cPrivateKey) Public() kem.PublicKey {
+	x, y := sk.scheme.curve.ScalarBaseMult(sk.key)
```
armfazh, line 87: can we cache the public key as part of the private key?

kem/hybrid/ckem.go
```diff
@@ -0,0 +1,211 @@
+package hybrid
+
+// TODO move over yo crypto/ecdh once we can assume Go 1.20.
```
armfazh, line 3:
```suggestion
// TODO move over to crypto/ecdh once we can assume Go 1.20.
```

kem/hybrid/ckem.go
```diff
@@ -0,0 +1,211 @@
...
+func (sch *cScheme) DeriveKeyPair(seed []byte) (kem.PublicKey, kem.PrivateKey) {
+	if len(seed) != sch.SeedSize() {
+		panic(kem.ErrSeedSize)
+	}
+	h := sha3.NewShake256()
+	_, _ = h.Write(seed)
+	buf := make([]byte, sch.PrivateKeySize())
+	_, _ = h.Read(buf)
+	rnd := bytes.NewReader(buf)
+	key, x, y, err := elliptic.GenerateKey(sch.curve, rnd)
```
armfazh, line 129: Note that GenerateKey function can loop more than once, requesting more bytes from the reader.

```suggestion
	key, x, y, err := elliptic.GenerateKey(sch.curve, h)
```

armfazh (approved):
looks good, check my comments.
Also I was wondering if we can reuse the group interface.

kem/hybrid/ckem.go
bwesterb: Ah, good point. `h` doesn't implement `io.Reader` though.

bwesterb: (yet)

kem/hybrid/ckem.go
bwesterb: We could, but we don't do it for `xKem` either at the moment. For the currently intended application (tls), it doesn't make a difference.

kem/hybrid/ckem.go
bwesterb: Thanks.

kem/hybrid/ckem.go
bwesterb: Ah, I can just use the `xof` package, nice.

bwesterb:
> Also I was wondering if we can reuse the group interface.

Might be possible, yes.

kem/hybrid/ckem.go
armfazh: this doesn't address my comment.
you need to pass the `xof` object, and remove `buf`.
1) xof already implements io.reader
2) the issue is that `buf` has fixed size, but it is not guaranteed that reader will be called only one time. It could be called many times.

bwesterb: You're not looking at the latest diff. :)

PR #403, opened 2023-02-16
ckem: pass xof to elliptic.GenerateKey directly

bwesterb:


kem/hybrid/ckem.go
```diff
@@ -123,10 +122,7 @@ func (sch *cScheme) DeriveKeyPair(seed []byte) (kem.PublicKey, kem.PrivateKey) {
 	}
 	h := xof.SHAKE256.New()
 	_, _ = h.Write(seed)
-	buf := make([]byte, sch.PrivateKeySize())
-	_, _ = h.Read(buf)
-	rnd := bytes.NewReader(buf)
-	key, x, y, err := elliptic.GenerateKey(sch.curve, rnd)
+	key, x, y, err := elliptic.GenerateKey(sch.curve, h)
```
Lekensteyn, line 125: Is this a bugfix? Previously, only enough randomness was provided for exactly one "private key". After this change, it could sample another key if the first secret was invalid:
```go
func GenerateKey(curve Curve, rand io.Reader) (priv []byte, x, y *big.Int, err error) {
	N := curve.Params().N
	bitSize := N.BitLen()
	byteLen := (bitSize + 7) / 8
	priv = make([]byte, byteLen)

	for x == nil {
		_, err = io.ReadFull(rand, priv)
// ...
		// If the scalar is out of range, sample another random number.
		if new(big.Int).SetBytes(priv).Cmp(N) >= 0 {
			continue
		}

		x, y = curve.ScalarBaseMult(priv)
	}
```

For reference, `sch.PrivateKeySize()` ends up with the same size:
```go
func (sch *cScheme) scSize() int {
	return (sch.curve.Params().N.BitLen() + 7) / 8
}
// ...
func (sch *cScheme) PrivateKeySize() int {
	return sch.scSize()
}
```

Lekensteyn (approved):
Change looks reasonable, perhaps the commit message could include a further motivation.

PR #404, opened 2023-02-19
Add Ascon-80pq to cipher\ascon

dhcgn:
Since this repo has a focus on PQ, I would like to suggest including the PQ optimized variant of Ascon.

> Ascon-80pq has an increased key-size to provide more resistance against a quantum
> adversary using Grover’s algorithm for key search. Since Ascon-128 and Ascon-
> 80pq share the same building blocks and same parameters except the size of the key,
> we claim the same security for Ascon-80pq against classical attacks as for Ascon-128.
>[...]
> The only difference between Ascon-80pq and Ascon-128 is the increased length of
> the key. This increased key length provides additional protection against exhaustive
> key search in the case the availability of quantum computers becomes evident. Since
> the other tunable security parameters (the number of rounds of the permutations)
> have not been increased, the security claim for Ascon-80pq against classical attacks
> stays the same as for Ascon-128.
>
> Source: https://ascon.iaik.tugraz.at/files/asconv12-nist.pdf


armfazh (approved):
This looks good, thanks @dhcgn

armfazh:
Could you please squash the commits, so we can merge it.

armfazh:
I just generated the test vectors file for Ascon80pq using pyascon.

https://github.com/meichlseder/pyascon

The ciphertext do not matches with the test vector in Go. Could you please confirm it.

armfazh (changes requested):
hold until having matching test vectors

armfazh (approved):
Ok, my local branch of pyascon was outdated. Pyascon impl had a bug (see https://github.com/meichlseder/pyascon/pull/2). Confirming test vectors are ok.

PR #409, opened 2023-03-06
tkn20: prevent panics on key gen errors

tmthrgd:
Both SystemSecretKey.KeyGen and Setup try to dereference the return values from abe/cpabe/tkn20/internal/tkn without checking for an error. On error, these values are nil and the functions panic.

This is easy to reproduce by passing an io.Reader that returns an error.


tmthrgd:
I'm not sure if you want tests for this or not, let me know and I can add them.

tanyav2 (approved):

tanyav2:
Feel free to merge

Issue #286, opened 2021-09-21
Add constant-time function for bls12/Scalar/SetBytes


armfazh:
Provided we fix the constant-time issue soon enough.

_Originally posted by @bwesterb in https://github.com/cloudflare/circl/pull/277#pullrequestreview-759586309_

PR #411, opened 2023-03-06
tkn20,kyber,x25519,x448: plug constant-time leaks

tmthrgd:
In particular leaking z in kyber could be quite damaging: https://groups.google.com/a/list.nist.gov/g/pqc-forum/c/SJ31w0QSmIM/m/XgdBgh3wAwAJ

The changes to x25519 and x448 are unlikely to be needed, but it's more idiomatic at least.

### DIFF
--- a/pke/kyber/kyber768/internal/cpapke.go
+++ b/pke/kyber/kyber768/internal/cpapke.go
@@ -3,6 +3,8 @@
 package internal
 
 import (
+	"crypto/subtle"
+
 	"github.com/cloudflare/circl/internal/sha3"
 	"github.com/cloudflare/circl/pke/kyber/internal/common"
 )
@@ -172,5 +174,5 @@ func (sk *PrivateKey) Equal(other *PrivateKey) bool {
 			ret |= sk.sh[i][j] ^ other.sh[i][j]
 		}
 	}
-	return ret == 0
+	return subtle.ConstantTimeEq(int32(ret), 0) == 1
 }
--- a/pke/kyber/kyber768/internal/cpapke.go
+++ b/pke/kyber/kyber768/internal/cpapke.go
@@ -3,8 +3,6 @@
 package internal
 
 import (
-	"crypto/subtle"
-
 	"github.com/cloudflare/circl/internal/sha3"
 	"github.com/cloudflare/circl/pke/kyber/internal/common"
 )
@@ -174,5 +172,5 @@ func (sk *PrivateKey) Equal(other *PrivateKey) bool {
 			ret |= sk.sh[i][j] ^ other.sh[i][j]
 		}
 	}
-	return subtle.ConstantTimeEq(int32(ret), 0) == 1
+	return ret == 0
 }
--- a/abe/cpabe/tkn20/internal/tkn/tk.go
+++ b/abe/cpabe/tkn20/internal/tkn/tk.go
@@ -3,7 +3,7 @@
 package tkn
 
 import (
-	"bytes"
+	"crypto/subtle"
 	"encoding/binary"
 	"fmt"
 	"io"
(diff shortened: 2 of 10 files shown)


pke/kyber/kyber768/internal/cpapke.go
```diff
@@ -172,5 +174,5 @@ func (sk *PrivateKey) Equal(other *PrivateKey) bool {
 			ret |= sk.sh[i][j] ^ other.sh[i][j]
 		}
 	}
-	return ret == 0
+	return subtle.ConstantTimeEq(int32(ret), 0) == 1
```
bwesterb, line 177: This doesn't do anything: the code generated is the same. https://godbolt.org/z/eGfrdcr7s

ConstantTimeEq exists so you can pass its value to a constant-time select or copy. In this case, the (in)equality of the _whole of_ the private key is not sensitive, so the current code is fine.

bwesterb (changes requested):
Thanks.

One of your changes doesn't do anything.

About the Kyber one: when can you get someone to compare two private keys where you know the z of one of them? So I'm not convinced this is "damaging". Anyway, it's good cryptographic hygiene to always compare secrets in constant-time, so it's a good one to take.

The curve ones are comparing public keys... but it already uses a ct-compare. Well, I'll leave that to @armfazh .

I'll leave the tkn to @tanyav2 — I don't know whether the pdfKey is sensitive from the top of my head.

pke/kyber/kyber768/internal/cpapke.go
tmthrgd: `subtle.ConstantTimeEq(v, 0) == 1` is already used widely throughout the library. See dilithiums `(*PrivateKey).Equal` and bls12381/ff's `ctUint64Eq`.

bwesterb: To quote Dilithium's `PrivateKey.Equal()`:

```
    ret := (subtle.ConstantTimeCompare(sk.rho[:], other.rho[:]) &
        subtle.ConstantTimeCompare(sk.key[:], other.key[:]) &
        subtle.ConstantTimeCompare(sk.tr[:], other.tr[:]))

    acc := uint32(0)
    for i := 0; i < L; i++ {
        for j := 0; j < common.N; j++ {
            acc |= sk.s1[i][j] ^ other.s1[i][j]       
        }                                             
    }
    for i := 0; i < K; i++ {          
        for j := 0; j < common.N; j++ {
            acc |= sk.s2[i][j] ^ other.s2[i][j]
            acc |= sk.t0[i][j] ^ other.t0[i][j]
        }                                                                     
    }
return (ret & subtle.ConstantTimeEq(int32(acc), 0)) == 1
```

The equivalent use to your proposal would be:

```
return subtle.ConstantTimeEq(int32(acc)(ret & subtle.ConstantTimeEq(int32(acc), 0)),1) == 1)
```

Regardless, my original point stands: the code generated between the current original in Kyber and your proposal is the same.

tanyav2:
> re: `tkn`, the matrix comparisons in that same line are also not constant time. I'll create an issue to assess the need and make all necessary changes together.

This is also bottlenecked by https://github.com/cloudflare/circl/issues/286 somewhat

pke/kyber/kyber768/internal/cpapke.go
tmthrgd: Huh? The equivalent code is right there: `subtle.ConstantTimeEq(int32(acc), 0)`, and the generated code looks different to me. The dilithium code could easily look like `return ret | acc != 0` which would be the same as what is currently done. Regardless I can remove it if you would like.

armfazh (approved):
it looks good, and doesn't add a significant overhead.

pke/kyber/kyber768/internal/cpapke.go
bwesterb: The code generated is different, but in both cases it's constant time. The reason `ConstantTimeEq` exists, is to be able to pass its return value into `ConstantTimeSelect`. Indeed: regular `==` between uint32, will give you the right bool in constant time, but you can't easily convert that `bool` into the integer required for `ConstantTimeSelect`. In other languages `bool`s are also dangerous because the compiler likes to get clever, but in Go that's less of a concern.

armfazh:
Thanks @tmthrgd

PR #416, opened 2023-03-14
Make ascon cipher go routine safe

### LINKED ISSUE #415: cipher/ascon: go routine safe?
A quick peak at the ascon implementation suggests that the cipher is unsafe to use across go routines:

https://github.com/cloudflare/circl/blob/278354d01755a901d6629247e464265218d9d120/cipher/ascon/ascon.go#L178

I believe the general intent is for a `cipher.AEAD` to be safely reusable across go routines to avoid the setup cost.

Either way, it would be good to at least document.

enj:
Fixes #415

cc @bwesterb @armfazh

### DIFF
--- a/cipher/ascon/ascon.go
+++ b/cipher/ascon/ascon.go
@@ -60,7 +60,6 @@ const (
 const permA = 12
 
 type Cipher struct {
-	s    [5]uint64
 	key  [3]uint64
 	mode Mode
 }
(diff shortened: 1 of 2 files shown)


armfazh (approved):

armfazh:
Thanks @enj  for pointing that out.

PR #422, opened 2023-04-11
Export HPKE ciphersuite algorithms

chris-wood:
Callers construct a Suite from a (KEM, AEAD, KDF) tuple and use it for things like creating senders and receivers, but then they must still keep the tuple elements around if they want to get access to things like the KEM Scheme or the AEAD key size. It would be simpler if the caller could just hang onto a Suite value and then access the KEM, AEAD, and KDF values internally. So this change exports things to make that happen.

### DIFF
--- a/hpke/aead.go
+++ b/hpke/aead.go
@@ -29,7 +29,7 @@ type (
 // greater than 255*N bytes, where N is the size (in bytes) of the KDF's
 // output.
 func (c *encdecContext) Export(exporterContext []byte, length uint) []byte {
-	maxLength := uint(255 * c.suite.kdfID.ExtractSize())
+	maxLength := uint(255 * c.suite.Kdf.ExtractSize())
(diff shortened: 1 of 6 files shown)


bwesterb (approved):

armfazh (commented):
One side-effect of making fields public is that the tuple can be modified into an invalid tuple. 
Recall the `NewSuite` method validates the tuple components.
It would be great to allow read-only access to the tuple elements.

armfazh:
> An alternative design would be to expose getters that return the internal values. That way they can't be modified with a reference to the struct itself. Would that work?

Yes, a function similar to 
https://github.com/cloudflare/circl/blob/main/hpke/util.go#L71
but that returns the objects rather than the identifiers.

chris-wood:
Indeed it does! I didn't know that function existed. Closing.

PR #427, opened 2023-04-14
hpke: omit sender key in xyber tests as it's unauthenticated

bwesterb:


bwesterb:
Nvm, I misunderstood the meaning of these entries.

PR #428, opened 2023-04-14
hpke: fix encapsulation seed in test for xyber

bwesterb:
Instead of providing an ephemeral keypair for test vectors,
for which Kyber does not have an analogue, we provide the encapsulation seed.


hpke/vectors_test.go
```diff
@@ -228,16 +228,17 @@ type vector struct {
 	KdfID              uint16             `json:"kdf_id"`
 	AeadID             uint16             `json:"aead_id"`
 	Info               string             `json:"info"`
+	Iem                string             `json:"iem,omitempty"`
 	IkmR               string             `json:"ikmR"`
-	IkmE               string             `json:"ikmE"`
+	IkmE               string             `json:"ikmE,omitempty"`
```
chris-wood, line 233:
```suggestion
```

hpke/vectors_test.go
```diff
@@ -228,16 +228,17 @@ type vector struct {
...
 	SkRm               string             `json:"skRm"`
-	SkEm               string             `json:"skEm"`
+	SkEm               string             `json:"skEm,omitempty"`
```
chris-wood, line 235:
```suggestion
```

hpke/vectors_test.go
```diff
@@ -228,16 +228,17 @@ type vector struct {
...
 	SkSm               string             `json:"skSm,omitempty"`
 	Psk                string             `json:"psk,omitempty"`
 	PskID              string             `json:"psk_id,omitempty"`
 	PkSm               string             `json:"pkSm,omitempty"`
 	PkRm               string             `json:"pkRm"`
-	PkEm               string             `json:"pkEm"`
+	PkEm               string             `json:"pkEm,omitempty"`
```
chris-wood, line 241:
```suggestion
```

hpke/vectors_test.go
```diff
@@ -228,16 +228,17 @@ type vector struct {
 	KdfID              uint16             `json:"kdf_id"`
 	AeadID             uint16             `json:"aead_id"`
 	Info               string             `json:"info"`
+	Iem                string             `json:"iem,omitempty"`
```
chris-wood, line 231: Is this "input encapsulation material"? Can we maybe rename it to "input encapsulation randomness" or "input encapsulation seed"?

hpke/vectors_test.go
```diff
@@ -389,6 +389,10 @@ func TestHybridKemRoundTrip(t *testing.T) {
 			panic("unsupported mode")
 		}
 
+		if rnd2.Len() != 0 {
+			t.Fatal()
+		}
+
```
chris-wood, line 395:
```suggestion
```

It doesn't seem like this is needed since we always create it as 64 bytes up above?

hpke/vectors_test.go
bwesterb: We can't, it's used elsewhere.

chris-wood: Where is it used?

hpke/vectors_test.go
bwesterb: Renamed to  "input encapsulation randomness"

hpke/vectors_test.go
bwesterb: This is defensive programming, to make sure we don't have a mismatch between the amount of randomness in the buffer and the actual amount consumed. It's not necessary now, but it might be once we check the upstream test vectors and copy-paste this code.

chris-wood (approved):

PR #434, opened 2023-05-01
Add HPKE benchmarks

chris-wood:
These benchmarks let one quickly compare the cost difference between two different HPKE ciphersuites (KEM, KDF, AEAD).

Here's a sample of the output when running the benchmarks (`go test -bench=.`) in the HPKE directory:

```
✗ go test -bench=.
goos: darwin
goarch: arm64
pkg: github.com/cloudflare/circl/hpke
BenchmarkHpkeRoundTrip/SetupSender-0020-0001-0001-10         	   18325	     65444 ns/op
BenchmarkHpkeRoundTrip/SetupReceiver-0020-0001-0001-10       	   28868	     41394 ns/op
BenchmarkHpkeRoundTrip/Encrypt-0020-0001-0001-10             	14436075	        84.96 ns/op
BenchmarkHpkeRoundTrip/Decrypt-0020-0001-0001-10             	14667714	        83.58 ns/op
BenchmarkHpkeRoundTrip/SetupSender-0030-0001-0001-10         	   13032	     91840 ns/op
BenchmarkHpkeRoundTrip/SetupReceiver-0030-0001-0001-10       	   15948	     75430 ns/op
BenchmarkHpkeRoundTrip/Encrypt-0030-0001-0001-10             	14285422	        83.86 ns/op
BenchmarkHpkeRoundTrip/Decrypt-0030-0001-0001-10             	14927398	        83.00 ns/op
PASS
ok  	github.com/cloudflare/circl/hpke	16.127s
```

### DIFF
--- a/hpke/hpke_test.go
+++ b/hpke/hpke_test.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"crypto/rand"
 	"fmt"
+	"testing"
 
 	"github.com/cloudflare/circl/hpke"
 )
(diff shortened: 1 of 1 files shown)


bwesterb (approved):
Bit ad hoc, but fine by me.

chris-wood:
> Bit ad hoc, but fine by me.

Is there a better approach we should take?

armfazh (approved):

PR #440, opened 2023-05-30
Set CIRCL version for generated assembler code.

armfazh:
This helps to remove our own security advisory.

@Lekensteyn could you please verify if the version of stdlib dependencies are ok, regarding our integration with cfgo.


Lekensteyn (approved):
LGTM.

The go.mod and go.sum changes are for `/asm` which are not imported directly in the main circl module, so it won't affect the minimum x/sys versions in vendored copies.

PR #445, opened 2023-06-13
Add partially blind RSA implementation

chris-wood:
See [the specification](https://datatracker.ietf.org/doc/html/draft-amjad-cfrg-partially-blind-rsa-00) for more information. This is needed to help experiment with a version of Privacy Pass that supports public metadata, described [here](https://github.com/smhendrickson/draft-hendrickson-privacypass-public-metadata-issuance).

This change removes some of the interfaces that were previously unused, like the generic `blindsign.Verifier` and `BlindSign.Signer` interfaces.

cc @Guss123


blindsign/blindrsa/pbrsa.go
```diff
@@ -209,7 +209,10 @@ func (v PBRSAVerifier) Blind(random io.Reader, message, metadata []byte) ([]byte
 
 	// Pick a random string rand of length 32 bytes
 	rand := make([]byte, 32)
-	random.Read(rand)
+	_, err = random.Read(rand)
```
bwesterb, line 212: I think I missed this last time, but you even need to use `io.FullRead`.

blindsign/blindrsa/brsa.go
```diff
@@ -58,61 +71,78 @@ func convertHashFunction(hash crypto.Hash) hash.Hash {
...
 		return nil, nil, err
 	}
 
 	if r.Sign() == 0 {
 		r = bigOne
 	}
-	rInv := new(big.Int).ModInverse(r, key.N)
+	rInv := new(big.Int).ModInverse(r, N)
 	if rInv == nil {
+		fmt.Println(r, N)
```
bwesterb, line 135: Stray debug statement.

blindsign/blindrsa/rsa.go
```diff
@@ -105,73 +104,45 @@ func decrypt(random io.Reader, priv *rsa.PrivateKey, c *big.Int) (m *big.Int, er
...
-			m2.Mul(m2, values.Coeff)
-			m2.Mod(m2, prime)
-			if m2.Sign() < 0 {
-				m2.Add(m2, prime)
-			}
-			m2.Mul(m2, values.R)
-			m.Add(m, m2)
-		}
-	}
+	m = new(big.Int).Exp(c, priv.d, priv.pk.N)
```
bwesterb, line 126: Why is the CRT path removed?

blindsign/blindrsa/pbrsa.go
```diff
@@ -0,0 +1,351 @@
...
+//
+// See the specification for more details:
+// https://datatracker.ietf.org/doc/html/draft-amjad-cfrg-partially-blind-rsa-00#name-blind
+func (v PBRSAVerifier) Blind(random io.Reader, message, metadata []byte) ([]byte, PBRSAVerifierState, error) {
+	if random == nil {
+		return nil, PBRSAVerifierState{}, ErrInvalidRandomness
+	}
+
+	salt := make([]byte, v.hash.Size())
+	_, err := random.Read(salt)
```
bwesterb, line 200: Ibidem.

blindsign/blindrsa/rsa.go
chris-wood: It was never actually used.

bwesterb:
This breaks API. In particular, it will break [this online example](https://medium.com/asecuritysite-when-bob-met-alice/cloudflare-circl-and-blinded-signatures-fdf9d3bb906b).

blindsign/blindrsa/brsa.go
```diff
@@ -45,6 +49,14 @@ type DeterminsiticRSAVerifier struct {
 	hash hash.Hash
 }
 
+type BRSAVerifier interface {
```
bwesterb, line 52: Blind is already in the name of the package, twice. I don't think we need to repeat "B" for blind. (Just like it's `big.Int` and not `big.BigInt`.)

blindsign/blindrsa/brsa.go
```diff
@@ -45,6 +49,14 @@ type DeterminsiticRSAVerifier struct {
...
+	Blind(random io.Reader, message []byte) ([]byte, BRSAVerifierState, error)
+	FixedBlind(message, blind, salt []byte) ([]byte, BRSAVerifierState, error)
+	Verify(message, signature []byte) error
+	Hash() hash.Hash
+	PublicKey() *rsa.PublicKey
+}
```
bwesterb, line 58: Please document the interface.

blindsign/blindrsa/pbrsa.go
```diff
@@ -0,0 +1,337 @@
...
+
+	"golang.org/x/crypto/hkdf"
+)
+
+// BigPublicKey is the same as an rsa.PublicKey struct, except the public
+// key is represented as a big integer as opposed to an int. For the partially
+// blind scheme, this is required since the public key will typically be
+// any value in the RSA group.
+type BigPublicKey struct {
+	N *big.Int
```
bwesterb, line 21: N is public, but e is private.

armfazh (changes requested):
I did one pass, will do another one once the suggested changes are addressed.

blindsign/blindsign.go
```diff
@@ -1,42 +0,0 @@
-// Package blindsign provides a blind signature protocol.
```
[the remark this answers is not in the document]
chris-wood: So it would be a file without any code?

chris-wood:
> This breaks the old API. In particular, it will break [this online example](https://medium.com/asecuritysite-when-bob-met-alice/cloudflare-circl-and-blinded-signatures-fdf9d3bb906b). Do we need to break it?

I suppose not, but I don't think we need to worry about API stability for a library like this.

blindsign/blindrsa/brsa.go
```diff
@@ -141,34 +169,44 @@ func fixedBlind(message, salt []byte, r, rInv *big.Int, pk *rsa.PublicKey, hash
...
+	return opts.SaltLength
+}
+
+// FixedBlind runs the Blind function with fixed blind and salt inputs.
+func (v DeterminsiticBRSAVerifier) FixedBlind(message, blind, salt []byte) ([]byte, BRSAVerifierState, error) {
+	if blind == nil {
+		return nil, BRSAVerifierState{}, ErrInvalidRandomness
+	}
+
+	r := new(big.Int).SetBytes(blind)
```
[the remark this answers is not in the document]
chris-wood: `r = 0` will be caught by the `ModInverse` check. I'll add a check to make sure `r < N`.

blindsign/blindrsa/brsa.go
chris-wood: Done.

blindsign/blindrsa/pbrsa.go
chris-wood: Fixed to make both private.

blindsign/blindrsa/brsa.go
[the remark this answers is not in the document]
chris-wood: Fixed.

chris-wood:
@bwesterb @armfazh most specific changes have been addressed, but we have open the API compatibility question and stuff about naming.

blindsign/blindsign.go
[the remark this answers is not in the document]
armfazh: It's ok if the file is removed.
as an FYI, a file without code but with documentation, only renders the documentation of the package. See for example the path `/circl/abe/cpabe`.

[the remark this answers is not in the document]
chris-wood: ACK, good to know!

armfazh (commented):
I'm ok with refactoring the code (and breaking API) if it improves the code. So, here is my suggestion.

The package `blindrsa` already indicates it is a blind signature of RSA type. So we can remove some prefixes.

BRSASigner -> Signer
BRSAVerifier -> Verifier
BRSAVerifierState -> VerifierState

Make them unexported since there exist New<> functions for them:
DeterminsiticBRSAVerifier -> determinsiticBRSAVerifier
RandomBRSAVerifier -> randomBRSAVerifier
BigPrivateKey -> bigPrivateKey
BigPublicKey -> bigPublicKey 

* Coalesce `NewBRSAVerifier` and `NewDeterministicBRSAVerifier` in a single function `NewVerifier(pk,hash,flag) Verifier` with a boolean flag to choose whether is randomized or deterministic.

For Partially-blind RSA, my suggestion is to create a subpackage:

/circl/blindsign/blindrsa                   # blind RSA 
/circl/blindsign/blindrsa/partialblindrsa   # partial blind RSA 
/circl/blindsign/blindrsa/internal/         # common code

This ease the naming to have plain Signer, Verifier, etc. names.

blindsign/blindrsa/pbrsa.go
```diff
@@ -0,0 +1,347 @@
...
+
+func convertToCustomPublicKey(pk *rsa.PublicKey) *BigPublicKey {
+	return &BigPublicKey{
+		n: pk.N,
+		e: new(big.Int).SetInt64(int64(pk.E)),
+	}
+}
+
+func convertToCustomPrivateKey(sk *rsa.PrivateKey) *BigPrivateKey {
+	return &BigPrivateKey{
```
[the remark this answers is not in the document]
chris-wood: Do you mean it's not _implemented_? The specification [is here](https://datatracker.ietf.org/doc/html/draft-amjad-cfrg-partially-blind-rsa-00#name-key-generation-2).

chris-wood:
Thanks @armfazh -- I enacted all your suggestions except for the one to merge constructors. I think it's better to avoid constructors with boolean flags. For the other comments regarding key generation and whatnot, we could implement that in this change or separately. Do you have a preference?

blindsign/blindrsa/common.go
```diff
@@ -0,0 +1,121 @@
...
+	switch hash {
+	case crypto.SHA256:
+		return sha256.New()
+	case crypto.SHA384:
+		return sha512.New384()
+	case crypto.SHA512:
+		return sha512.New()
+	default:
+		panic(errUnsupportedHashFunction)
+	}
```
armfazh, line 29: Since every `crypto.Hash` has a `New` method to get a `hash.Hash`, this code can be simplified

```suggestion
	switch hash {
	case crypto.SHA256, crypto.SHA384, crypto.SHA512:
		return hash.New()
	default:
		panic(errUnsupportedHashFunction)
	}
```

blindsign/blindrsa/pbrsa.go
```diff
@@ -0,0 +1,347 @@
...
+}
+
+func convertToCustomPublicKey(pk *rsa.PublicKey) *BigPublicKey {
+	return &BigPublicKey{
+		n: pk.N,
+		e: new(big.Int).SetInt64(int64(pk.E)),
+	}
+}
+
+func convertToCustomPrivateKey(sk *rsa.PrivateKey) *BigPrivateKey {
```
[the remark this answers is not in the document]
chris-wood: I added a check in the PBRSA constructor function.

blindsign/blindrsa/brsa.go
```diff
@@ -141,34 +128,47 @@ func fixedBlind(message, salt []byte, r, rInv *big.Int, pk *rsa.PublicKey, hash
...
-	})
-	return err
+// FixedBlind runs the Blind function with fixed blind and salt inputs.
+func (v determinsiticBRSAVerifier) FixedBlind(message, blind, salt []byte) ([]byte, VerifierState, error) {
+	if blind == nil {
+		return nil, VerifierState{}, ErrInvalidRandomness
+	}
+
+	r := new(big.Int).SetBytes(blind)
+	if r.Cmp(v.pk.N) < 0 {
```
armfazh, line 158: condition is reversed, pls double check.

blindsign/blindrsa/brsa.go
```diff
@@ -177,48 +177,48 @@ func (v DeterminsiticRSAVerifier) Verify(message, signature []byte) error {
...
 
 // FixedBlind runs the Blind function with fixed blind and salt inputs.
-func (v RSAVerifier) FixedBlind(message, blind, salt []byte) ([]byte, blindsign.VerifierState, error) {
+func (v randomBRSAVerifier) FixedBlind(message, blind, salt []byte) ([]byte, VerifierState, error) {
 	if blind == nil {
-		return nil, nil, ErrInvalidRandomness
+		return nil, VerifierState{}, ErrInvalidRandomness
 	}
 
 	r := new(big.Int).SetBytes(blind)
```
armfazh, line 205: the value of `r` is not bound-checked.

armfazh (commented):
adding some comments to blindrsa, still need to review partialblind

blindsign/blindrsa/common.go
chris-wood: Reverting this change as it doesn't work:

```
panic: crypto: requested hash function #6 is unavailable [recovered]
	panic: crypto: requested hash function #6 is unavailable
```

blindsign/blindrsa/brsa.go
```diff
@@ -177,48 +177,48 @@ func (v DeterminsiticRSAVerifier) Verify(message, signature []byte) error {
...
-		return nil, nil, err
+		return nil, VerifierState{}, err
 	}
 
 	return fixedBlind(message, salt, r, rInv, v.pk, v.hash)
 }
 
 // FixedBlind runs the Blind function with fixed blind and salt inputs.
-func (v RSAVerifier) FixedBlind(message, blind, salt []byte) ([]byte, blindsign.VerifierState, error) {
+func (v randomBRSAVerifier) FixedBlind(message, blind, salt []byte) ([]byte, VerifierState, error) {
```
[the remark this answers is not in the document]
chris-wood: No, the salt length doesn't need to be checked.

blindsign/blindrsa/brsa.go
chris-wood: Fixed.

blindsign/blindrsa/brsa.go
chris-wood: Good catch! Fixed.

blindsign/blindrsa/common.go
```diff
@@ -0,0 +1,121 @@
...
+	rInv := new(big.Int).ModInverse(r, N)
+	if rInv == nil {
+		return nil, nil, ErrInvalidBlind
+	}
+
+	return r, rInv, nil
+}
+
+// VerifyMessageSignature verifies the input message signature against the expected public key
+func VerifyMessageSignature(message, signature []byte, saltLength int, pk *keys.BigPublicKey, hash crypto.Hash) error {
```
armfazh, line 67: there are some exported functions that require keys from `/internal/keys` package, which is an internal package.
Thus, these keys cannot be instantiated in other packages (unless providing a generator function).

Can we reconsider to unexport, for example, `VerifyMessageSignature`, since we already have two constructors to get a Verifier?

blindsign/blindrsa/partiallyblindrsa/pbrsa_test.go
```diff
@@ -0,0 +1,314 @@
...
+	if err != nil {
+		t.Fatalf("Error producing test vectors: %v", err)
+	}
+
+	// TODO(caw): verify that we process them correctly
+	// verifyPBRSATestVectors(t, encoded)
+
+	var outputFile string
+	if outputFile = os.Getenv(pbrsaTestVectorOutEnvironmentKey); len(outputFile) > 0 {
+		err := os.WriteFile(outputFile, encoded, 0600)
```
armfazh, line 260:
```suggestion
		err := os.WriteFile(outputFile, encoded, 0o600)
```

blindsign/blindrsa/partiallyblindrsa/pbrsa.go
```diff
@@ -0,0 +1,319 @@
...
+
+	// Identifier of the cryptographic hash function used in producing the message signature
+	cryptoHash crypto.Hash
+
+	// Hash function used in producing the message signature
+	hash hash.Hash
+}
+
+// RandomizedPBRSAVerifier creates a new PBRSAVerifier using the corresponding Signer parameters.
+func NewRandomizedPBRSAVerifier(pk *rsa.PublicKey, hash crypto.Hash) Verifier {
```
armfazh, line 43:
```suggestion
// NewVerifier creates a new Verifier using the corresponding Signer parameters.
func NewVerifier(pk *rsa.PublicKey, hash crypto.Hash) Verifier {
```

blindsign/blindrsa/partiallyblindrsa/pbrsa.go
```diff
@@ -0,0 +1,319 @@
...
+
+// augmentPublicKey tweaks the public key based on the input metadata.
+//
+// See the specification for more details:
+// https://datatracker.ietf.org/doc/html/draft-amjad-cfrg-partially-blind-rsa-00#name-public-key-augmentation
+//
+// See the following issue for more discussion on HKDF vs hash-to-field:
+// https://github.com/cfrg/draft-irtf-cfrg-hash-to-curve/issues/202
+func augmentPublicKey(h crypto.Hash, pk *keys.BigPublicKey, metadata []byte) *keys.BigPublicKey {
+	// expandLen = ceil((ceil(log2(\lambda)) + k) / 8), where k is the security parameter of the suite (e.g., k = 128).
```
armfazh, line 60:
```suggestion
	// expandLen = ceil((ceil(log2(\lambda)/2) + k) / 8), where k is the security parameter of the suite (e.g., k = 128).
```

blindsign/blindrsa/partiallyblindrsa/pbrsa.go
```diff
@@ -0,0 +1,319 @@
...
+//
+// See the specification for more details:
+// https://datatracker.ietf.org/doc/html/draft-amjad-cfrg-partially-blind-rsa-00#name-verification-2
+func (v randomizedVerifier) Verify(message, metadata, signature []byte) error {
+	metadataKey := augmentPublicKey(v.cryptoHash, v.pk, metadata)
+	inputMsg := encodeMessageMetadata(message, metadata)
+	return blindrsa.VerifyMessageSignature(inputMsg, signature, v.hash.Size(), metadataKey, v.cryptoHash)
+}
+
+// Hash returns the hash function associated with the PBRSAVerifier.
```
armfazh, line 197:
```suggestion
// Hash returns the hash function associated with the Verifier.
```

blindsign/blindrsa/common.go
armfazh: must keep the imports to the packages.
```go
import _ "crypto/sha256"
import _ "crypto/sha512"
```

armfazh (commented):
few minor changes only, otherwise lgtm

chris-wood:
@armfazh thanks again for the suggestions -- I applied them.

blindsign/blindrsa/common.go
chris-wood: > Can we reconsider to unexport, for example, VerifyMessageSignature, since we already have two constructors to get a Verifier?

I'm not sure I follow this. Can you rephrase?

armfazh: Currently there are three methods to verify a signature:
1) building a `NewVerifier`, and calling its `Verify` method.
2) building a `NewDeterministicVerifier`, and calling its `Verify` method.
3) using the `VerifyMessageSignature` function.

I see redundancy on having so many functions that do the same thing (unless they **behave** differently).
It seems to me that 3) should not be exposed and moved to an internal package.

chris-wood: Got it -- thanks for clarifying. I'll try to work things out such that (3) is no longer exported.

armfazh (approved):

PR #450, opened 2023-07-18
tss/rsa: key generation for threshold RSA (safe primes)

armfazh:
Includes a method that generates RSA keys where the modulus is the product of two safe prime numbers.

### DIFF
--- a/math/primes.go
+++ b/math/primes.go
@@ -1,0 +1,34 @@
+package math
+
+import (
+	"crypto/rand"
+	"io"
+	"math/big"
+)
+
+// IsSafePrime reports whether p is (probably) a safe prime.
+// The prime p=2*q+1 is safe prime if both p and q are primes.
+// Note that ProbablyPrime is not suitable for judging primes
+// that an adversary may have crafted to fool the test.
+func IsSafePrime(p *big.Int) bool {
+	pdiv2 := new(big.Int).Rsh(p, 1)
+	return p.ProbablyPrime(20) && pdiv2.ProbablyPrime(20)
+}
+
+// SafePrime returns a number of the given bit length that is a safe prime with high probability.
+// The number returned p=2*q+1 is a safe prime if both p and q are primes.
+// SafePrime will return error for any error returned by rand.Read or if bits < 2.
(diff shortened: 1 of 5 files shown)


math/primes.go
```diff
@@ -0,0 +1,32 @@
...
+
+import (
+	"crypto/rand"
+	"io"
+	"math/big"
+)
+
+// IsSafePrime reports whether p is (probably) a safe prime.
+// The prime p=2*q+1 is safe prime if both p and q are primes.
+// The n parameter is the number of checks for primality test.
```
bwesterb, line 11: Good to add the caveat from `ProbablyPrime`:

```
ProbablyPrime is not suitable for judging primes that an adversary may
have crafted to fool the test.
````

File: math/primes.go [hunk 2]
```diff
@@ -0,0 +1,32 @@
...
+func IsSafePrime(p *big.Int, n int) bool {
```
bwesterb, line 12: Why make `n` configurable? Perhaps best to just use 20, as [crypto/rand.Prime](https://pkg.go.dev/crypto/rand#Prime) does.

math/primes.go
```diff
@@ -0,0 +1,32 @@
...
+	pdiv2 := new(big.Int).Rsh(p, 1)
+	return p.ProbablyPrime(n) && pdiv2.ProbablyPrime(n)
+}
+
+// SafePrime returns a number of the given bit length that is a safe prime with high probability.
+// The number returned p=2*q+1 is a safe prime if both p and q are primes.
+// SafePrime will return error for any error returned by rand.Read or if bits < 2.
+func SafePrime(random io.Reader, bits int) (*big.Int, error) {
```
bwesterb, line 20: I think it's faster to find a prime `q`, and then check whether `2q+1` is prime.

bwesterb (approved):
I left some suggestions.

math/primes.go
armfazh: added

math/primes.go
armfazh: removed the parameter.

math/primes.go
armfazh: true, changed

PR #451, opened 2023-07-18
zk/dleq: Adding DLEQ proofs for Qn, the subgroup of squares in (Z/nZ)*

armfazh:
This is required to enable verification of shares in threshold RSA.

### DIFF
--- a/zk/qndleq/qndleq.go
+++ b/zk/qndleq/qndleq.go
@@ -1,0 +1,133 @@
+// Package qndleq provides zero-knowledge proofs of Discrete-Logarithm Equivalence (DLEQ) on Qn.
+//
+// This package implements proofs on the group Qn (the subgroup of squares in (Z/nZ)*).
+//
+// # Notation
+//
+//	Z/nZ is the ring of integers modulo N.
+//	(Z/nZ)* is the multiplicative group of Z/nZ, a.k.a. the units of Z/nZ, the elements with inverse mod N.
+//	Qn is the subgroup of squares in (Z/nZ)*.
+//
+// A number x belongs to Qn if
+//
+//	gcd(x, N) = 1, and
+//	exists y such that x = y^2 mod N.
+//
+// # References
+//
+// [DLEQ Proof] "Wallet databases with observers" by Chaum-Pedersen.
+// https://doi.org/10.1007/3-540-48071-4_7
+//
+// [Qn] "Practical Threshold Signatures" by Shoup.
+// https://www.iacr.org/archive/eurocrypt2000/1807/18070209-new.pdf
+package qndleq
+
+import (
+	"crypto/rand"
+	"io"
+	"math/big"
+
+	"github.com/cloudflare/circl/internal/sha3"
+)
+
+type Proof struct {
+	Z, C     *big.Int
+	SecParam uint
+}
+
+// SampleQn returns an element of Qn (the subgroup of squares in (Z/nZ)*).
+// SampleQn will return error for any error returned by crypto/rand.Int.
+func SampleQn(random io.Reader, N *big.Int) (*big.Int, error) {
+	one := big.NewInt(1)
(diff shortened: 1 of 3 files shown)


zk/qndleq/qndleq.go
```diff
@@ -0,0 +1,144 @@
...
+)
+
+type Proof struct {
+	Z, C     *big.Int
+	SecParam uint
+}
+
+// SampleQn returns an element of Qn (the subgroup of squares in (Z/nZ)*).
+// SampleQn will return error for any error returned by crypto/rand.Int.
+func SampleQn(random io.Reader, N *big.Int) (*big.Int, error) {
```
bwesterb, line 40: Why is this public?

zk/qndleq/qndleq.go
```diff
@@ -0,0 +1,144 @@
...
+	}
+	r := new(big.Int).SetBytes(rBytes)
+
+	gP := new(big.Int).Exp(g, r, N)
+	hP := new(big.Int).Exp(h, r, N)
+
+	c := doChallenge(g, gx, h, hx, gP, hP, N, secParam)
+	z := new(big.Int)
+	z.Mul(c, x).Add(z, r)
+	r.Xor(r, r)
```
bwesterb, line 85: Why?

zk/qndleq/qndleq.go
```diff
@@ -0,0 +1,144 @@
...
+	hP := hPNum.Mul(hPNum, hPDen)
+	hP.Mod(hP, N)
+
+	c := doChallenge(g, gx, h, hx, gP, hP, N, p.SecParam)
+
+	return p.C.Cmp(c) == 0
+}
+
+func mustWrite(w io.Writer, b []byte) {
+	n, err := w.Write(b)
```
bwesterb, line 116: Write on Shake128 never errors.

zk/qndleq/qndleq.go
```diff
@@ -0,0 +1,144 @@
...
+
+// Prove creates a DLEQ Proof that attests that the pairs (g,gx)
+// and (h,hx) have the same discrete logarithm equal to x.
+//
+// Given g, h in Qn (the subgroup of squares in (Z/nZ)*), it holds
+//
+//	gx = g^x mod N
+//	hx = h^x mod N
+//	x  = Log_g(g^x) = Log_h(h^x)
+func Prove(random io.Reader, x, g, gx, h, hx, N *big.Int, secParam uint) (*Proof, error) {
```
bwesterb, line 68: We must add a big warning that this is not constant-time and might leak the secret by timing.

bwesterb (changes requested):
See comments.

zk/qndleq/qndleq.go
armfazh: it will be used when RSA has safe primes, see
https://github.com/cloudflare/circl/pull/453/files#diff-af5195e2923cc828aa543b4ac74c5b9b62c8c2ba072e70b4d07d4e74f698748cR154

zk/qndleq/qndleq.go
armfazh: it destroys the r value. it may not needed

zk/qndleq/qndleq.go
armfazh: true, but linter can compliant

zk/qndleq/qndleq.go
armfazh: done

zk/qndleq/qndleq.go
armfazh: removed now

zk/qndleq/qndleq.go
bwesterb: `_, _ = w.Write(b)` will work.

bwesterb (approved):
One little nit, but otherwise it's ok.

PR #453, opened 2023-07-19
tss/rsa: Adding verification of signature shares

armfazh:
This enables verification of signature shares, provided safe primes are used in the RSA key.

Reference: Protocol 1 of https://www.iacr.org/archive/eurocrypt2000/1807/18070209-new.pdf

Stacked on top of #451


tss/rsa/rsa_threshold.go
```diff
@@ -143,18 +150,32 @@ func Deal(randSource io.Reader, players, threshold uint, key *rsa.PrivateKey, ca
...
 	for i := uint(1); i <= players; i++ {
 		shares[i-1].Players = players
 		shares[i-1].Threshold = threshold
 		// Σ^{k-1}_{i=0} | a_i * X^i (mod m)
-		poly := computePolynomial(threshold, a, i, &m)
-		shares[i-1].si = poly
+		si := computePolynomial(threshold, a, i, &m)
+		shares[i-1].si = si
 		shares[i-1].Index = i
-		if cache {
```
jbis9051, line 156: nit: If caching is no longer optional, documentation should be update here https://github.com/cloudflare/circl/blob/tssRSAverif/tss/rsa/keyshare.go#L30

tss/rsa/keyshare.go
```diff
@@ -21,6 +32,13 @@ type KeyShare struct {
 
 	Players   uint
 	Threshold uint
+
+	// It stores keys to produce verifiable signature shares.
+	// If it's nil, signature shares are still produced but
+	// they are not verifiable.
+	// This field is present only if the RSA private key is
+	// composed of two safe primes.
+	vk *VerifyKeys
```
jbis9051, line 41: This should be added to the MarshalBinary/UnmarshalBinary functions https://github.com/cloudflare/circl/blob/tssRSAverif/tss/rsa/keyshare.go#L51

tss/rsa/rsa_threshold.go
jbis9051: also the if check should be removed here: https://github.com/cloudflare/circl/blob/5dcac578945a71dc912e27babc46148672c9e61c/tss/rsa/keyshare.go#L73C2-L73C2

tss/rsa/keyshare.go
armfazh: to ease review, I will propose changes to the serialization functions in a different PR.

tss/rsa/rsa_threshold.go
armfazh: done

PR #454, opened 2023-08-03
tss/rsa: Rewrite serialization functions,

armfazh:
Use cryptoybyte to leverage TLS syntax. Serialization of KeyShare and SignShares.

Staked on top of #453

### DIFF
--- a/internal/conv/conv.go
+++ b/internal/conv/conv.go
@@ -5,6 +5,8 @@ import (
 	"fmt"
 	"math/big"
 	"strings"
+
+	"golang.org/x/crypto/cryptobyte"
 )
 
 // BytesLe2Hex returns an hexadecimal string of a number stored in a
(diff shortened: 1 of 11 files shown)


internal/conv/conv.go
```diff
@@ -138,3 +140,36 @@ func BigInt2Uint64Le(z []uint64, x *big.Int) {
...
+
+// A UnmarshalingValue decodes itself from a cryptobyte.String and advances the pointer.
+// Returns true indicating the reading was successful.
+type UnmarshalingValue interface {
+	ReadValue(*cryptobyte.String) bool
+}
+
+// UnmarshalBinary recovers a value from a byte array.
+// Returns an error if the recovered value is invalid.
+// Any panic raised when calling to ReadValue is recovered and an error is returned instead.
```
bwesterb, line 159: Why do we need this?

internal/conv/conv.go
```diff
@@ -138,3 +140,36 @@ func BigInt2Uint64Le(z []uint64, x *big.Int) {
...
+		if r != nil {
+			err = fmt.Errorf("%T failed to unmarshal: %v", v, r)
+		}
+	}()
+
+	r := cryptobyte.String(data)
+	ok := v.ReadValue(&r)
+	if !ok {
+		return fmt.Errorf("cannot read %T from input string", v)
+	}
```
bwesterb, line 172: What if there are trailing bytes in `data`?

bwesterb (approved):

bwesterb (changes requested):
See comments in previous review.

internal/conv/conv.go
armfazh: this helper function generically implements UnmarshalBinary if there is an implementation of ReadValue (which internally uses the `cryptobyte` package).

internal/conv/conv.go
armfazh: it consumes only the data needed, so this allows to do reading continuations in composed data structures. User can call `Empty()` to check for trailing data.

PR #461, opened 2023-10-16
Draft BBS implementation

chris-wood:
Starting as a draft to collect early feedback. I've not checked against all of the test vectors [in the draft](https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-bbs-signatures-03), but most of the positive cases. I'm interested in feedback on the API at this point. Function internals will be improved in the future PR. Please ignore all of the random comments and notes for myself -- a lot of them are notes to send feedback to the CFRG draft for improvement.

This branch is necessary to support prototype work for [a Privacy Pass token type built on BBS](https://datatracker.ietf.org/doc/html/draft-ladd-privacypass-bbs-00).

cc @wbl, @tplooker

### DIFF
--- a/bbs/bbs.go
+++ b/bbs/bbs.go
@@ -1,0 +1,684 @@
+package bbs
+
+import (
+	"crypto"
+	"crypto/rand"
+	"encoding/binary"
+	"encoding/hex"
+	"fmt"
+	"math/big"
+
+	"github.com/cloudflare/circl/ecc/bls12381"
+	pairing "github.com/cloudflare/circl/ecc/bls12381"
+	"github.com/cloudflare/circl/expander"
+)
+
+var (
+	ciphersuiteID = []byte{0x00, 0x01}
+)
+
+func ciphersuiteString(suffix string) []byte {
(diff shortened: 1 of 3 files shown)


bbs/bbs.go
```diff
@@ -0,0 +1,892 @@
...
+	enc, err := s.sk.MarshalBinary()
+	if err != nil {
+		panic(err)
+	}
+	return enc
+}
+
+// Key generation
+// https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-bbs-signatures-03#name-key-generation-operations
+func KeyGen(ikm []byte, keyInfo, keyDst []byte) (SecretKey, error) {
```
wbl, line 232: Given this first argument should be the output from crypto/rand.Reader, should there have an internal form for testing and then this function reads from the reader?

chris-wood: Yeah, accepting a crypto/rand.Reader instead of raw bytes is better here. I passed the raw seed in for now to expedite checking against the test vectors.

PR #464, opened 2023-10-25
blindrsa: updating blindrsa to be compliant with RFC9474

armfazh:
Updates file of test vectors to be compatible with those in [RFC9474](https://www.rfc-editor.org/info/rfc9474).

And updates all the links so they point to the RFC.


blindsign/blindrsa/brsa.go
```diff
@@ -1,13 +1,28 @@
-package blindrsa
-
-// This package implements the blind RSA protocol based on the CFRG specification:
-// https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-rsa-blind-signatures
+// Package blindrsa implements the RSA Blind Signature Protocol.
```
thibmeu, line 1:
```suggestion
// Package blindrsa implements the RSA Blind Signature Protocol as defined in [RFC9474].
```

blindsign/blindrsa/brsa.go
```diff
@@ -1,13 +1,28 @@
...
+//
+// The RSA Blind Signature protocol, also called RSABSSA
```
thibmeu, line 3:
```suggestion
// The RSA Blind Signature protocol, and its variant RSABSSA
```

might be an overlook, but the RFC intro is the following
```
   This document specifies (1) a protocol for computing RSA blind
   signatures using RSA-PSS encoding and (2) a family of variants
   (Section 5) for this protocol, denoted RSABSSA (RSA Blind Signature
   with Appendix).  In order to facilitate deployment, it is defined in
   such a way that the resulting (unblinded) signature can be verified
   with a standard RSA-PSS library.
```

my understanding is RSA Blind Signature protocol is more general than RSABSSA, which is a category of variants for the protocol

blindsign/blindrsa/brsa_test.go
```diff
@@ -463,3 +494,48 @@ func BenchmarkBRSA(b *testing.B) {
...
+	pk := &sk.PublicKey
+	server := NewSigner(sk)
+
+	// Client: stores Server's public key.
+	verifier := NewVerifier(pk, crypto.SHA384)
+
+	// Protocol (online)
+
+	// Client blinds a message.
+	msg := []byte("alice and bob")
```
thibmeu, line 516: interestingly, having simpler test vectors like this in the spec could be beneficial. It's not going to be for this protocol as it's already RFC, but I sense it would be quite useful to have a test vector that fits in a simple example.

blindsign/blindrsa/brsa.go
```diff
@@ -98,6 +113,10 @@ func (v randomBRSAVerifier) Hash() hash.Hash {
 	return v.hash
 }
 
+func prepareMsg(message, prefix []byte) []byte {
+	return append(append([]byte{}, prefix...), message...)
```
thibmeu, line 117:
```suggestion
	return append(prefix, message...)
```

is there a specific reason for the append in append? the code above does the same and is more readable. you can check on the [go playground](https://play.golang.com/p/42WXhPQ7BTK).
```go
a := []byte{4, 5, 6}
b := []byte{1, 2, 3}
c := append(b, a...)
d := append(append([]byte{}, b...), a...)

fmt.Printf("a: %v\nb: %v\nc: %v\nd: %v\n", a, b, c, d)
/*
a: [4 5 6]
b: [1 2 3]
c: [1 2 3 4 5 6]
d: [1 2 3 4 5 6]
*/
```

thibmeu (approved):
only some minor nits. thanks for the update!

blindsign/blindrsa/brsa_test.go
armfazh: a good suggestion for those spec who hasn't been published (frozen) yet.

PR #465, opened 2023-10-27
Releasing CIRCL v1.3.6

armfazh:
Updating readme with new algorithms.
Bumping to v1.3.6 to jump branch tags.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,60 +25,114 @@ You can get CIRCL by fetching:
 go get -u github.com/cloudflare/circl
 ```
 
+Alternatively, look at the [Cloudflare Go](https://github.com/cloudflare/go/tree/cf) fork to see how to integrate CIRCL natively in Go.
+
 ## List of Algorithms
 
-#### Diffie-Hellman Protocol
-- [X25519](https://datatracker.ietf.org/doc/html/rfc7748/)
-- [X448](https://datatracker.ietf.org/doc/html/rfc7748/)
-- [Curve4Q](https://datatracker.ietf.org/doc/draft-ladd-cfrg-4q/)
+[RFC-7748]: https://doi.org/10.17487/RFC7748
+[RFC-8032]: https://doi.org/10.17487/RFC8032
+[RFC-8235]: https://doi.org/10.17487/RFC8235
+[RFC-9180]: https://doi.org/10.17487/RFC9180
+[RFC-9380]: https://doi.org/10.17487/RFC9380
+[RFC-9474]: https://doi.org/10.17487/RFC9474
+[RFC-9496]: https://doi.org/10.17487/RFC9496
+[RFC-9497]: https://doi.org/10.17487/RFC9497
+[FIPS 202]: https://doi.org/10.6028/NIST.FIPS.202
+[FIPS 186-5]: https://doi.org/10.6028/NIST.FIPS.186-5
+[BLS12-381]: https://electriccoin.co/blog/new-snark-curve/
+[ia.cr/2015/267]: https://ia.cr/2015/267
+[ia.cr/2019/966]: https://ia.cr/2019/966
 
-#### Digital Signature Schemes
-- [Ed25519](https://datatracker.ietf.org/doc/rfc8032/)
-- [Ed448](https://datatracker.ietf.org/doc/rfc8032/)
+### Elliptic Curve Cryptography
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -25,60 +25,119 @@ You can get CIRCL by fetching:
 go get -u github.com/cloudflare/circl
 ```
 
+Alternatively, you can use CIRCL natively by using [Cloudflare Go](https://github.com/cloudflare/go/tree/cf) fork.
```
bwesterb, line 28: It doesn't contain all packages, does it?

README.md
```diff
@@ -25,60 +25,119 @@ You can get CIRCL by fetching:
...
+
+ - [CSIDH](./dh/csidh): Post-Quantum Commutative Group Action ([CSIDH](https://csidh.isogeny.org/)).
+ - [Kyber KEM](./kem/kyber): modes 512, 768, 1024 ([KYBER](https://pq-crystals.org/kyber/)).
+ - [FrodoKEM](./kem/frodo): modes 640-SHAKE. ([FrodoKEM](https://frodokem.org/))
+ - (**insecure, deprecated**) ~~[SIDH/SIKE](./kem/sike)~~: Supersingular Key Encapsulation with primes p434, p503, p751 ([SIKE](https://sike.org/)).
+
+| PKE: Public-Key Encryption |
+|:---:|
+
+ - [Kyber PKE](./pke/kyber): modes 512, 768, 1024 ([Kyber](https://pq-crystals.org/kyber/)).
```
bwesterb, line 91: We shouldn't list this, or add a warning that this is not secure for general purpose.

bwesterb (approved):

README.md
armfazh: initially, we cherry picked packages.
the latest configuration uses an explicit require in go.mod file. so the compiler resolves the set of packages used.

README.md
armfazh: ok, it's better removing it.

README.md
```diff
@@ -25,60 +25,119 @@ You can get CIRCL by fetching:
...
- - [DLEQ](./zk/dleq): Prove knowledge of the Discrete Logarithm Equality.
+| XOF: eXtendable Output Functions |
+|:---:|
+
+ - [SHAKE128 and SHAKE256](./xof) ([FIPS 202]).
+ - [BLAKE2X](./xof): BLAKE2XB and BLAKE2XS ([Blake2x](https://www.blake2.net/blake2x.pdf))
+ - [KangarooTwelve](./xof/k12): fast hashing based on Keccak-p. ([KangarooTwelve](https://keccak.team/kangarootwelve.html)).
+ - SIMD [Keccak](https://keccak.team/keccak_specs_summary.html) f1600 Permutation.
+
+| LWC: Lightweight Cyptography |
```
SulemanAhmadd, line 114:
```suggestion
| LWC: Lightweight Cryptography |
```

README.md
bwesterb: Check the `src/vendor/github.com/cloudflare/circl` folder of cfgo. It doesn't contain many packages such as `math/fp448` and such as they aren't used by cfgo itself.

README.md
```diff
@@ -25,60 +25,119 @@ You can get CIRCL by fetching:
...
+ - [OT](./ot/simot): Simplest Oblivious Transfer ([ia.cr/2015/267]).
+ - [Threshold RSA](./tss/rsa) Signatures ([Shoup Eurocrypt 2000](https://www.iacr.org/archive/eurocrypt2000/1807/18070209-new.pdf)).
+
+### Post-Quantum Cryptography
+
+| KEM: Key Encapsulation Methods |
+|:---:|
+
+ - [CSIDH](./dh/csidh): Post-Quantum Commutative Group Action ([CSIDH](https://csidh.isogeny.org/)).
+ - [Kyber KEM](./kem/kyber): modes 512, 768, 1024 ([KYBER](https://pq-crystals.org/kyber/)).
```
SulemanAhmadd, line 84: Should we name it to ML-KEM based on NIST's renaming?

SulemanAhmadd (approved):

README.md
bwesterb: It's not the same as the draft of ML-KEM. I think we might just add the final version of ML-KEM as a separate package, so we don't break early adopters (such as ourselves.)

PR #470, opened 2024-01-04
Add ML-KEM (FIPS 203).

bwesterb:
Implementation of [ML-KEM (FIPS 203)](https://nvlpubs.nist.gov/nistpubs/fips/nist.fips.203.pdf).

We keep Kyber around (for now) as it's currently widely deployed. Code differences between them are minimal anyway.

Changes from IPD:

1. Add test against [ACVP test vectors](https://github.com/usnistgov/ACVP-Server/tree/master/gen-val/json-files/ML-KEM-encapDecap-FIPS203) — <del>which [still have to be updated to match FIPS203](https://groups.google.com/a/list.nist.gov/g/pqc-forum/c/s_Wez9FanHw/m/YGB-zccRBgAJ).</del> <ins>are updated</ins>
2. Add domain separation in internal PKE key generation.
3. Add modulus check on public key.


kem/kyber/templates/pkg.templ.go
```diff
@@ -143,6 +151,9 @@ func (pk *PublicKey) EncapsulateTo(ct, ss []byte, seed []byte) {
 	// c = Kyber.CPAPKE.Enc(pk, m, r)
 	pk.pk.EncryptTo(ct, m[:], kr[32:])
```
Lekensteyn, line 152: While at it, maybe we can use a name for this constant. This matches the type of `K` in the Algorithm 16 from FIPS 203 ("derive shared secret K and randomness r") and is consistent with your other use of `kr[:SharedKeySize]` below.
```suggestion
	pk.pk.EncryptTo(ct, m[:], kr[SharedKeySize:])
```

kem/kyber/templates/pkg.templ.go
```diff
@@ -5,10 +5,14 @@
...
 
 // Package {{.Pkg}} implements the IND-CCA2 secure key encapsulation mechanism
-// {{.Name}}.CCAKEM as submitted to round 3 of the NIST PQC competition and
+{{ if .NIST }}
+// {{.KemName}} as submitted to round 3 of the NIST PQC competition and
 // described in
 //
 // https://pq-crystals.org/kyber/data/kyber-specification-round3.pdf
+{{ else }}
+// {{.KemName}} as defined in FIPS203.
```
Lekensteyn, line 14: Non-NIST Kyber768.CCAKEM is not defined in FIPS203.

```suggestion
// {{.KemName}} as defined in
// https://pq-crystals.org/kyber/data/kyber-specification-round3.pdf
```

Lekensteyn (commented):
Lots of nits, the only really "wrong" thing is a documentation comment.

LGTM otherwise!

kem/kyber/templates/pkg.templ.go
bwesterb: This is accidental, and that 32 is not the shared key size.

kem/kyber/templates/pkg.templ.go
bwesterb: Oops, swapped the branches around.

bwesterb:
> Lots of nits, the only really "wrong" thing is a documentation comment.

Thanks. Addressed.

kem/kyber/templates/pkg.templ.go
Lekensteyn: I am not following.

My understanding is that the first part of `kr` is the shared secret key. So `kr[SharedKeySize:]` would skip the shared secret key, and refer to the remaining part, `r`.

```
Output: shared key K ∈ B32
```
```
(K, r) ← G(m ∥ H(ek))           ▷ derive shared secret key K and randomness r
c ← K-PKE.Encrypt(ek, m, r)     ▷ encrypt m using K-PKE with randomness r
```

bwesterb: In Kyber the first 32 bytes does not contain the returned shared secret, but an intermediate key that happens to be 32 bytes as well.

Lekensteyn: In that case, what about replacing `kr[:SharedKeySize]` by `kr[:32]` to avoid that implication?

Lekensteyn (approved):

mvd-ows:
Any updates on this? Is there something that prevents it from being merged?

Our view is that the "draft" status will probably remain for some time to come. Hopefully we can see this merged before the draft label is removed.

bwesterb:
> Is there something that prevents it from being merged?

ML-KEM is not final and could well have a breaking change compared to the initial public draft.

> Our view is that the "draft" status will probably remain for some time to come.

We expect the final version of ML-KEM this year, and that could be as early as this month.

mvd-ows:
Thank you for those clarifications, and the quick reply.

kem/kyber/gen.go
```diff
@@ -7,8 +7,10 @@ package main
```
armfazh, line 7: Could you please move these files used to generate code to an `internal` folder.
```
/kem/internal/gen.go
/kem/internal/pkg.templ.go
```
Generated packages:
```
/kem/kyber*
/kem/mlkem*
```

bwesterb: > Could you please move these files used to generate code to an `internal` folder.

Shall we do that in a separate PR? It's not only Kyber that has a `gen.go` outside of an `internal`.

kem/mlkem/acvp_test.go
```diff
@@ -0,0 +1,270 @@
...
+				ek2, err := scheme.UnmarshalBinaryPublicKey(result.Ek)
+				if err != nil {
+					t.Fatalf("tc=%d: %v", test.TcID, err)
+				}
+				dk2, err := scheme.UnmarshalBinaryPrivateKey(result.Dk)
+				if err != nil {
+					t.Fatal(err)
+				}
+
+				if dk.Equal(dk2) {
```
Lekensteyn, line 165: Based on the error message, aren't these two conditions inverted?
```suggestion
				if !dk.Equal(dk2) {
```

bwesterb: Oof. Well spotted. Let me figure out why it fails...

bwesterb: Ok, apparently the ACVP test vectors [haven't been updated yet](https://groups.google.com/a/list.nist.gov/g/pqc-forum/c/s_Wez9FanHw/m/YGB-zccRBgAJ).

pke/kyber/templates/pkg.templ.go
```diff
@@ -133,6 +152,17 @@ func (pk *PublicKey) Unpack(buf []byte) {
 	(*internal.PublicKey)(pk).Unpack(buf)
 }
 
+// Unpacks pk from the given buffer.
+//
+// Returns an error if the buffer is not of the right size, or the public
+// key is not normalized.
+func (pk *PublicKey) UnpackMLKEM(buf []byte) error {
+	if len(buf) != PublicKeySize {
+		return kem.ErrPubKeySize
```
armfazh, line 161: The previous `Pack/Unpack` function panics instead of returning an error when the size does not match.

armfazh (approved):

PR #476, opened 2024-02-07
Reduce x/crypto and x/sys versions to match Go 1.21, bump GitHub Actions and Go versions

Lekensteyn:
* Bump GitHub Actions versions, use current Go 1.21/1.22 versions

    * Use latest macOS and Windows versions instead of referring to an
      outdated version that might be removed later. See the table at
      https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories
    * Use the `stable` and `oldstable` aliases instead of hard-coding
      specific point releases for `actions/setup-go`. The arm64 builder is
      using a Docker image, so use the `1.22` image tag. We can use the
      `latest` tag in the future to avoid having to update the version.
    * Bump golangci-lint to the latest version.
    * Bump `github/codeql-action` actions to v3 as v2 is deprecated:
      https://github.com/github/codeql-action?tab=readme-ov-file#supported-versions-of-the-codeql-action
    * Bump Go version in go.mod to the minimum Go version supported by Go.

* Reduce x/crypto and x/sys versions to match Go 1.21

    CIRCL does not necessarily depend on newer x/crypto and x/sys versions.
    Lower the version requirements to match Go 1.21.7, this avoids the need
    to upgrade stdlib dependencies when integrating circl.

    For reference, we drop these `golang.org/x/crypto` commits:

        $ git -C ../golang-crypto log --oneline 2e82bdd1719d..v0.17.0 $(go list -deps ./... | grep golang.org/x/crypto/ | sed 's,golang.org/x/crypto/,./,')
        $ git -C ../golang-crypto log --oneline 2e82bdd1719d..v0.17.0 ./blake2b ./blake2s ./hkdf ./internal/alias ./chacha20 ./internal/poly1305 ./chacha20poly1305 ./cryptobyte/asn1 ./cryptobyte ./sha3
        b8ffc16 blake2b: drop Go 1.6, Go 1.8 compatibility
        bda2f3f argon2: avoid clobbering BP
        a2edfb5 cryptobyte: fix ReadOptionalASN1Boolean
        df0bc9e chacha20poly1305: guard PSHUFB usage with GOAMD64_v2
        74c2ba9 crypto/hkdf: remove useless call to Reset
        8779cbd all: update go directive to 1.18
        ec07f4e chacha20: drop Go 1.10 compatibility for arm64
        3f0842a sha3: have ShakeHash extend hash.Hash
        e90f1e1 cryptobyte: add uint48 methods

    No commits are dropped for `golang.org/x/sys/cpu`.


.github/workflows/ci-actions.yml
```diff
@@ -13,18 +13,18 @@ jobs:
     runs-on: ubuntu-22.04
     strategy:
       matrix:
-        GOVER: ['1.21.1', '1.20.8']
+        GOVER: ['stable', 'oldstable']
```
bwesterb, line 16: Do we want upstream to dictate our CI? @armfazh

.github/workflows/ci-actions.yml
```diff
@@ -68,11 +68,11 @@ jobs:
...
     steps:
       - name: Checkout
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
       - name: Setup Go
-        uses: actions/setup-go@v3
+        uses: actions/setup-go@v5
         with:
-          go-version: '1.21.1'
+          go-version: 'stable'
```
bwesterb, line 75: same

.github/workflows/ci-actions.yml
Lekensteyn: The motivation for this is to use the latest Go versions supported by upstream. Go 1.20 won't receive any new updates anymore.

I understand if you want to pin the older version here, but that means frequent updates like this. It's a trade-off to be made.

bwesterb: Ok. We can revert if Armando disagrees.

bwesterb (approved):

Lekensteyn:
I am puzzled why I am seeing golangci-lint errors in CI. I can reproduce part of it locally:
```
$ docker run --rm -v $(pwd):/app -w /app golangci/golangci-lint:v1.51.2 golangci-lint run --config=./.etc/golangci.yml ./...
::error file=xof/xof.go,line=56,col=10::cannot use blake2xb{…} (value of type blake2xb) as XOF value in return statement: blake2xb does not implement XOF (missing method Read) (typecheck)
::error file=xof/xof.go,line=59,col=10::cannot use blake2xs{…} (value of type blake2xs) as XOF value in return statement: blake2xs does not implement XOF (missing method Read) (typecheck)
::error file=xof/xof.go,line=62,col=10::cannot use k12d10{…} (value of type k12d10) as XOF value in return statement: k12d10 does not implement XOF (missing method Read) (typecheck)
::error file=xof/xof.go,line=74,col=40::cannot use blake2xb{…} (value of type blake2xb) as XOF value in return statement: blake2xb does not implement XOF (missing method Read) (typecheck)
::error file=xof/xof.go,line=78,col=40::cannot use blake2xs{…} (value of type blake2xs) as XOF value in return statement: blake2xs does not implement XOF (missing method Read) (typecheck)
::error file=xof/xof.go,line=84,col=9::cannot use k12d10{…} (value of type k12d10) as XOF value in return statement: k12d10 does not implement XOF (missing method Read) (typecheck)
::error file=group/group_test.go,line=6,col=2::"fmt" imported and not used (typecheck)
::error file=group/hash_test.go,line=6,col=2::"fmt" imported and not used (typecheck)
```

But it goes away as soon as I `go mod vendor` 🤷 
Something to check after some sleep.

Lekensteyn:
I dropped the GitHub actions update change. If that fixes the build failure, I will open a new PR to try to update GitHub Actions and address golangci-lint issues.

armfazh:
I can update golanci-lint version later, as I have already started working on this locally.

.github/workflows/ci-actions.yml
armfazh: I think it's ok to explicitly maintain the two most recent versions, changes happen not so often. So, let's leave it as is.

armfazh (approved):

.github/workflows/ci-actions.yml
Lekensteyn: Do you have a preference for hardcoding the major version as opposed to using `stable` and `oldstable`?

Note that specifying `1.21` will install the latest patch version: https://github.com/actions/setup-go/blob/6c1fd22b67f7a7c42ad9a45c0f4197434035e429/__tests__/setup-go.test.ts#L433-L439

In the current situation on the `main` branch we even hardcoded the patch version which I think is unnecessary.

Lekensteyn:
I now understand why golangci-lint showed such weird errors. The golangci-lint v1.51.2 binary was built with Go 1.20.1 while the golangci-lint v1.56.0 binary was properly built with Go 1.22.0. So in order to be able to update to Go 1.22, a golangci-lint update is needed. That does not seem to be too complicated, so I can make extra changes to the GitHub Actions update.

.github/workflows/ci-actions.yml
armfazh: the style e.g. `1.21` should be enough.

Lekensteyn: This has been addressed in #477.

PR #478, opened 2024-02-10
Adding rule for constant values

armfazh:
- fix threshold for constant values.
 - go mod tidy inside the /asm folders.

### DIFF
--- a/.etc/golangci.yml
+++ b/.etc/golangci.yml
@@ -54,7 +54,7 @@ linters-settings:
   nestif:
     min-complexity: 6
   goconst:
-    ignore-tests: true
+    min-occurrences: 4
   govet:
     check-shadowing: true
   staticcheck:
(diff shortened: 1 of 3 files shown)


group/group_test.go
```diff
@@ -17,20 +17,34 @@ var allGroups = []group.Group{
 	group.Ristretto255,
 }
 
+const (
```
bwesterb, line 20: I think these changes have made the code worse.

bwesterb (approved):

group/group_test.go
Lekensteyn: My first thought as well, it looks worse now. The point of the const checker is to avoid logic errors due to typos, or prevent some form of repetition. That is not the case here, the labels are just used in tests and can be anything.

blindsign/blindrsa/brsa_test.go
```diff
@@ -17,52 +15,12 @@ import (
...
-		return nil, fmt.Errorf("PEM private key decoding failed")
-	}
-
-	privateKey, err := x509.ParsePKCS1PrivateKey(block.Bytes)
-	if err != nil {
-		return nil, err
-	}
-
-	return privateKey, nil
+	deterministicReader := xof.SHAKE128.New()
```
Lekensteyn, line 22: What was the reason to convert the existing code? Having hard-coded constant test secrets is not uncommon, and probably the simplest approach.

In the tests below we have static secrets as well.

group/group_test.go
mmcloughlin: I'm not sure the rule `goconst` is trying to enforce is actually a good idea in all cases.

Perhaps if you configure it with a larger `min-len` and `min-occurrences` then it's more reasonable?

https://golangci-lint.run/usage/linters/#goconst

armfazh: @mmcloughlin that is a good idea, adopting it.

blindsign/blindrsa/brsa_test.go
armfazh: that's true, leave it as is.

PR #479, opened 2024-02-14
blindrsa: add support for all variants of RFC9474

armfazh:
This package is now fully compliant with the [RFC-9474](https://www.rfc-editor.org/info/rfc9474) document, and supports all the variants.

API changes: see `func Example_blindrsa()` for the new API.


blindsign/blindrsa/brsa_test.go
```diff
@@ -503,41 +558,53 @@ func Example_blindrsa() {
...
+	if err != nil {
+		fmt.Printf("client failed to generate blinded message: %v", err)
 		return
 	}
 
 	// Server signs a blinded message, and produces a blinded signature.
 	blindedSignature, err := server.BlindSign(blindedMsg)
 	if err != nil {
-		fmt.Fprintf(os.Stderr, "server failed to sign: %v", err)
+		fmt.Printf("server failed to sign: %v", err)
```
thibmeu, line 594: why move the error from stderr to stdout?

blindsign/blindrsa/brsa_test.go
```diff
@@ -20,7 +19,7 @@ import (
 )
 
 // 2048-bit RSA private key
-var testPrivateKey = testingKey(`
+const testPrivateKey = `
 -----BEGIN RSA TESTING KEY-----
```
thibmeu, line 23: nits: you might want to use a test RSA key from RFC 9500 https://github.com/thibmeu/rfc9500?tab=readme-ov-file#rsa-keys

blindsign/blindrsa/brsa.go
```diff
@@ -28,271 +26,176 @@ import (
...
+	case SHA384PSSRandomized:
+		return "RSABSSA-SHA384-PSS-Randomized"
+	case SHA384PSSZeroRandomized:
+		return "RSABSSA-SHA384-PSSZero-Randomized"
+	case SHA384PSSDeterministic:
+		return "RSABSSA-SHA384-PSS-Deterministic"
+	case SHA384PSSZeroDeterministic:
+		return "RSABSSA-SHA384-PSSZero-Deterministic"
+	default:
+		panic(ErrInvalidVariant)
```
thibmeu, line 56: `panic` sounds brutal for calling `String`. I understand `Stringer` does now allow to return an error, however, what about `return "invalid RSABSSA"` instead?

thibmeu (commented):
change looks ok.
the change from Verifier to Client also reflects the spec better.

blindsign/blindrsa/brsa_test.go
armfazh: Because the Go tester program will check the output in stdout:
```go
// Output: Valid signature: true
```

blindsign/blindrsa/brsa.go
armfazh: done

blindsign/blindrsa/brsa_test.go
armfazh: good idea.

blindsign/blindrsa/testdata/test_rsa2048_rfc9500.json
thibmeu, line 1: consider renaming this file to `testRSA2048.rfc9500.pem`.

at least the extension will be correct
regarding the file name, this is because the RFC 9500 suggests [a name](https://www.rfc-editor.org/rfc/rfc9500#section-2.1-6)

thibmeu (approved):
minor modification on file name.

the pr looks good otherwise. it's great to see the transition to rfc 9474

blindsign/blindrsa/testdata/test_rsa2048_rfc9500.json
armfazh: sounds good, filename was updated

PR #481, opened 2024-02-17
Explicitly installs Go with version before CodeQL analysis.

armfazh:
Report from CodeQL

```
Newer Go version needed
[View workflow run](https://github.com/cloudflare/circl/actions/runs/7938355244)
Version v1.20.14 of Go is installed, but this is lower than v1.21 required by your project's go.mod. [Install a newer version of Go before analyzing your project](https://github.com/actions/setup-go#basic).
```


thibmeu (approved):
that's a good change indeed

Issue #482, opened 2024-02-29
Implement MAYO


ilway25:
[MAYO](https://pqmayo.org/about/) is a practical UOV variant with small key and signature sizes. It has also been submitted to Round 1 of NIST's Additional PQC Signatures.

We have finished a Go implementation, and plan to incorporate it into CIRCL.

PR #483, opened 2024-02-29
Implement Mayo

ilway25:
We chose to implement the newer version of MAYO proposed by the authors instead of the one submitted to NIST.

The authors proposed the change to the spec [here](https://eprint.iacr.org/2023/1683.pdf):
"Nibbling MAYO: Optimized Implementations for AVX2 and Cortex-M4" by Ward Beullens, Fabio Campos, Sofía Celi, Basil Hess, Matthias J. Kannwischer.

This pure Go code is written based on the tricks described in that paper and in their reference C code, specifically the [nibbling-mayo](https://github.com/PQCMayo/MAYO-C/tree/nibbling-mayo) branch.

It also passes the KAT tests.

Closes https://github.com/cloudflare/circl/issues/482


bwesterb:
Thank you for this. I will review over the coming days.

sign/mayo/mode1/internal/mayo.go
```diff
@@ -0,0 +1,728 @@
...
+	P3 := viewBytesAsUint64Slice(epk.p3[:])
+
+	// Note: the variable time approach is overall about 30% faster
+	// compute P * S^t = [ P1  P2 ] * [S1] = [P1*S1 + P2*S2]
+	//                   [  0  P3 ]   [S2]   [        P3*S2]
+	var pst [M * N * K / 16]uint64
+	// mulAddMMatXMatTrans(pst[:], P1, s[:], V, V, K, N, true)
+	// mulAddMMatXMatTrans(pst[:], P2, s[V:], V, O, K, N, false)
+	// mulAddMMatXMatTrans(pst[M*V*K/16:], P3, s[V:], O, O, K, N, true)
+	variableTime(pst[:], P1, P2, P3, s[:])
```
bwesterb, line 641: It's not clear at all how this part matches up with what is written in the spec. You need to give an overview how the computation matches up.

sign/mayo/mode1/internal/matrix.go
```diff
@@ -0,0 +1,289 @@
...
+				for p := 0; p < P; p++ {
+					out[P*pos+p] ^= in[P*(c*size+r)+p]
+				}
+			}
+			pos++
+		}
+	}
+}
+
+func variableTime(sps []uint64, p1 []uint64, p2 []uint64, p3 []uint64, s []uint8) {
```
bwesterb, line 157: This function needs a better name, and a description of what it does.

sign/mayo/mode1/internal/matrix.go
```diff
@@ -0,0 +1,289 @@
...
+
+	for i := 0; i < K*K; i++ {
+		aggregate(P, accumulator[i], sps[P*i:])
+	}
+}
+
+// p is always P, but is still kept to be consistent with other functions
+//
+//nolint:unparam
+func vecAddPacked(p int, in []uint64, acc []uint64) {
```
bwesterb, line 223: If `p` is always `P`, then just use `P` and not have the parameter.

sign/mayo/mode1/internal/matrix.go
```diff
@@ -0,0 +1,289 @@
...
+			}
+			pos++
+		}
+	}
+}
+
+func variableTime(sps []uint64, p1 []uint64, p2 []uint64, p3 []uint64, s []uint8) {
+	var accumulator [K * N][P * 16]uint64
+
+	// compute P * S^t = [ P1  P2 ] * [S1] = [P1*S1 + P2*S2]
```
bwesterb, line 160: Missing transpose on S1 and S2?

sign/mayo/mode1/internal/matrix.go
```diff
@@ -0,0 +1,289 @@
...
+					out[P*pos+p] ^= in[P*(c*size+r)+p]
+				}
+			}
+			pos++
+		}
+	}
+}
+
+func variableTime(sps []uint64, p1 []uint64, p2 []uint64, p3 []uint64, s []uint8) {
+	var accumulator [K * N][P * 16]uint64
```
bwesterb, line 158: Explain the strategy.

sign/mayo/mode1/internal/matrix.go
```diff
@@ -0,0 +1,289 @@
...
+// p is always P, but is still kept to be consistent with other functions
+//
+//nolint:unparam
+func vecAddPacked(p int, in []uint64, acc []uint64) {
+	for i := 0; i < p; i++ {
+		acc[i] ^= in[i]
+	}
+}
+
+func aggregate(p int, bins [P * 16]uint64, out []uint64) {
```
bwesterb, line 229: Explain what this function does.

bwesterb: This function is only called with `P`. Why bother with the parameter?

sign/mayo/mode1/internal/matrix.go
```diff
@@ -0,0 +1,289 @@
...
+// 	// vecMulAddPacked(p, in, 2, acc)
+
+// 	msb := uint64(0x8888888888888888)
+// 	for i := 0; i < p; i++ {
+// 		t := in[i] & msb
+// 		acc[i] ^= ((in[i] ^ t) << 1) ^ ((t >> 3) * 3)
+// 	}
+// }
+
+func vecMulAddPackedByInvX(p int, in []uint64, acc []uint64) {
```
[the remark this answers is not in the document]
bwesterb: Only ever called with `p=P`. Why bother with the parameter?

bwesterb (changes requested):
This is only a partial review. There is a lot to like in here, but it still needs work to be easier to review: you need to explain how things are encoded, computed, etc.

Also, did you implement this from scratch or did you translate an existing implementation?

sign/mayo/mode1/internal/matrix.go
```diff
@@ -0,0 +1,289 @@
...
+	// vecMulAddPacked(p, in, 9, acc)
+
+	lsb := uint64(0x1111111111111111)
+	for i := 0; i < p; i++ {
+		t := in[i] & lsb
+		acc[i] ^= ((in[i] ^ t) >> 1) ^ (t * 9)
+	}
+}
```
bwesterb, line 289: This is Figure 2 method 3 right? Wouldn't method 2 be faster as we're not using AVX2 instructions?

ilway25: Here we chose to multiply by x^-1 all the way through, unlike Method 3 which interleaves *x and *x^-1 which probably gives more parallelism on more complex CPUs. On my MacBook M1 Pro, Method 2 was not faster.

sign/mayo/mode1/internal/matrix.go
ilway25: Here a design choice was made so that this is consistent with other similar functions which can be sometimes called with `p != P`.

sign/mayo/internal/common/nist/drbg.go
```diff
@@ -0,0 +1,71 @@
+// Package nist implements helpers to generate NIST's Known Answer Tests (KATs).
```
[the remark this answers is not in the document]
ilway25: `github.com/cloudflare/circl/internal/nist` does not implement `io.Reader`, which was somehow used during the debug process when implementing MAYO. Can change back to this one.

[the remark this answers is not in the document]
ilway25: Please ignore the previous comment. I think it would be better if `github.com/cloudflare/circl/internal/nist` implements `io.Reader` because MAYO's sign takes a random source which is used during KAT tests.

sign/mayo/mode1/internal/mayo_test.go
```diff
@@ -0,0 +1,222 @@
...
...5aa33519a56d740cbab7b9ee13e1b8add763bd07a25455e44e60f664846e32eec329c46e0741b6649c34358905d9d01124c3c8ae14e3fb47cc6477ba8e63bd4a3937588251952f5232bb9999aef509414e11a36d110ea24bedc235c55ab6f80116b036845e291148135a32b0d44fe5a4e9ea9f3ae150e11c62ea91318129f79318d3506ba50db3dbe235427d405a5baa9e6192a17014dd19863087445eb4cc9b6164aab3d5af4e73bc3edd97c76dabba0bc26c43d70f7880a0f7c042465da77ae3f8a31fa65903d90797fc0b5deddf578d52940040276c9b415d880e3782e03a2965a6d3b265628c690bd1b096a355362eb751d528fbc6fa365d1cf3d39cf49d8296c28fe8eb52a31a07e877d1e57e91971cbbd14216db76a08da45258a1a801d0cb232a87280df4522329ecd9312e5489d90a39547735cbd11493d0d86389af40c1b9ec5a993d0cbc508da424392695e1cdf205a294122412fac5a2b4d6e0ed5eed7b4de18f0d4b81f746753fa8744ab845c18800befc926d3290531053bf2cecf23d1b6d7fa9fb8d4827e87abf574afe39a636dbb9fce48854c641981f236da01bd8a5901871891aac7fa1f02300101540e7b05665e5f19912ee898666b64bb3cef688577c7b5482e123c0e8fcc466302eca8fc02f35260569cc1d4b00662be924f188226dbc6086d4aaed1ff9c7604748a556d00f87823f9fafaccf752d645b76a5851723274737e4dd7d5d76fdb256e3f1ef733914eb36932eb1c4c666df436351fb817e7055fbfca2b3103fcec84233feaf3fefd600721d3ca1514ee26db6a608dd4333377a9e36dd2dbadb1651eb99d8e3b4fe6"},
```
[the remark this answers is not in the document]
ilway25: I do not quite agree with this suggestion, because
- we do not have an extra Mode layer like Dilithium does
- this is intended as the internal implementation unit test for each mayo variant, not integration test

Maybe just keep test vectors that belong to each mode in their respective module? (Then we need also to skip `mayo_test.go` in`gen.go`)

sign/mayo/mode1/internal/mayo.go
```diff
@@ -0,0 +1,728 @@
...
+	"io"
+	"unsafe"
+
+	"github.com/cloudflare/circl/internal/sha3"
+)
+
+type (
+	PrivateKey [PrivateKeySize]byte
+	PublicKey  [PublicKeySize]byte
+)
```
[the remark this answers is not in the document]
ilway25: We made the conclusion that key expansion of pk and sk should go with keygen but not sign or verify.

Does this mean the keygen will expand the pk and sk, even though under some circumstances one just wants to store these new keys without doing signing and verification?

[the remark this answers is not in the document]
bwesterb: > Does this mean the keygen will expand the pk and sk, even though under some circumstances one just wants to store these new keys without doing signing and verification? (expansion will be wasted)

Yes, that's a downside. Keygen is the rarest operation, so I think it's alright.

bwesterb:
> Thanks for pointing out things to consider. No, it is not written not from scratch. The code basically follows the thought process of the reference code.

You should add a comment acknowledging on which code you've based yours, and mention its license (and make sure it's compatible with Circl's license.)

armfazh:
> > Thanks for pointing out things to consider. No, it is not written not from scratch. The code basically follows the thought process of the reference code.
> 
> You should add a comment acknowledging on which code you've based yours, and mention its license (and make sure it's compatible with Circl's license.)

about the licensing, I think it suffices to add a line in the gen.go file something such as:

```
This implementation is a port from the C implementation of MAYO [link} distributed under <LICENSE>.
```

and there is no need to include the `NOTICE` and `LICENSE` files.

armfazh (commented):
These are a few comments, not a full review. In the meantime @bwesterb will help you polishing the code, and I will review again after that.

internal/nist/drbg.go
```diff
@@ -11,6 +12,8 @@ type DRBG struct {
 	v   [16]byte
 }
 
+var _ io.Reader = (*DRBG)(nil)
```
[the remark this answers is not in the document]
ilway25: Yes it can be remove, but this is intended as a static check that DRBG does implement io.Reader.

sign/mayo/example_test.go
```diff
@@ -0,0 +1,56 @@
...
+	// Checks whether a signature is correct
+	if !mode.Verify(pk2, msg, signature) {
+		panic("incorrect signature")
+	}
+
+	fmt.Printf("O.K.")
+
+	// Output:
+	// Supported modes: [MAYO_1 MAYO_2 MAYO_3 MAYO_5]
+	// O.K.
```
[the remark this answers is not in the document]
ilway25: This files is copied from Dilithium. Could you explain what is the intention behind the modification here?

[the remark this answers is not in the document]
armfazh: it's a suggestion to improve the code.

internal/nist/drbg.go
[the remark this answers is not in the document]
armfazh: this check can be done in a test file.

bwesterb:
@ilway25 Thank you so much for the quick changes on our preliminary review. I'll be travelling for another week. After that I'll sit down and continue the review. Thanks again.

sign/mayo/mode1/internal/matrix.go
```diff
@@ -0,0 +1,294 @@
...
+			}
+		}
+	}
+
+	for i := 0; i < K*K; i++ {
+		accumulate(P, accumulator[i], sps[P*i:])
+	}
+}
+
+func vecAddPacked(in []uint64, acc []uint64) {
```
bwesterb, line 222: I'm not sure it makes sense that this is a separate function.

bwesterb:
The code improved a lot. Still it needs better documentation on the actual arithmetic to make review easier.

sign/mayo/NOTICE
```diff
@@ -0,0 +1,4 @@
+Copyright 2022-2023 the MAYO team. All rights reserved.
```
cjpatton, line 1: Copyright may need to be help by the maintainers of the repo.

PR #489, opened 2024-03-14
hpke: Enforces passing a slice of exact size to UnmarshalBinary for KEM keys

### LINKED ISSUE #488: hpke: slice bounds out of range in shortKEM.UnmarshalBinaryPrivateKey
When passing a byte slice of size > 66 bytes to `shortKEM.UnmarshalBinaryPrivateKey` when using the `DHKEM(P-521, HKDF-SHA512)` KEM, the following panic is triggered:

```
panic: runtime error: slice bounds out of range [-1:]

goroutine 3014 [running]:
panic({0x78c980?, 0xc00050a090?})
	/usr/lib/go/src/runtime/panic.go:770 +0x132
github.com/cloudflare/circl/hpke.shortKEM.UnmarshalBinaryPrivateKey({{{0x12, {0x7aa461, 0x19}, 0x7}, {0x820178, 0xc000016240}}, {0x81fe98, 0x9ad8e0}}, {0xc0002800a0, 0x43, ...})
	/home/simon/src/circl/hpke/shortkem.go:87 +0x35b
```

armfazh:
hpke: Allows passing a larger buffer to `UnmarshalBinary` for KEM keys. 
It checks data passed is in the right boundaries.
Also adds a test showing the issue.

Fixes #488

### DIFF
--- a/hpke/kem_test.go
+++ b/hpke/kem_test.go
@@ -1,0 +1,63 @@
+package hpke_test
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/cloudflare/circl/hpke"
+	"github.com/cloudflare/circl/internal/test"
+)
+
+func TestKemKeysMarshal(t *testing.T) {
+	for _, kem := range []hpke.KEM{
+		hpke.KEM_P256_HKDF_SHA256,
+		hpke.KEM_P384_HKDF_SHA384,
+		hpke.KEM_P521_HKDF_SHA512,
+		hpke.KEM_X25519_HKDF_SHA256,
+		hpke.KEM_X448_HKDF_SHA512,
+		hpke.KEM_X25519_KYBER768_DRAFT00,
+	} {
+		checkIssue488(t, kem)
+	}
+}
+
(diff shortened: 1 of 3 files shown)


bwesterb:
Why should we accept a buffer that is too large? Wouldn't we miss certain bugs if we do?

armfazh:
Agree, I have changed the code, so now, it enforces passing a slice of the exact size when unmarshaling keys. Otherwise, it returns an error instead of panic, which is the actual bug reported in #488 .

karalabe:
I just shot myself in the foot with this. I assumed the key size is always checked (it is for other unmarshal methods). I had a flaw in parsing out a PEM key from a file and gave the unmarshal method the entire pem container, not just the raw bytes. No error, no nothing until I paired it up in prod with a Rust HPKE lib that rejected the crypto. That's when I found the unmarshal was doing junk all along. Would be very nice to get this in.

cjpatton:
Now fixed in https://github.com/cloudflare/circl/pull/627.

PR #492, opened 2024-03-26
Align to the `purego` build tag, removing `noasm` build tag

### LINKED ISSUE #491: Align on `purego` build tag.
golang/go#23172 proposes the de-facto `purego` build tag.

`circl` uses both `purego` _and_ `noasm` build tags, inconsistently.

If a user wants a `purego` build, the build should not include assembly.

Would `circl` be willing to align on `purego` meaning `noasm`?

For example, this is what the upstream golang `crypto` package does: https://go-review.googlesource.com/c/go/+/561935

Another option is to make `purego` and `noasm` equivalent in the code-base if `noasm` needs to be a backwards supported build flag.

I would be willing to contribute the patch.

mattyclarkson:
The `purego` build tag has been proposed[1] as the de-facto tag for pure Go builds. This is a superset of the `noasm` tag currently used.

This patch aligns the code base to toggle to pure Go implementation when `purego` is defined.

Verified that all assembly is excluded by doing:

    find . -name '*.s' -exec sed -i -e '$athis-is-not-valid-assembly 12345' {} \;
    go test --tags purego ./...

Closes #491

[1]: https://github.com/golang/go/issues/23172


mattyclarkson:
@armfazh would it be possible to get approval to run the CI on this PR? Thanks.

armfazh (approved):

armfazh:
Thanks @mattyclarkson for reporting.

mattyclarkson:
@armfazh would it be possible to get this in a `1.3.8` release? No pressure, we can patch locally. If there are other things in flight on mainline before the next release, happy to wait.

PR #494, opened 2024-04-04
ecc/bls12381: Ensures pairing operations don't overwrite their input

### LINKED ISSUE #493: In ecc/bls12381, Pairing operations seems racy
It seems there are internal functions in circl that are "racy", because they are passing arrays using pointers, reading from these arrays and writing to these same arrays, causing data races when multiple go routines are operating on the same points at the same time.

This was discovered in the drand/kyber repo, where we have recently accepted a PR to add circl support for BLS12-381, but sadly despite having a test for data races in pairings, we didn't catch it earlier. (we weren't using `-race` in CI by mistake).

To reproduce:
```
git clone https://github.com/drand/kyber.git
cd kyber
go test -race ./pairing/circl_bls12381/...
``` 

This should fail in the `TestRacePairings` test.

armfazh:
Issue: Pairing functions internally convert input points to affine, overwriting the input. This causes troubles in a multiple goroutine scenario.

Fix: Ensures pairing operations don't overwrite the input.
Added a test for checking input's integrity.

Fixes #493

### DIFF
--- a/ecc/bls12381/g1.go
+++ b/ecc/bls12381/g1.go
@@ -392,7 +392,8 @@ func G1Generator() *G1 {
 }
 
 // affinize converts an entire slice to affine at once
-func affinize(points []*G1) {
+func affinize(points []*G1) (out []G1) {
+	out = make([]G1, len(points))
 	if len(points) == 0 {
 		return
 	}
@@ -410,8 +411,9 @@ func affinize(points []*G1) {
 		zinv.Mul(w, &ws[i])
 		w.Mul(w, &points[i].z)
 
-		points[i].x.Mul(&points[i].x, zinv)
-		points[i].y.Mul(&points[i].y, zinv)
-		points[i].z.SetOne()
+		out[i].x.Mul(&points[i].x, zinv)
+		out[i].y.Mul(&points[i].y, zinv)
+		out[i].z.SetOne()
 	}
+	return
 }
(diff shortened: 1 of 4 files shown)


ecc/bls12381/g1.go
```diff
@@ -392,7 +392,8 @@ func G1Generator() *G1 {
 }
 
 // affinize converts an entire slice to affine at once
-func affinize(points []*G1) {
+func affinize(points []*G1) (out []G1) {
```
thibmeu, line 395: What about not declaring `out` as a return variable, but declaring it on the line below? My sense is there is one less allocation, and the code reads the same.
Another question is about returning `[] *G1`, but I understand motivations reason to stick to `[]G1`

thibmeu (approved):
Looks good. One minor nit on variable declaration.

Tests are passing locally as well.

ecc/bls12381/g1.go
bwesterb: To avoid allocations you could pass it as an argument.

bwesterb (approved):

ecc/bls12381/g1.go
armfazh: I've tried passing the `out` slice as a parameter, but still one allocation is needed.

ecc/bls12381/g1.go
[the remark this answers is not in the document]
armfazh: For points in G2, the Miller's loop only uses the points in read-only mode.

PR #495, opened 2024-04-12
FourQ ecc_mul_double implementation

LINCKODE:
Implemented ecc_mul_double as DoubleScalarMult in curve.go .
Please advise me if the code does not adhere to project standards, I'm somewhat new to Go. 😅


armfazh:
Hi @LINCKODE what is the use case for this function? and usually, the operation `kG+lP` is performed using a better algorithm than performing `kG`, `lP`, and then add these points.

LINCKODE:
Thanks for the reply!
This function is required for verifying Schnorr signatures over FourQ.
You can see how it is being used in [FourQLib's SchnorrQ_Verify function](https://github.com/microsoft/FourQlib/blob/1031567f23278e1135b35cc04e5d74c2ac88c029/FourQ_64bit_and_portable/schnorrq.c#L171).
I see that FourQLib defines another way of performing this calculation, but that requires the use of endomorphisms, which I am not familiar with.
If you know of another way of doing this, I'd be happy to give it a go.

LINCKODE:
That looks interesting, doing both multiplications in one loop.
Ill have to look at this a bit closer to fully understand how it works, then I'll try to replicate it for FourQ.
Thanks for the hint!

PR #507, opened 2024-08-19
Add ML-KEM decapsulation key check.

bwesterb:
Described in section 7.3 of FIPS 203.

The check is only required if the private key is from an untrusted source. We do not distinguish between a trusted and untrusted source in the current API, so we'll perform the check every time we unmarshal the private key.

### DIFF
--- a/kem/kyber/templates/pkg.templ.go
+++ b/kem/kyber/templates/pkg.templ.go
@@ -256,19 +256,41 @@ func (sk *PrivateKey) Pack(buf []byte) {
 // Unpacks sk from buf.
 //
 // Panics if buf is not of size PrivateKeySize.
+{{ if .NIST -}}
+//
+// Returns an error if buf is not of size PrivateKeySize, or private key
+// doesn't pass the ML-KEM decapsulation key check.
+func (sk *PrivateKey) Unpack(buf []byte) error {
+	if len(buf) != PrivateKeySize {
+		return kem.ErrPrivKeySize
+	}
+{{- else -}}
 func (sk *PrivateKey) Unpack(buf []byte) {
 	if len(buf) != PrivateKeySize {
 		panic("buf must be of length PrivateKeySize")
 	}
+{{- end }}
 
 	sk.sk = new(cpapke.PrivateKey)
 	sk.sk.Unpack(buf[:cpapke.PrivateKeySize])
 	buf = buf[cpapke.PrivateKeySize:]
 	sk.pk = new(cpapke.PublicKey)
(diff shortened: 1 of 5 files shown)


kem/kyber/templates/pkg.templ.go
```diff
@@ -256,19 +256,41 @@ func (sk *PrivateKey) Pack(buf []byte) {
...
+	var hpk [32]byte
+	h := sha3.New256()
+	h.Write(buf[:cpapke.PublicKeySize])
+	h.Read(hpk[:])
+{{ end -}}
 	buf = buf[cpapke.PublicKeySize:]
 	copy(sk.hpk[:], buf[:32])
 	copy(sk.z[:], buf[32:])
+{{ if .NIST -}}
+	if !bytes.Equal(hpk[:], sk.hpk[:]) {
```
Lekensteyn, line 289: I manually compared these steps with the one from section 7.3 and 8 of https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.203.pdf, it matches 👍

kem/kyber/templates/pkg.templ.go
```diff
@@ -256,19 +256,41 @@ func (sk *PrivateKey) Pack(buf []byte) {
 // Unpacks sk from buf.
 //
 // Panics if buf is not of size PrivateKeySize.
+{{ if .NIST -}}
+//
+// Returns an error if buf is not of size PrivateKeySize, or private key
+// doesn't pass the ML-KEM decapsulation key check.
+func (sk *PrivateKey) Unpack(buf []byte) error {
+	if len(buf) != PrivateKeySize {
```
Lekensteyn, line 264: With this addition, the extra check in `UnmarshalBinaryPrivateKey` for the same condition becomes redundant. Can we simplify the implementation template by unconditionally including this check and returning `error` for this function?

Are there users of `Unpack` other than `UnmarshalBinaryPrivateKey`? If so, should we try harder to prevent intermediate values in `sk.sk`, `sk.pk`, `sk.hpk` and `sk.z` when the error check fails? (E.g. by writing these only at the end, or clearing values on error?)

Lekensteyn (approved):
Checks looks reasonable, but I have an open question.

kem/kyber/templates/pkg.templ.go
bwesterb: I don't want to change the existing Kyber API.

kem/kyber/templates/pkg.templ.go
bwesterb: Thanks!

armfazh (approved):

PR #510, opened 2024-09-05
kem: add X25519MLKEM768 TLS hybrid KEM

bwesterb:


kem/hybrid/hybrid.go
```diff
@@ -57,6 +58,9 @@ func Kyber1024X448() kem.Scheme { return kyber1024X }
 // Returns the hybrid KEM of Kyber768Draft00 and P-256.
 func P256Kyber768Draft00() kem.Scheme { return p256Kyber768Draft00 }
 
+// Returns the hybrid KEM of ML-KEM-768 and X25519.
```
Lekensteyn, line 61: Looks reasonable, would it be worth linking to <https://www.ietf.org/archive/id/draft-kwiatkowski-tls-ecdhe-mlkem-01.html#section-3.1.3>?

Lekensteyn (approved):

PR #511, opened 2024-09-07
feature(): add concurrent generate safePrime

zyjblockchain:
## new feature
When generating an RSA private key, most of the time is spent searching for safe primes. This feature takes advantage of Golang’s concurrency to allow the safe prime search to be executed concurrently, which theoretically can save time.


armfazh:
This is a great idea, I would like to make some tests locally before merging.

PR #519, opened 2024-10-16
Add OIDs to ML-DSA

bwesterb:


sign/dilithium/gen.go
```diff
@@ -59,6 +61,20 @@ func (m Mode) NIST() bool {
 	return strings.HasPrefix(m.Name, "ML-DSA-")
 }
 
+func (m Mode) OidGo() string {
```
armfazh, line 64: this can be replaced by calling:
```go
fmt.Sprintf("%#v", m.Oid)
```

armfazh, line 64: Just add a documentation pointer.

```suggestion
// https://csrc.nist.gov/Projects/computer-security-objects-register/algorithm-registration
func (m Mode) OidGo() string {
```

armfazh (approved):

PR #526, opened 2025-01-17
Run semgrep cronjob only in upstream repository.

armfazh:
Disables semgrep job in forked repos.

### DIFF
--- a/.github/workflows/semgrep.yml
+++ b/.github/workflows/semgrep.yml
@@ -11,6 +11,7 @@ jobs:
   semgrep:
     name: semgrep/ci
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


bwesterb (approved):

.github/workflows/semgrep.yml
```diff
@@ -11,6 +11,7 @@ jobs:
   semgrep:
     name: semgrep/ci
     runs-on: ubuntu-latest
+    if: github.repository == 'cloudflare/circl'
```
thibmeu, line 14: this seems like it should be the default @hrushikeshdeshpande 

possibly along these lines

```
if: github.repository_owner == 'cloudflare'
```

PR #529, opened 2025-01-18
ckem: move crypto/elliptic to crypto/ecdh

MingLLuo:
> TODO move over to crypto/ecdh once we can assume Go 1.20.

I found that there is an unfinished TODO in `ckem.go`, and the go version we use now is higher than 1.20.

Due to the limited code using crypto/ecdh in the entire repository, I can only go by my own experience with the transition, so some of the implementations may not be so satisfying.

This also help fix some problem of 
> TODO: replace deprecated elliptic.Marshal, elliptic.GenerateKey,
> elliptic.Unmarshal, params.ScalarBaseMult before re-enabling SA1019.
in circl/.golangci.yaml

### DIFF
--- a/kem/hybrid/ckem.go
+++ b/kem/hybrid/ckem.go
@@ -3,6 +3,7 @@ package hybrid
 import (
 	"crypto/ecdh"
 	cryptoRand "crypto/rand"
+
 	"github.com/cloudflare/circl/kem"
 	"github.com/cloudflare/circl/xof"
 )
--- a/kem/hybrid/ckem.go
+++ b/kem/hybrid/ckem.go
@@ -9,11 +9,11 @@ import (
 )
 
 type cPublicKey struct {
-	scheme *cScheme
+	scheme cScheme
 	key    *ecdh.PublicKey
 }
 type cPrivateKey struct {
-	scheme *cScheme
+	scheme cScheme
 	key    *ecdh.PrivateKey
 }
 type cScheme struct {
@@ -22,7 +22,7 @@ type cScheme struct {
 
 var p256Kem = &cScheme{ecdh.P256()}
 
-func (sch *cScheme) Name() string {
+func (sch cScheme) Name() string {
 	switch sch.curve {
 	case ecdh.P256():
 		return "P-256"
@@ -35,7 +35,7 @@ func (sch *cScheme) Name() string {
 	}
 }
 
-func (sch *cScheme) PublicKeySize() int {
+func (sch cScheme) PublicKeySize() int {
 	switch sch.curve {
 	case ecdh.P256():
 		return 65
@@ -48,7 +48,7 @@ func (sch *cScheme) PublicKeySize() int {
 	}
 }
 
-func (sch *cScheme) PrivateKeySize() int {
+func (sch cScheme) PrivateKeySize() int {
 	switch sch.curve {
 	case ecdh.P256():
 		return 32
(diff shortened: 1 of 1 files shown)


kem/hybrid/ckem.go
```diff
@@ -1,57 +1,72 @@
...
-func (sch *cScheme) ptSize() int {
-	return (sch.curve.Params().BitSize + 7) / 8
-}
+var p256Kem = &cScheme{ecdh.P256()}
 
 func (sch *cScheme) Name() string {
-	return sch.curve.Params().Name
+	switch sch.curve {
+	case ecdh.P256():
+		return "P256"
```
bwesterb, line 28: Missing `-`

bwesterb: (same for the others)

kem/hybrid/ckem.go
```diff
@@ -79,16 +92,15 @@ func (sk *cPrivateKey) Equal(other kem.PrivateKey) bool {
...
+	return oth.key.Equal(sk.key)
 }
 
 func (sk *cPrivateKey) Public() kem.PublicKey {
-	x, y := sk.scheme.curve.ScalarBaseMult(sk.key)
-	return &cPublicKey{
-		sk.scheme,
-		x,
-		y,
+	pk, err := sk.scheme.curve.NewPublicKey(sk.key.Bytes())
```
bwesterb, line 99: @armfazh You see the issue with this line ;).

MingLLuo: 😢 There's a redundancy, let me fix them

bwesterb: It's not just a redundancy: you're trying to parse the private key as if it were the public key. Luckily the Go library checks the length, because otherwise you'd be leaking the private key as the public key.

MingLLuo: Ops, You are right. I intuitively thought it was reading a private key and then constructing a public key, but after carefully reading the source code and doc, I found it wrong. 

Maybe I can write some tests cover this area later. Thanks for your correction, otherwise I would still be confusing.

kem/hybrid/ckem.go
```diff
@@ -147,9 +157,14 @@ func (pk *cPublicKey) X(sk *cPrivateKey) []byte {
...
 	}
 
-	sharedKey := make([]byte, pk.scheme.SharedKeySize())
-	xShared, _ := pk.scheme.curve.ScalarMult(pk.x, pk.y, sk.key)
-	xShared.FillBytes(sharedKey)
+	privKey, err := pk.scheme.curve.NewPrivateKey(sk.key.Bytes())
+	if err != nil {
+		panic(err)
+	}
+	sharedKey, err := privKey.ECDH(pk.key)
```
armfazh, line 164:
```suggestion
    sharedKey, err := sk.key.ECDH(pk.key)
```

kem/hybrid/ckem.go
```diff
@@ -193,15 +208,21 @@ func (sch *cScheme) UnmarshalBinaryPublicKey(buf []byte) (kem.PublicKey, error)
...
 		return nil, kem.ErrPrivKeySize
 	}
-	ret := cPrivateKey{sch, make([]byte, sch.PrivateKeySize())}
-	copy(ret.key, buf)
+	key, err := sch.curve.NewPrivateKey(buf)
+	if err != nil {
+		return nil, err
+	}
+	ret := cPrivateKey{sch, key}
 	return &ret, nil
```
armfazh, line 227:
```suggestion
	return &cPrivateKey{sch, key}, nil
```

kem/hybrid/ckem.go
```diff
@@ -1,57 +1,72 @@
...
-
-func (sch *cScheme) scSize() int {
-	return (sch.curve.Params().N.BitLen() + 7) / 8
+	curve ecdh.Curve
 }
 
-func (sch *cScheme) ptSize() int {
-	return (sch.curve.Params().BitSize + 7) / 8
-}
+var p256Kem = &cScheme{ecdh.P256()}
```
armfazh, line 23: Let's make a `cScheme{}` struct to satisfy the `kem.Scheme` interface, instead of the pointer to a `cScheme`.

Thus, `cPrivateKey` and `cPublicKey` does not store a pointer to `cScheme`.

armfazh (changes requested):

armfazh (approved):

armfazh:
Thanks @MingLLuo

PR #530, opened 2025-01-22
hpke: Update HPKE code to use ecdh stdlib package.

armfazh:
Reling on the ecdh package from stdlib.
Similar to #529


hpke/shortkem.go
```diff
@@ -49,122 +68,84 @@ func (s shortKEM) DeriveKeyPair(seed []byte) (kem.PublicKey, kem.PrivateKey) {
...
+	for ctr := 0; ctr <= 255; ctr++ {
+		bytes := s.labeledExpand(
 			dkpPrk,
 			[]byte("candidate"),
 			[]byte{byte(ctr)},
 			uint16(s.byteSize()),
 		)
 		bytes[0] &= bitmask
-		skBig.SetBytes(bytes)
+		sk, err := s.UnmarshalBinaryPrivateKey(bytes)
```
lukevalenta, line 84: I confirmed that [Curve.NewPrivateKey](https://pkg.go.dev/crypto/ecdh#Curve) (called by UnmarshalBinaryPrivateKey) checks that the value is lower than the order of the curve, so is equivalent to the existing check.

lukevalenta (approved):

bwesterb (approved):

PR #541, opened 2025-03-20
kem/hybrid: ensure X25519 hybrids fails with low order points

Lekensteyn:
X25519 public keys with low order points result in an all-zeroes shared secret. Ensure that hybrids with X25519 and X448 fail during Encapsulate and Decapsulate when the peer provides such a garbage public key.
___
I discovered this while rebasing our Go fork on Go 1.24 which upgraded the BoGo test suite as part of implementing X25519MLKEM768: https://github.com/golang/go/commit/4b7f7cd87dfcbc17861c908b20a6101e5915ef59#diff-db7320adf5b82005863ef324a30e91235d0268afe79856722226022f3afc2611

That upgrade pulled in this change that adds extra tests for key shares: https://github.com/google/boringssl/commit/ed95627d2f074b6a8636fc8eaaa595903dbfcc9

Without this fix, these `crypto/tls` tests would fail as the TLS handshake continues instead of aborting:
```
TestBogoSuite/CurveTest-Invalid-LowOrderX25519Point-Server-MLKEM-TLS13
TestBogoSuite/CurveTest-Invalid-LowOrderX25519Point-Client-MLKEM-TLS13
```

I don't think this is a security issue, garbage in, garbage out. Nevertheless, it doesn't hurt to add the extra checks.

### DIFF
--- a/kem/hybrid/xkem.go
+++ b/kem/hybrid/xkem.go
@@ -131,7 +131,7 @@ func (sch *xScheme) Encapsulate(pk kem.PublicKey) (ct, ss []byte, err error) {
 	return sch.EncapsulateDeterministically(pk, seed)
 }
 
-func (pk *xPublicKey) X(sk *xPrivateKey) []byte {
+func (pk *xPublicKey) X(sk *xPrivateKey) ([]byte, error) {
 	if pk.scheme != sk.scheme {
 		panic(kem.ErrTypeMismatch)
 	}
@@ -141,14 +141,18 @@ func (pk *xPublicKey) X(sk *xPrivateKey) []byte {
 		var ss2, pk2, sk2 x25519.Key
 		copy(pk2[:], pk.key)
 		copy(sk2[:], sk.key)
-		x25519.Shared(&ss2, &sk2, &pk2)
-		return ss2[:]
+		if !x25519.Shared(&ss2, &sk2, &pk2) {
+			return nil, kem.ErrPubKey
+		}
+		return ss2[:], nil
 	case x448.Size:
 		var ss2, pk2, sk2 x448.Key
 		copy(pk2[:], pk.key)
 		copy(sk2[:], sk.key)
-		x448.Shared(&ss2, &sk2, &pk2)
-		return ss2[:]
+		if !x448.Shared(&ss2, &sk2, &pk2) {
+			return nil, kem.ErrPubKey
+		}
+		return ss2[:], nil
 	}
 	panic(kem.ErrTypeMismatch)
 }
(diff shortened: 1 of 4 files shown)


kem/hybrid/xkem.go
```diff
@@ -165,7 +169,10 @@ func (sch *xScheme) EncapsulateDeterministically(
 	}
 
 	pk2, sk2 := sch.DeriveKeyPair(seed)
-	ss = pub.X(sk2.(*xPrivateKey))
+	ss, err = pub.X(sk2.(*xPrivateKey))
+	if err != nil {
+		return
```
armfazh, line 174: otherwise returns `ss` filled.
```suggestion
		return nil, nil, err
```

kem/xwing/xwing.go
```diff
@@ -252,7 +252,9 @@ func (pk *PublicKey) EncapsulateTo(ct, ss, seed []byte) {
 	copy(ekx[:], seed[32:])
 
 	x25519.KeyGen(&ctx, &ekx)
-	x25519.Shared(&ssx, &ekx, &pk.x)
+	if !x25519.Shared(&ssx, &ekx, &pk.x) {
+		// TODO Propagate invalid public key error, https://github.com/dconnolly/draft-connolly-cfrg-xwing-kem/issues/28
```
armfazh, line 256: `panic(err)` in the meantime.

kem/xwing/xwing.go
```diff
@@ -283,7 +285,9 @@ func (sk *PrivateKey) DecapsulateTo(ss, ct []byte) {
 	copy(ctx[:], ct[mlkem768.CiphertextSize:])
 
 	sk.m.DecapsulateTo(ssm[:], ctm)
-	x25519.Shared(&ssx, &sk.x, &ctx)
+	if !x25519.Shared(&ssx, &sk.x, &ctx) {
```
bwesterb, line 288: For X-Wing we can ignore for now.

bwesterb (approved):

kem/xwing/xwing.go
Lekensteyn: Since it could be triggered remotely, I didn't want to add a `panic`. Garbage in, garbage out seemed acceptable. @bwesterb WDYT?

kem/xwing/xwing.go
Lekensteyn: It would be nice to have this ambiguity clarified in the specification :)
https://github.com/dconnolly/draft-connolly-cfrg-xwing-kem/issues/28

See also Armando's [comment above](https://github.com/cloudflare/circl/pull/541#discussion_r2006633757) for an alternative suggestion.

Lekensteyn:
This PR currently rejects bad cases after the ECDH calculation.

This could also be fixed by rejecting public keys with low-order points in UnmarshalBinaryPublicKey, consistent with how bad cases for NIST curves are prevented. This could require a new `NewPublicKey() (*Key, error)` method in the `dh/x25519` and `dh/x448` CIRCL packages.

Any particular preference?

kem/xwing/xwing.go
Lekensteyn: The linter didn't like the empty branch, so I just moved the comment above.

bwesterb:
> Any particular preference?

You can't hit the case unless you try, so hitting the error late is not a problem.

lukevalenta (approved):
Looks good. I agree this isn't a security concern, but doesn't hurt to reject.

PR #540, opened 2025-03-20
Change extension of assembler headers .h -> .s

armfazh:
Related to #537

PR #542, opened 2025-03-21
.github: Use native ARM64 builders instead of QEMU

Lekensteyn:
The QEMU builder sporadically crashes in gcc. Since there are now native ARM64 build agents available, use that instead. Re-enable Go 1.23 for ARM64, the original Go 1.23 incompatibility with QEMU got resolved.
___
This should prevent flaky builds such as:
- https://github.com/cloudflare/circl/actions/runs/13960263989/job/39080251377 (#540)
- https://github.com/cloudflare/circl/actions/runs/13989058956/job/39168905189 (#541)

### DIFF
--- a/.github/workflows/ci-actions.yml
+++ b/.github/workflows/ci-actions.yml
@@ -41,27 +41,23 @@ jobs:
         run: go build -v ./...
       - name: Testing
         run: go test -v -count=1 ./...
-  exotic_job:
-    name: Go-${{matrix.CFG[2]}}/${{matrix.CFG[0]}}
+  arm64_job:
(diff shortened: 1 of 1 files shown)


Lekensteyn:
After 45 minutes, the ARM and Windows builds are still in queued state. Builds are not blocking, so that might be fine.

"While the arm64 runners are in public preview, you may experience longer queue times during peak usage hours."
https://github.blog/changelog/2025-01-16-linux-arm64-hosted-runners-now-available-for-free-in-public-repositories-public-preview/
https://github.com/orgs/community/discussions/148648

bwesterb (approved):

armfazh (approved):

PR #556, opened 2025-07-17
group: Removes use of elliptic Marshal and Unmarshal functions.

### LINKED ISSUE #549: group: elliptic.Marshal is deprecated
https://github.com/cloudflare/circl/blob/aa837fd1539dc5eb9e33dcbb09c79b7fccb99a6b/group/short.go#L231

armfazh:
Removes use of elliptic Marshal and Unmarshal functions.

Fixes #549


group/short.go
```diff
@@ -226,9 +228,20 @@ func (e *wElt) MarshalBinary() ([]byte, error) {
 	if e.IsIdentity() {
 		return []byte{0x0}, nil
 	}
+
+	ec := e.wG.c
+	if e.wG == P384 {
+		ec = elliptic.P384()
+	}
```
bwesterb, line 235: Why this line?

group/short.go
```diff
@@ -254,11 +268,22 @@ func (e *wElt) UnmarshalBinary(b []byte) error {
 		}
 		e.x, e.y = x, y
 	case l == 1+2*byteLen && b[0] == 0x04: // uncompressed
-		x, y := elliptic.Unmarshal(e.wG.c, b)
-		if x == nil {
+		var err error
+		switch bitSize {
```
bwesterb, line 272: Maybe better match on `e.wG.c` if we ever want to add Curve25519 or some other curve with clashing bitlen.

bwesterb (changes requested):

group/short.go
armfazh: this is because the [`ECDH` function](https://cs.opensource.google/go/go/+/refs/tags/go1.24.5:src/crypto/ecdsa/ecdsa.go;l=52) checks whether ec is an `elliptic.Curve` object.

I made changes to avoid this check.

bwesterb (approved):

PR #559, opened 2025-08-05
Support encoding/decoding ML-DSA private keys (as long as they contain seeds)

bwesterb:
Also:

- Fixes OIDs for ML-DSA.
- Add `Seed()` interface to extract seed from private key.


pki/pki.go
```diff
@@ -161,17 +219,38 @@ func MarshalPEMPrivateKey(sk sign.PrivateKey) ([]byte, error) {
 }
 
 func MarshalPKIXPrivateKey(sk sign.PrivateKey) ([]byte, error) {
-	data, err := sk.MarshalBinary()
-	if err != nil {
-		return nil, err
-	}
+	var (
+		data []byte
+		err  error
```
lukevalenta, line 224: nit: error is already defined above

lukevalenta (approved):

bwesterb:
Whoopsie.

pki/pki.go
```diff
@@ -105,6 +113,56 @@ func UnmarshalPKIXPrivateKey(data []byte) (sign.PrivateKey, error) {
...
+	// ML-DSA unfortunately has a complex private key format, which we
+	// handle here separately. If future schemes require custom parsing
+	// as well, we can introduce an interface for that.
+	if isMLDSA(scheme) {
+		// Handle case of seed-only private key
+		ss := cryptobyte.String(pkix.PrivateKey)
+		tag := casn1.Tag(0).ContextSpecific()
+		if ss.PeekASN1Tag(tag) {
+			var ss2 cryptobyte.String
+			ss.ReadASN1(&ss2, tag)
```
armfazh, line 126: add a check for ReadASN1 was ok.

sign/dilithium/mode3/internal/dilithium.go
```diff
@@ -232,6 +240,13 @@ func NewKeyFromSeed(seed *[common.SeedSize]byte) (*PublicKey, *PrivateKey) {
 	return &pk, &sk
 }
 
+func (sk *PrivateKey) Seed() []byte {
+	if !sk.seedSet {
+		return nil
+	}
+	return sk.seed[:]
```
armfazh, line 247: it could return a copy, otherwise the slice can be modified.

armfazh (approved):
two nits

PR #562, opened 2025-08-14
Preparation for ARM64 Implementation of poly operations for dilithium package.

elementrics:
To test performance difference on arm64 chips: "go test -benchmem -run=^$ ./sign/internal/dilithium -bench=Add"

```
BenchmarkAddGeneric-10          12393860                95.46 ns/op            0 B/op          0 allocs/op
BenchmarkAdd-10                 68264402                17.40 ns/op            0 B/op          0 allocs/op
```

### DIFF
--- a/sign/internal/dilithium/arm64.s
+++ b/sign/internal/dilithium/arm64.s
@@ -8,7 +8,7 @@ TEXT ·polyAdd(SB), NOSPLIT|NOFRAME, $0-24
     MOVD    a+8(FP), R1
     MOVD    b+16(FP), R2
 
-    // loop counter (each iterations processes 16 elements so 16 * 16 = 256 = N)
+    // loop counter (each iteration processes 16 elements so 16 * 16 = 256 = N)
     // manually unrolling could also be done, for now skipped
     MOVW    $16, R3
 


elementrics:
once this PR is approved, I will provide the the other PR due to the fact that the other PR's need the base files (arm64.s and arm64.go)

sign/internal/dilithium/arm64.s
```diff
@@ -0,0 +1,29 @@
...
+
+    // loop counter (each iteration processes 16 elements so 16 * 16 = 256 = N)
+    // manually unrolling could also be done, for now skipped
+    MOVW    $16, R3
+
+add:
+    VLD1.P  (64)(R1), [V0.S4, V1.S4, V2.S4, V3.S4]
+    VLD1.P  (64)(R2), [V4.S4, V5.S4, V6.S4, V7.S4]
+
+    VADD    V4.S4, V0.S4, V8.S4
```
bwesterb, line 19: Note: it's not necessary here, but you can reuse `V0` or `V8` as target register.

bwesterb (approved):

armfazh (approved):

sign/internal/dilithium/arm64.s
elementrics: you are absolutely right!

PR #565, opened 2025-08-16
add arm64 version of polyMulBy2toD

elementrics:
To test performance difference on arm64 chips: "go test -benchmem -run=^$ ./sign/internal/dilithium -bench=MulBy2"

```
BenchmarkMulBy2toDGeneric-10            13260348                88.91 ns/op            0 B/op          0 allocs/op
BenchmarkMulBy2toD-10                   78609487                15.04 ns/op            0 B/op          0 allocs/op
```


bwesterb (approved):

bwesterb:
Thanks again! Just waiting for CI to complete before merging.

PR #566, opened 2025-08-17
add arm64 version of polySub

elementrics:
To test performance difference on arm64 chips: "go test -benchmem -run=^$ ./sign/internal/dilithium -bench=Sub"

```
BenchmarkSubGeneric-10          12127086                89.72 ns/op            0 B/op          0 allocs/op
BenchmarkSub-10                 56654660                20.93 ns/op            0 B/op          0 allocs/op
```

### DIFF
--- a/sign/internal/dilithium/arm64.s
+++ b/sign/internal/dilithium/arm64.s
@@ -91,3 +91,35 @@ loop:
     BGT     loop
 
     RET
+
+// func polySubARM64(p, a, b *Poly)
+TEXT ·polySubARM64(SB), NOSPLIT|NOFRAME, $0-24
+    MOVD    p+0(FP), R0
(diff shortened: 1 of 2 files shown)


sign/internal/dilithium/arm64.s
```diff
@@ -91,3 +91,35 @@ loop:
...
+    MOVD    a+8(FP), R1
+    MOVD    b+16(FP), R2
+
+    MOVW    $(const_N / 16), R3
+    MOVW    $(const_Q << 1), R4
+
+    VDUP    R4, V8.S4
+
+loop:
+    VLD1.P  (64)(R1), [V0.S4, V1.S4, V2.S4, V3.S4]
```
bwesterb, line 107: Nit: add a note that we compute as a + (2q - b)

PR #573, opened 2025-10-23
Add ExportRaw() and ImportRaw() serialization helpers for Prio3 share…

test-sum:
### Summary
This PR adds minimal, backward-compatible serialization helpers for CIRCL’s Prio3 types to support **cross-process and distributed deployments** of Verifiable Distributed Aggregation Functions (VDAFs).

### Background
The current CIRCL VDAF implementation assumes that all aggregators and the collector operate within a single process.  
Serialization helpers (e.g., for `InputShare`, `PrepShare`, `OutShare`) are defined under `vdaf/prio3/internal/prio3`, which cannot be accessed externally due to Go’s `internal/` package visibility rules.  

This makes distributed deployments — where aggregators and the collector run on distinct hosts — infeasible without modifying CIRCL.  
Our patch exposes a minimal public serialization interface to enable share transmission between processes while preserving CIRCL’s internal cryptographic semantics.

### Changes Introduced
- Added public methods for raw export/import of Prio3 share types:
  ```go
  func (s *OutShare[V,E]) ExportRaw() ([]byte, error)
  func (s *OutShare[V,E]) ImportRaw([]byte) error

### DIFF
--- a/vdaf/prio3/arith/fp64/vector.go
+++ b/vdaf/prio3/arith/fp64/vector.go
@@ -4,6 +4,7 @@ package fp64
 
 import (
 	"encoding/binary"
+	"errors"
 	"io"
 	"math/bits"
 	"slices"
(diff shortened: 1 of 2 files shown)


armfazh:
it looks like this PR needs to be rebase on top of the current main branch

armfazh (changes requested):
Could you please verify whether `MarshalBinary` already produces the same encoding?

PR #578, opened 2025-12-19
Revert to using package-declared HPKE errors for shortkem instead of standard library errors

harshiniwho:
A commit to update HPKE code to use ecdh stdlib package: https://github.com/cloudflare/circl/commit/342ad81204fc580bd54aec19f09cb55b214b7d39 - changed the error returned from the Unmarshal functions from `shortkem` to return the underlying error (which returns the standard library errors). Considering `xkem` still returns the package defined errors and if the project doesn't need the underlying error for some reasons, it would make sense to keep the error returned consistent and also helps projects handle errors better.


armfazh (approved):

bwesterb:
Thank you!

PR #593, opened 2026-03-30
Perform final reduction of Horner poly evaluation

sug0:
Amendment to #590

Gets rid of modular arithmetic in additions, which stays within the bounds of the ring modulus, saving a few CPU cycles. The operations are const time on a best-effort basis... Switching from `math/big` to a library with Montgomery reductions is actually the better approach here.


armfazh:
I measured and the time difference is a small percentage on my machine. Also note that adding without performing modular reduction (i.e., lazy reduction) may reveal some information about the coefficients in the case the number of words of `sum` grows, making that the subsequent multiplication takes more time. Of course, the entire arithmetic still uses `big.Int` and some replacement is needed too. Let's do modular reduction to cover that edge case.  

@sug0 thanks for the proposed change.

PR #594, opened 2026-04-01
Add UnblindSerialized method to Client

thomas-samoht:


armfazh (changes requested):
The Client has to use a `FinalizeData` struct and an `Evaluation` to invoke `Finalize()`. 
`Evaluation` has its fields public, `FinalizeData` does not. 
It is possible to add some serialization logic to `FinalizeData` and then invoke `Finalize`.

PR #595, opened 2026-04-01
sign/bls: rejects aggregated signatures built with duplicated messages.

armfazh:
[AggregateVerify](https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-bls-signature-05#section-3.1.1) must reject attempts to verify aggregated signatures with duplicated messages.

### DIFF
--- a/sign/bls/bls.go
+++ b/sign/bls/bls.go
@@ -365,6 +365,18 @@ func VerifyAggregate[K KeyGroup](pubs []*PublicKey[K], msgs [][]byte, aggSig Sig
 		return false
 	}
 
+	// 1. If any two input messages are equal, return INVALID.
+	set := make(map[string]struct{}, len(msgs))
+	for _, m := range msgs {
+		k := string(m)
+		if _, found := set[k]; found {
+			return false
+		}
+		set[k] = struct{}{}
+	}
+
+	// 2. CoreAggregateVerify algorithm checks an aggregated signature over
+	// several (PK, message) pairs.
 	for _, p := range pubs {
 		if !p.Validate() {
 			return false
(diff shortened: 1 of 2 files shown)


sign/bls/bls.go
```diff
@@ -365,6 +365,18 @@ func VerifyAggregate[K KeyGroup](pubs []*PublicKey[K], msgs [][]byte, aggSig Sig
 		return false
 	}
 
+	// 1. If any two input messages are equal, return INVALID.
+	set := make(map[string]struct{}, len(msgs))
+	for _, m := range msgs {
+		k := string(m)
```
rozbb: Why cast to string?

armfazh: a `[]byte` type does not get native comparison (it has to use `bytes.Equal`), while `string` does.

rozbb: Is it guaranteed that `bytes.Equal(x, y)` if and only if `string(x) == string(y)` for all bytestrings `x`, `y`?

armfazh: interesting, the code now encodes the messages into string explictly.
another option is to hash messages, but more costly and collisions might happen.

rozbb: Ah I see. My concern now is that this might allocate a huge amount if messages are large.

I think the original string thing might be fine, actually. Go's [string comparison](https://cs.opensource.google/go/go/+/refs/tags/go1.26.1:src/strings/compare.go;l=15) is actually a byte comparison. And according to Gemini, Go's compiler avoids allocating for `string(x)` when it's only used for a map lookup.

The other option is you can do `make(map[[32]byte]struct{}, len(msgs))`, and use Blake3 or something fast for hashing. Blake3 is cryptographic, so collisions won't be an issue.

rozbb: I am learning a lot about go 😃

bwesterb (approved):

PR #596, opened 2026-04-16
zk/qndleq: Fixes challenge calculation

armfazh:
Two checks:
1. Challenge computation must ensure the serialization of the canonical representatives of numbers mod N.
2. Challenge generated cannot be congruent to zero mod m.

### DIFF
--- a/zk/qndleq/qndleq.go
+++ b/zk/qndleq/qndleq.go
@@ -72,30 +72,57 @@ func SampleQn(random io.Reader, N *big.Int) (*big.Int, error) {
 func Prove(random io.Reader, x, g, gx, h, hx, N *big.Int, secParam uint) (*Proof, error) {
 	rSizeBits := uint(N.BitLen()) + 2*secParam
 	rSizeBytes := (rSizeBits + 7) / 8
-
 	rBytes := make([]byte, rSizeBytes)
-	_, err := io.ReadFull(random, rBytes)
-	if err != nil {
-		return nil, err
+
+	ONE := big.NewInt(1)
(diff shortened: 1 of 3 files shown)


zk/qndleq/qndleq.go
```diff
@@ -72,30 +72,52 @@ func SampleQn(random io.Reader, N *big.Int) (*big.Int, error) {
...
+			return nil, err
+		}
 
-	return &Proof{z, c, secParam}, nil
+		// Challenge must not be congruent to zero.
+		//   c != 0 mod m, where m = (p-1)(q-1)/4, and N = p*q.
+		// Check this by doing an Exp because m is unknown.
+		gc.Exp(g, c, N)
+		hc.Exp(h, c, N)
+		if gc.Cmp(ONE) != 0 && hc.Cmp(ONE) != 0 {
```
rozbb, line 100: Why is this check the same as checking c != 0 mod m?

armfazh: This is valid assuming `N` is a power of two safe prime numbers.
`g^c=1` means `c` is congruent to zero in the exponents domain. 

In the verification equation, `c` multiplies the witness. So when c is zero, it removes the witness allowing to trivially pass the verification check.

rozbb: Ah neat ok. Is this documented in a spec somewhere? Could you write the reasoning in a comment otherwise?

armfazh: just that info added as a comment.

bwesterb (approved):

PR #602, opened 2026-04-28
Make Unblind function public

henkerik:
Hi 👋 , 

Thanks for a the great work creating this library. I would like to request a small feature change to the OPRF client. 

Our usecase of the OPRF library is perhaps a bit different, so let me explain. There are three party parties involved:

- (A) a sender, which knows some secret ID number of a person. 
- (B) a receiver, which needs to process a message from the sender and needs a stable identifier of the person but is not allowed to see te secret ID number.
- (C) a service capable of generating pseudonyms. 

The OPRF client in this case is _both_ the sender and the receiver. However, from the `finalizeData` structure only the `blinds` are communicated between both. The original inputs are _not_ shared with the receiver because the are secret. 

The OPRF server in this case is the pseudonym generating service. We also don't want to share the secret ID with this service. That's why we are using OPRF. 

Of course, without the original inputs the receiver won't be able to reconstruct the final output using the `client.Finalize` function. However, we are capable of executing the `unblind` function on the receiver. This results in a stable identifier which the receiver can use. Luckily, for our use case that is enough.

Currently, the `unblind` function is private, this PR makes it public. 

For reference, I wrote some code tying these three services together and using the `unblind` function directly: 

https://github.com/henkerik/oprf-cloudflare/blob/master/main.go#L206

Feel free to ask me for any updates.

### DIFF
--- a/oprf/client.go
+++ b/oprf/client.go
@@ -64,6 +64,17 @@ func (c client) blind(inputs [][]byte, blinds []Blind) (*FinalizeData, *Evaluati
 	return finData, evalReq, nil
 }
 
+func (c Client) Unblind(serUnblindeds [][]byte, blindeds []group.Element, blind []Blind) (err error) {
+	if len(serUnblindeds) == 0 {
+		return ErrInvalidInput
+	}
+	if len(serUnblindeds) != len(blindeds) || len(blindeds) != len(blind) {
+		return ErrInvalidInput
+	}
+
+	return c.unblind(serUnblindeds, blindeds, blind)
+}
+
 func (c client) unblind(serUnblindeds [][]byte, blindeds []group.Element, blind []Blind) (err error) {
 	invBlind := c.params.group.NewScalar()
 	U := c.params.group.NewElement()


henkerik:
This PR looks related: https://github.com/cloudflare/circl/pull/594

armfazh:
One question first: 
If the sender is also the receiver, then the `FinalizeData` struct is known (i.e., it knows the private inputs), and call to `client.Finalize`, whose output can be used as the stable identifier.

The final hash needs to be computed (Finalize=unblind+hash), because it ensures the OPRF be indistinguishable from random bytes. Otherwise, the stable identifier is easy to distinguish since it is always the encoding of a curve point.

henkerik:
Hey @armfazh, 

Sorry, perhaps my first message was unclear. The sender and the receiver are different machines (from different organisation even). Just from _a OPRF point of view_ they preform the function of a client. The sender takes care  of the blinding, and the receiver takes care of the unblinding (which is possible because the blinds are transferred from sender to receiver). 

> The final hash needs to be computed (Finalize=unblind+hash), because it ensures the OPRF be indistinguishable from > random bytes. Otherwise, the stable identifier is easy to distinguish since it is always the encoding of a curve point.

I'm not sure I understand this remark 😅  As far as I understand the algorithm, the blinds are randomly chosen for each invocation. As a result the result of the `client.blind` differs from one request to the other and the OPRF server (which creates a pseudonym in our case) differs from one request to the other, even if the same original input is used. Only after unblinding on the receiver we obtain a stable identifier. 

Edit 1:

I think I'm starting to understand your remark. If we only unblind but don't hash, the result is always a point on the curve. Not every random value will occur. After hashing however, this information disappears. So without hashing, the outputs are distinguishable from random. 

Edit 2: 

So, currently we just unblind so the receiver ends up with `Z = Fk​(x)`. Now we need to apply the hash function H. There are two option:

- A) `output = H(Z)` 
- B) `output = H(x ++ Z)`, where `++` is some form of concatenation.

CIRCL does B, but since our receiver doesn't have access to the original inputs `x` , the best we can do is A. Even though B is preferred, it is still beter than using Z directly. Is that correct?

armfazh:
> ... Not every random value will occur. After hashing however, this information disappears. So without hashing, the outputs are distinguishable from random.

Correct.

This may work for your application. Perform the OPRF protocol with `x'= H(x)`. So `x'` can be shared with the Receiver without revealing the private input `x`. The receiver can do `client.Finalize` with the OPRF inputs (i.e., with `x'`).

CIRCL does `output=H(x, Z)` because it is following RFC9497 which provides secure OPRF functionality.

henkerik:
I created a sequence diagram, does this make sense? 

```mermaid
sequenceDiagram;

participant A as Sender
participant B as Pseudonym Service
participant C as Receiver

autonumber

A->>A: P = H1(x), hash to curve
A->>A: P' = r ⋅ P, where r is random
A->>B: P'
B->>B: Z' = k ⋅ P', where k is secret
B-->>A: Z'
A->>C: Register [Z', r, Organisation ID]
C->>C: Z = 1 / r ⋅ Z', unblinding using r
C->>C: Pseudonym = H(Organisation ID || Z), where || is concatenation and H is the hashing function
C->>C: Store in Database: (Pseudonym, Organisation ID)
```

Each Sender has it own `Organisation ID`, so `Organisation ID || Z` results in a different a different pseudonym for each organisation.

armfazh:
I dont have a strong opinion, it depends on what you want to achieve in your protocol.

This seems a possible solution: define `x=Hash(SecretID)` and then run the OPRF on `x`, then the Sender sends `finalizeData=(x,r)` and the `Evaluation=(Z)`, so the Receiver can `Finalize` the protocol.

In any case, the unblind method consists only of scalar multiplications with the inverse of the blind. 
These operations can be done with the `Group` interface (still using CIRCL) without exposing `unblind` as a public method.

armfazh:
> Feel free to ask me for any updates.

any update here.

henkerik:
Hi @armfazh ,

Yes, I think hashing the Secret ID before starting the OPRF algorithm is probably the best approach. I contacted the authors of the server application I’m using to discuss this. I will close this PR for now. Thank you for your time and explanation! They have been most helpful.

PR #605, opened 2026-05-06
sign/dilithium: hold tr by value in pk returned from Public()

### LINKED ISSUE #600: sign/mldsa: is the pointer-sharing of sk.tr into the pk returned by (*PrivateKey).Public() intentional?
> **Question-shaped report, not a vulnerability claim.** Reading this alongside `NewKeyFromSeed` (same file, `:240`), the most plausible explanation is an intentional micro-optimization — avoiding a `pkEncode` + `SHAKE256(·, 64)` on every `Public()` call. A one-line rationale in a comment would give future auditors something concrete to cite. If you'd prefer to address it, the fix is a one-line value copy (`pk.tr = sk.tr`) — happy to send a CL.

### Observation

`sign/mldsa/mldsa65/internal/dilithium.go:474-485`:

```go
func (sk *PrivateKey) Public() *PublicKey {
    var t0 VecK
    pk := &PublicKey{
        rho: sk.rho,
        A:   &sk.A,
        tr:  &sk.tr,        // pointer-shared with sk.tr
    }
    sk.computeT0andT1(&t0, &pk.t1)
    pk.t1.PackT1(pk.t1p[:])
    return pk
}
```

`PublicKey.UnmarshalBinary` at `:115-128` recomputes `tr = SHAKE256(pk_bytes, 64)` honestly. So the same pk has two `tr` states: one from `Public()` (aliased to `sk.tr`), one from `Pack → UnmarshalBinary` (recomputed). Signatures produced under a mutated `sk.tr` verify against the former, fail against the latter.

### Spec reading

FIPS 204 Alg. 6 step 9 defines `tr ← H(pkEncode(ρ, t1), 64)` — a computed value. The SK-stored copy is a cache. Aliasing lets any divergence in `sk.tr` (including #581's unvalidated parse path) silently shape the derived pk.

### Reach

`(*PrivateKey).Public()` is `crypto.Signer`'s public interface.

### What would be useful

Any of:
- "Intentional — here's why" (most likely; close WAI, comment welcome).
- "Worth tightening — patch welcome" (one-line `pk.tr = sk.tr` value copy).
- "Reproducer first" (reproducer available).

bshastry:
## Summary

Resolves cloudflare/circl#600.

`(*PrivateKey).Public()` and `NewKeyFromSeed` previously cached `tr`
on the returned `*PublicKey` as a pointer aliasing `sk.tr`. As a
result, any later mutation of `sk.tr` -- e.g. through a subsequent
`(*PrivateKey).Unpack` on attacker-controlled bytes -- silently
propagated into a `pk` that had already been handed back to the
caller.

This PR switches both sites to a value copy. The returned `pk` now
holds its own `[TRSize]byte` and is independent of subsequent changes
to `sk`.

The change matches the direction in @bwesterb's [reply on #600](https://github.com/cloudflare/circl/issues/600) ("not use a pointer, but rather store a copy").

## Implementation

Edited in `sign/dilithium/mode3/internal/dilithium.go` (the master).
Regenerated to `mode2`, `mode5`, and `sign/mldsa/{mldsa44,mldsa65,mldsa87}`
via `go generate ./sign/dilithium/`.

A regression test `TestPublicTrIsIndependentOfPrivate` is added in
the same master and propagates to all six internal packages. It
asserts that mutating `sk.tr` after `NewKeyFromSeed` and after
`sk.Public()` does not change the `tr` byte values stored in the
returned `pk`.

## Test plan

- [x] ` + "`go generate -v ./...`" + ` clean (no drift outside the 12 intended files)
- [x] ` + "`go vet ./sign/dilithium/... ./sign/mldsa/...`" + `
- [x] ` + "`go build ./sign/dilithium/... ./sign/mldsa/...`" + `
- [x] ` + "`go test -count=1 ./sign/dilithium/... ./sign/mldsa/...`" + ` — 12/12 packages pass, including the new regression test in all six internal packages
- [x] ` + "`go vet -vettool=$(which shadow) ./sign/dilithium/... ./sign/mldsa/...`" + ` clean

### DIFF
--- a/sign/dilithium/mode2/internal/dilithium.go
+++ b/sign/dilithium/mode2/internal/dilithium.go
@@ -236,8 +236,13 @@ func NewKeyFromSeed(seed *[common.SeedSize]byte) (*PublicKey, *PrivateKey) {
 	_, _ = h.Write(packedPk[:])
 	_, _ = h.Read(sk.tr[:])
 
-	// Finish cache of public key
-	pk.tr = &sk.tr
+	// Finish cache of public key. Copy the value rather than aliasing
+	// sk.tr so that later mutation of sk (for instance through a
+	// subsequent (*PrivateKey).Unpack on attacker-controlled bytes)
+	// does not silently propagate into a pk returned earlier from
+	// NewKeyFromSeed. See cloudflare/circl#600.
+	pk.tr = new([TRSize]byte)
+	*pk.tr = sk.tr
 
 	return &pk, &sk
 }
@@ -472,13 +477,18 @@ func SignTo(sk *PrivateKey, msg func(io.Writer), rnd [32]byte, signature []byte)
 }
 
 // Computes the public key corresponding to this private key.
+//
+// The returned pk holds its own copy of tr; later mutation of sk.tr
+// does not affect a pk previously returned from this method.
+// See cloudflare/circl#600.
 func (sk *PrivateKey) Public() *PublicKey {
 	var t0 VecK
 	pk := &PublicKey{
 		rho: sk.rho,
 		A:   &sk.A,
-		tr:  &sk.tr,
+		tr:  new([TRSize]byte),
 	}
+	*pk.tr = sk.tr
 	sk.computeT0andT1(&t0, &pk.t1)
 	pk.t1.PackT1(pk.t1p[:])
 	return pk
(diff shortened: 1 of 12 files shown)


bwesterb:
Thanks! Your PR is included in #606

PR #607, opened 2026-05-07
pki: check pem.Decode returned nil block.

z9z:
UnmarshalPEMPublicKey and UnmarshalPEMPrivateKey panic with a nil pointer dereference when given input that is not a valid PEM block (pem.Decode returns nil for the block). Add a nil check before dereferencing block.Type.
Test added with empty/garbage input.


armfazh (approved):

armfazh:
Thanks @z9z

PR #610, opened 2026-05-30
abe/cpabe/tkn20: fix AND-gate secret sharing.

cjpatton:
(*Formula).share split an AND gate's parent share incorrectly: In0 was set to a fresh random matrix but then overwritten with (Out - In1) where In1 was the zero matrix, so In0 received the entire parent share and In1 received zero. A single AND leaf could therefore reconstruct the shared secret on its own.

The CCA (Boneh-Katz) transform wraps every policy in a new outer AND gate whose left child is an internal wildcard leaf, and every CCA attribute key carries that attribute as a wildcard. Combined with the sharing bug, that single leaf received the full KEM secret, so a key that does not satisfy the public policy could recover the message by running the decapsulation pairing equations on the wildcard wire alone. The stock API's Satisfaction() check is only local logic and does not prevent this, breaking the scheme's access-control (soundness) guarantee for any policy containing at least one gate.

Share the parent across the two AND inputs correctly: one input gets a fresh random share and the other gets (parent - random), so the shares sum to the parent but neither input alone reveals it.

This changes only how the secret is shared at encryption time; the ciphertext format and decryption are unchanged, and both old and new sharings are valid additive sharings that a satisfying set reconstructs. Ciphertexts produced by the previous code remain decryptable by satisfying keys.

Add regression tests:

  - tkn.TestShareAndGateNoSingleLeafReconstructs: at the share() level, the two AND child shares sum to the secret but neither alone equals it or is zero.

  - tkn.TestBKWildcardAttackFails: simulates the attack end to end and asserts a non-satisfying key can no longer recover the plaintext via the wildcard wire.

  - tkn20.TestAndShareInteropBuggyCiphertext: decrypts a hard-coded ciphertext produced by the pre-fix code using a satisfying key derived on the fixed code, guarding ciphertext backwards-compatibility.


abe/cpabe/tkn20/internal/tkn/formula.go
```diff
@@ -248,12 +248,20 @@ func (f *Formula) share(rand io.Reader, k *matrixZp) ([]*matrixZp, error) {
...
 		switch gate.Class {
 		case Andgate:
+			// Secret-share the parent's share across the two AND inputs:
+			// one input gets a fresh random share and the other gets
+			// (parent - random). The two shares sum to the parent share, but
+			// neither input alone reveals it. Previously In0 was overwritten
+			// with (Out - In1) where In1 was the zero matrix, so In0 received
+			// the entire parent share and In1 received zero, letting a single
+			// AND leaf (e.g. the Boneh-Katz wildcard wire) reconstruct the
+			// whole KEM secret.
```
thibmeu: this seems to be a comment that talks about old code. it should either be reformulated or removed.

abe/cpabe/tkn20/internal/tkn/tk.go
thibmeu, line 1: the reason to rename this file seems to be in 17ea6424c3b7211f7b1291d4a0a1dc61ed8e5518

it's to avoid shadowing a `match` type that is declared at abe/cpabe/tkn20/internal/tkn/policy.go:296. It's more a cosmetic change than an actual bug.

abe/cpabe/tkn20/internal/tkn/formula.go
```diff
@@ -248,12 +248,20 @@ func (f *Formula) share(rand io.Reader, k *matrixZp) ([]*matrixZp, error) {
...
 			shares[gate.In0], err = randomMatrixZp(rand, k.rows, k.cols)
 			if err != nil {
 				return nil, err
 			}
 			shares[gate.In1] = newMatrixZp(k.rows, k.cols)
-			shares[gate.In0].sub(shares[gate.Out], shares[gate.In1])
+			shares[gate.In1].sub(shares[gate.Out], shares[gate.In0])
```
thibmeu, line 256: i coenfirm that without the fix, the negative/regression test fail

thibmeu (approved):
I confirm that fixes a bug. The comment should be edited to reflect the current version of the code rather than use a description about a "preovious" version

PR #616, opened 2026-06-01
blindsign/blindrsa: reject message unless co-prime with modulus

cjpatton:
Before blinding the message, ensure that the message and modulus don't have any factor in common. Otherwise, the signer would learn the GCD of the message and modulus which, depending on the modulus, could leak bits of the message.


bwesterb:
Couldn't find that stipulation in the spec.

cjpatton:
> Couldn't find that stipulation in the spec.

It's steps 4 and 5 in [`Blind()`](https://www.rfc-editor.org/rfc/rfc9474.html#section-4.2-4)

bwesterb (approved):

PR #617, opened 2026-06-02
sign/{ed25519,ed448}: check if public key is identity.

cjpatton:
When parsing a public key or verifying a signature, check that the public key is not the identity element.

---

<a href="https://app.devin.ai/review/cloudflare/circl/pull/617" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/sign/ed25519/ed25519.go
+++ b/sign/ed25519/ed25519.go
@@ -327,7 +327,7 @@ func verify(public PublicKey, message, signature, ctx []byte, preHash bool) bool
 	}
 
 	var P pointR1
-	if ok := P.FromBytes(public); !ok {
+	if ok := P.FromBytes(public); !ok || P.IsIdentity() {
 		return false
 	}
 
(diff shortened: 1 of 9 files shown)


bwesterb:
What's the reasoning here? We can't check for all ways a public key can be weak.

drmikecrypto (approved):
Nice catch — accepting the identity point as a valid public key would undermine signature verification assumptions.

The `IsIdentity()` check in `verify()` is the right layer (after `FromBytes` succeeds but before scalar math). The test vectors covering `Verify`, `VerifyPh`, `VerifyWithCtx`, and `UnmarshalBinaryPublicKey` look thorough.

One minor note: worth confirming `IsIdentity()` matches RFC 8032's encoding of the identity (y=1, x=0) for all code paths that ingest public keys, not just verify — but the unmarshal test already covers the scheme entry point.

LGTM from a security perspective.

PR #638, opened 2026-07-14
Add more explicit constant time warnings

bwesterb:
---

<a href="https://app.devin.ai/review/cloudflare/circl/pull/638" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/blindsign/blindrsa/partiallyblindrsa/pbrsa.go
+++ b/blindsign/blindrsa/partiallyblindrsa/pbrsa.go
@@ -1,4 +1,7 @@
 // Package partiallyblindrsa implements a partially blind RSA protocol.
+//
+// Warning: The per-request derivation of the metadata-specific key uses
+// math/big and is not constant time in the private key.
 package partiallyblindrsa
 
 import (
--- a/tss/rsa/rsa_threshold.go
+++ b/tss/rsa/rsa_threshold.go
@@ -3,6 +3,9 @@
 // This package implements the Protocol 1 of "Practical Threshold Signatures"
 // by Victor Shoup [1].
 //
+// Warning: Sign operations rely on math/big and are not constant time in the
+// secret key share. Pass a non-nil random source to Sign to enable blinding.
+//
 // # References
 //
 // [1] https://www.iacr.org/archive/eurocrypt2000/1807/18070209-new.pdf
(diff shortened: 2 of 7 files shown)


blindsign/blindrsa/partiallyblindrsa/pbrsa.go
```diff
@@ -1,4 +1,7 @@
 // Package partiallyblindrsa implements a partially blind RSA protocol.
+//
+// Warning: The per-request derivation of the metadata-specific key uses
+// math/big and is not constant time in the private key.
```
cjpatton, line 4:
```suggestion
// math/big and may therefore leak the private key via a timing side channel.
```

tss/rsa/rsa_threshold.go
```diff
@@ -3,6 +3,9 @@
 // This package implements the Protocol 1 of "Practical Threshold Signatures"
 // by Victor Shoup [1].
 //
+// Warning: Sign operations rely on math/big and are not constant time in the
+// secret key share. Pass a non-nil random source to Sign to enable blinding.
```
cjpatton, line 7: Do you mean "blinding" as in "masking side channels" or "blinding factor in a cryptographic operation"? If the former, mention how the random source is used for masking.

It also seems a bit weird that randomness source controls this behavior. What else is the randomness source used for in this function?

lukevalenta (approved):
Approved pending addressing Chris' comments

tss/rsa/rsa_threshold.go
bwesterb: Blinding is the standard way to do this for RSA: it's rare for RSA to be implemented in constant time. I'd rather mention the fact then give a lecture in the comments.

blindsign/blindrsa/partiallyblindrsa/pbrsa.go
bwesterb: I think it's better to write a bit about this in README.md (todo). Better to mark where things are not constant time for now.

bwesterb: https://github.com/cloudflare/circl/issues/657

PR #644, opened 2026-07-15
secretsharing: check that share ID is not zero.

bwesterb:
Recover() didn't check whether the share ID is zero, which allows one manipulated share to determine the final recovered secret. The applicability is rather narrow: the Verify() function that checks secret commitments already did check whether the share ID is zero. Without calling Verify(), an attacker can already manipulate the shared secret already to a certain degree. Arguably, we wouldn't need to check it here, but it's cheap to do so.

---

<a href="https://app.devin.ai/review/cloudflare/circl/pull/644" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/secretsharing/ss_test.go
+++ b/secretsharing/ss_test.go
@@ -104,6 +104,32 @@ func TestShareWithID(tt *testing.T) {
 	})
 }
 
+func TestRecoverZeroID(tt *testing.T) {
+	g := group.P256
+	t := uint(2)
+	n := uint(5)
+
+	secret := g.RandomScalar(rand.Reader)
+	ss := secretsharing.New(rand.Reader, t, secret)
+	shares := ss.Share(n)
+
+	// A single Share{ID:0, Value:V} would otherwise force Recover to return
+	// the attacker-chosen V, independent of the honest shares.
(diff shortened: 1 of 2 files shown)


secretsharing/ss_test.go
```diff
@@ -104,6 +104,32 @@ func TestShareWithID(tt *testing.T) {
...
+	secret := g.RandomScalar(rand.Reader)
+	ss := secretsharing.New(rand.Reader, t, secret)
+	shares := ss.Share(n)
+
+	// A single Share{ID:0, Value:V} would otherwise force Recover to return
+	// the attacker-chosen V, independent of the honest shares.
+	attackerValue := g.RandomScalar(rand.Reader)
+	tampered := make([]secretsharing.Share, 0, t+1)
+	tampered = append(tampered, secretsharing.Share{
+		ID:    g.NewScalar(),
```
lukevalenta, line 121: Optional, to make it a little more clear that this is the zero ID.
```suggestion
		ID:    g.NewScalar(), // a zero-valued scalar, the tampered share
```

lukevalenta (approved):

PR #648, opened 2026-07-15
kyber: document pk isn't checked like ML-KEM

bwesterb:
The round 3 Kyber reference implementation accepts unreduced coefficients for the public key, but hashes the unreduced coefficients into the shared secret, which binds them.

ML-KEM instead checks whether the coefficients of a public key are reduced.

Both approaches have up and downsides. Our Kyber implementation matches what the Kyber reference implementation did.

This commit adds a comment noting that, so that GenAI doesn't warn about the mismatch with ML-KEM.

---

<a href="https://app.devin.ai/review/cloudflare/circl/pull/648" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/kem/kyber/kyber1024/kyber.go
+++ b/kem/kyber/kyber1024/kyber.go
@@ -246,6 +246,11 @@ func (pk *PublicKey) Pack(buf []byte) {
 // Unpacks pk from buf.
 //
 // Panics if buf is not of size PublicKeySize.
+//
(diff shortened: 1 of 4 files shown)


lukevalenta:
Is there a reason to keep maintaining this version of kyber?

lukevalenta (approved):

bwesterb:
We should check how many users remain. It's a small diff on top of ML-KEM, so the bar is low.

PR #651, opened 2026-07-15
ecc/p384: document that package is not fully constant time

bwesterb:
---

<a href="https://app.devin.ai/review/cloudflare/circl/pull/651" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/ecc/p384/doc.go
+++ b/ecc/p384/doc.go
@@ -4,7 +4,7 @@
 //   - Around 10x faster in amd64 architecture.
 //   - Reduced number of memory allocations.
 //   - Native support for arm64 architecture.
-//   - ScalarMult is more constant time, but not fully.
+//   - ScalarMult is not fully constant time (see note below).
 //   - ScalarBaseMult fallbacks into ScalarMult.
 //   - A new method included for double-point multiplication.
 //
--- a/ecc/p384/doc.go
+++ b/ecc/p384/doc.go
@@ -4,7 +4,10 @@
 //   - Around 10x faster in amd64 architecture.
 //   - Reduced number of memory allocations.
 //   - Native support for arm64 architecture.
-//   - ScalarMult is performed using a constant-time algorithm.
+//   - ScalarMult is more constant time, but not fully.
 //   - ScalarBaseMult fallbacks into ScalarMult.
 //   - A new method included for double-point multiplication.
+//
+// Note: the scalar-multiplication routines in this package are not fully
+// constant time.
 package p384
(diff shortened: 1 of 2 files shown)


ecc/p384/doc.go
```diff
@@ -4,7 +4,10 @@
 //   - Around 10x faster in amd64 architecture.
 //   - Reduced number of memory allocations.
 //   - Native support for arm64 architecture.
-//   - ScalarMult is performed using a constant-time algorithm.
+//   - ScalarMult is more constant time, but not fully.
```
lukevalenta:
```suggestion
//   - ScalarMult is not fully constant time (see note below).
```

ecc/p384/doc.go
```diff
@@ -4,7 +4,10 @@
...
 //   - ScalarBaseMult fallbacks into ScalarMult.
 //   - A new method included for double-point multiplication.
+//
+// Note: the scalar-multiplication routines in this package are not fully
+// constant time.
```
lukevalenta, line 12: Just to add a little more detail:
```suggestion
// constant time. Execution time and memory access can depend on the value of
// the secret scalar, so avoid using these routines with secret scalars where
// timing side-channels are a concern.
```

ecc/p384/p384opt.go
```diff
@@ -78,6 +78,7 @@ func (c curve) toOdd(k []byte) ([]byte, int) {
 }
 
 // ScalarMult returns (Qx,Qy)=k*(Px,Py) where k is a number in big-endian form.
+// It is not fully constant time.
```
lukevalenta, line 81:
```suggestion
// It is not fully constant time; see the package documentation for details.
```

ecc/p384/p384opt.go
```diff
@@ -123,7 +124,7 @@ func (c curve) scalarMultOmega(x1, y1 *big.Int, k []byte, omega uint) (x, y *big
 }
 
 // ScalarBaseMult returns k*G, where G is the base point of the group
-// and k is an integer in big-endian form.
+// and k is an integer in big-endian form. It is not fully constant time.
```
lukevalenta, line 127:
```suggestion
// and k is an integer in big-endian form. It is not fully constant time;
// see the package documentation for details.
```

ecc/p384/doc.go
lukevalenta: Claude suggested this as a description of where the non-constant timedness comes from. I have not carefully checked though, so if it doesn't sound right please ignore:
```
// Note: the scalar-multiplication routines in this package are not fully
// constant time. The windowed point arithmetic itself is constant time, but
// the scalar reduction and recoding (reduceScalar and toOdd) use variable-time
// math/big operations, so execution time and memory allocations can depend on
// the value of the secret scalar. Avoid using these routines with secret
// scalars where timing side-channels are a concern.
```

lukevalenta (commented):
Is this worth updating to make this constant-time like `crypto/internal/nistec` is?

lukevalenta (approved):

PR #661, opened 2026-07-17
ecc/fourq: add PoC for reduction bug (amd64, !hasBMI2).

cjpatton:


bwesterb:
I presume superseded by #669 ?

PR #667, opened 2026-07-21
internal/test: unify ACVP test vector parsing

### LINKED ISSUE #533: pq: unify the ACVP code for testing
Unify the code used for parsing ACVP test vectors.

> Let's open an issue to unify the ACVP code.

_Originally posted by @bwesterb in https://github.com/cloudflare/circl/pull/512#discussion_r1875962940_

ihopenre-eng:
Fixes #533.

`kem/mlkem/acvp_test.go` and `sign/dilithium/templates/acvp.templ.go` each carried their own copy of the same 50 lines: read the gzipped `prompt` and `expectedResults` files, decode both into raw test groups, then index the results by `tcId` while rejecting duplicates. The template's copy is expanded into `sign/mldsa/mldsa{44,65,87}/acvp_test.go`, so the block existed four times in the tree.

This moves it into `internal/test` as `ReadACVP`:

```go
vectors := test.ReadACVP(t, "testdata/ML-KEM-"+sub+"-FIPS203")

for _, rawGroup := range vectors.Groups {
        ...
        vectors.Result(t, tst.TcID, &result)
}
```

`Groups` stays raw `json.RawMessage`, so each call site keeps unmarshalling into the shape its algorithm expects. `Result` fails the test on a missing id, which is what the inlined code did.

Net effect is 350 lines removed for 112 added, and a new ACVP-based test no longer has to re-derive the loader.

### Scope

`sign/slhdsa/acvp_test.go` is deliberately left alone. It matches prompts to results *positionally* through fully typed structs (`outputs.TestGroups[gi].Tests[ti]`) rather than by `tcId`, and asserts the ids line up. Folding it into `ReadACVP` would change what those tests assert, which did not seem like something to bundle into a refactor. Happy to do it as a follow-up if you would prefer the whole ACVP surface on one loader, since the id-keyed lookup is the more robust of the two.

### Verification

- `go test -count=1 ./...` passes.
- `go generate -v ./...` leaves the working tree unchanged; the three `mldsa*/acvp_test.go` files in this PR are the regenerated output of the template change, not hand edits.
- `go vet ./...` and `go vet -vettool=shadow ./...` are clean.
- `golangci-lint run` reports nothing on `internal/test`, `kem/mlkem` and `sign/mldsa/...`.
- The ACVP subtests still execute rather than silently skipping:

```console
$ go test -count=1 -run TestACVP -v ./kem/mlkem/ ./sign/mldsa/mldsa44/
=== RUN   TestACVP
=== RUN   TestACVP/keyGen
=== RUN   TestACVP/encapDecap
--- PASS: TestACVP (0.04s)
ok      github.com/cloudflare/circl/kem/mlkem
=== RUN   TestACVP
=== RUN   TestACVP/keyGen
=== RUN   TestACVP/sigGen
=== RUN   TestACVP/sigVer
--- PASS: TestACVP (0.07s)
ok      github.com/cloudflare/circl/sign/mldsa/mldsa44
```

No test vectors, assertions, public API or crypto paths are touched.

---

<a href="https://app.devin.ai/review/cloudflare/circl/pull/667" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/internal/test/acvp.go
+++ b/internal/test/acvp.go
@@ -1,0 +1,87 @@
+package test
+
+import (
+	"encoding/json"
+	"path/filepath"
+	"testing"
+)
+
+// ACVP is a decoded pair of ACVP test vector files: the prompt file, holding
+// the test groups, and the expectedResults file, holding the results indexed
+// by test case id.
+//
+// Vectors come from https://github.com/usnistgov/ACVP-Server, where every
+// algorithm directory pairs a prompt.json with an expectedResults.json.
+type ACVP struct {
+	// Groups are the raw test groups of the prompt file. Unmarshal each one
+	// into whatever shape the algorithm under test expects.
+	Groups []json.RawMessage
+
+	results map[int]json.RawMessage
+}
+
+// ReadACVP reads the gzipped prompt and expectedResults files of the ACVP
+// test vector directory dir.
+func ReadACVP(t testing.TB, dir string) *ACVP {
+	t.Helper()
+
+	acvp := &ACVP{
+		Groups:  readACVPGroups(t, filepath.Join(dir, "prompt.json.gz")),
+		results: make(map[int]json.RawMessage),
+	}
+
+	for _, rawGroup := range readACVPGroups(t, filepath.Join(dir, "expectedResults.json.gz")) {
+		var group struct {
+			Tests []json.RawMessage `json:"tests"`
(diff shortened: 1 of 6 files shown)


bwesterb:
Thanks for this!

bwesterb (approved):

PR #693, opened 2026-08-24
ml-dsa: support external-mu and update wycheproof test vectors

bwesterb:


sign/dilithium/templates/pkg.templ.go
```diff
@@ -258,15 +304,53 @@ func (sk *PrivateKey) Sign(rand io.Reader, msg []byte, opts crypto.SignerOpts) (
...
+			if len(o.Randomness) != len(rnd) {
+				return nil, errors.New("mldsa: randomness must be 32 bytes")
+			}
+			copy(rnd[:], o.Randomness)
+		}
+		ctx = []byte(o.Context)
+	case sign.SignatureOpts:
+		ctx = []byte(o.Context)
+	case *sign.SignatureOpts:
+		ctx = []byte(o.Context)
```
Lekensteyn, line 328: FIPS 204 (Sections 3.4 Hedged and Deterministic Signing, 5.2 ML-DSA Signing) state that deterministic is optional, suggesting to default to randomized "hedged" variant to mitigate side-channel attacks (particularly fault injection).

The Go crypto/mldsa implementation defaults to randomized signing and adds a SignDeterministically API to make it more clear.

The previous implementation already ignored the `rand` parameter, and used deterministic signing as well. So it is not a new problem, but would it be worth enabling it by default? And if not, clearly document this caveat. Could be a follow-up change.

Lekensteyn (approved):
Change itself looks reasonable, only one [comment](https://github.com/cloudflare/circl/pull/693/changes#r3843985827) on lack of extra randomness in signing.

sign/dilithium/templates/pkg.templ.go
bwesterb: Deterministic signing is completely fine in software and it has the distinct advantage that it makes testing test vectors easier: not just for ML-DSA itself, but also in downstream protocols. I'll add support for hedged signing in a later PR.

PR #694, opened 2026-08-26
oprf: reject input and info that overflow the length prefix

dxbjavid:
The OPRF and POPRF hashing steps frame each variable-length field with I2OSP(len, 2) as RFC 9497 specifies, and the prefix is built with a uint16 cast over the field length. For the client input, the server-side full-evaluation input, and the DeriveKey info that length is never bounded first, so a value of 65536 bytes or more wraps the two-byte prefix (a 65536-byte input encodes a length of 0) and the value is accepted rather than rejected. scalarFromInfo already guards its info with the same math.MaxUint16 check and returns ErrInvalidInfo, so the eval-path info is fine, but the sibling sites that frame the input and the derive-key info were missed. I came across it while comparing the length-prefix handling across the package and confirmed all three call sites accept an over-length value today. The change adds the same bound at blind, fullEvaluate, and DeriveKey, and keeps the maximum admissible length of 65535 bytes working so valid callers are unaffected.

---

<a href="https://cloudflare.devinenterprise.com/review/cloudflare/circl/pull/694" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/oprf/client.go
+++ b/oprf/client.go
@@ -2,6 +2,7 @@ package oprf
 
 import (
 	"crypto/rand"
+	"math"
 
 	"github.com/cloudflare/circl/group"
 	"github.com/cloudflare/circl/zk/dleq"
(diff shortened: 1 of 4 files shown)


bwesterb:
Thanks. Simpler to check in client.validate.

dxbjavid:
Makes sense, moved it. The input bound now sits in client.validate alongside the other finalize-state checks, so all three Finalize paths pick it up, and the test drives a full round through Finalize rather than stopping at Blind. One consequence worth noting is that Blind itself now accepts an over-length input and the rejection surfaces at Finalize. The FullEvaluate and DeriveKey checks are unchanged since they don't pass through validate.

cjpatton (approved):

Issue #695, opened 2026-08-27
Integer overflow in appendLen16Prefixed causes unrecoverable ciphertext in ABE scheme


ETCaton:
(This was previously reported to, and categorized as "Informational" and outside their threat model by, Cloudflare)

appendLen16Prefixed (in internal/tkn/util.go) stores the byte length in a uint16. Any policy whose overall length exceeds 65535 bytes will wrap at uint16(len(pBytes)) and cause encryption to "succeed" with an unrecoverable ciphertext (error at decryption time, nothing at encryption)

That is, any combination of numerocity and specific attribute length summing to over 65535 will overflow and not be recoverable

PR #696, opened 2026-08-27
Check for overflow when encrypting under ABE

ETCaton:
Because of the use of uint16, >65535 byte values put into the header will wrap around silently and cause decryption to later fail with a confusing "too short" message.

This fixes encryption so that it rejects loudly and this can't silently occur (either accidentally or by malicious policy)

See #695 

---

<a href="https://cloudflare.devinenterprise.com/review/cloudflare/circl/pull/696" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


cjpatton (commented):
Thanks for the patch! This ends up being a pretty big change since we need to update the internal serialization APIs. Ultimately I think the right thing to do is to replace the customer serialization logic with [`cryptobyte`](https://pkg.go.dev/golang.org/x/crypto/cryptobyte). Would you be interested in pivoting your PR to do this instead?

ETCaton:
Sure!

I've not written much Go outside of personal projects years ago, so let me read the docs and rethink it. I'll push when I have something that seems to work

cjpatton:
> Sure!
> 
> I've not written much Go outside of personal projects years ago, so let me read the docs and rethink it. I'll push when I have something that seems to work

Sounds good. Some considerations:

1. Lots of packages in this repo _could_ use cryptobyte, but for this PR, just focus on ABE.
2. For inspiration, have a look at crypto/tls in the Go standard library.

ETCaton:
I'm not sure if I'm missing something, but it seems like `cryptobyte` only works big-endian and so it wouldn't work here for the little-endian values. I'll see if I can use it to wrap some pieces without touching little/big endian

cjpatton:
> I'm not sure if I'm missing something, but it seems like `cryptobyte` only works big-endian and so it wouldn't work here for the little-endian values. I'll see if I can use it to wrap some pieces without touching little/big endian

Ugh, you're right. That sucks :)

I'll review the PR as is then.

PR #700, opened 2026-09-06
pki: return an error for schemes without an object identifier

lenamonj:
The four `pki` marshal functions asserted `scheme.(CertificateScheme)` to reach an object identifier, and `MarshalPKIXPrivateKey` asserted `sk.(sign.Seeded)` for ML-DSA. A scheme without an OID panicked instead of returning the error the signature declares. Dilithium and SLH-DSA are registered in `sign/schemes` without one, so this is reachable with the library's own keys.

Each site now returns an error. The new test marshals a key of every registered scheme that lacks an OID through all four functions; on main it panics.

---

<a href="https://cloudflare.devinenterprise.com/review/cloudflare/circl/pull/700" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-devin-review-dark.svg?v=3">
    Devin Review
  </picture>
</a>

### DIFF
--- a/pki/marshal_nooid_test.go
+++ b/pki/marshal_nooid_test.go
@@ -1,0 +1,68 @@
+package pki_test
+
+import (
+	"crypto"
+	"io"
+	"strings"
+	"testing"
+
+	"github.com/cloudflare/circl/pki"
+	"github.com/cloudflare/circl/sign"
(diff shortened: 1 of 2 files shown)


bwesterb (approved):

bwesterb:
Thanks!

